PR #1, opened 2025-01-28
First data modules

GaganNarula:
This PR is in a draft stage. I am trying to score feedback and opinions.

DESCRIPTION FROM COPILOT
This pull request introduces several new features and improvements to the ESP data project, including the addition of CI workflows, enhancements to the README file, new Pydantic models for data configuration, and CRUD operations for BigQuery. Below are the most important changes:

### CI Workflow:
* Added a new GitHub Actions workflow for continuous integration to run Python tests on push and pull request events to the main branch. This includes steps for setting up Python, installing dependencies, and running tests. (`.github/workflows/python-ci.yml`)

### Documentation:
* Updated the `README.md` file to include setup instructions for the project, including authentication with GCP, cloning the repository, installing dependencies, and running tests. (`README.md`)

### Data Configuration:
* Added new Pydantic models for data configuration, including `DataSample`, and `DatasetConfig`. These models include various fields, validators, and helper methods for managing data samples and datasets. (`esp_data/config/db_config.py`)

### Dataset Handling:
* Introduced base classes for datasets that include methods for data manipulation such as `__getitem__`, `__len__`, `subset`, `sample`, `from_dict`, `from_samples`, `from_path`, `to_dict`, `concatenate`, `filter`, `map`, and `save_to_path`. (`esp_data/datasets/base.py`)
* Added a new dataset class `HFDataset` that inherits from `BaseDataset` and implements methods for handling datasets. (`esp_data/datasets/hf.py`)
* Added a webdataset class `WebDataset`
* Sharding: shards can be created from a list[dict] or pandas DataFrame in `shard_creator.py`.

### BigQuery Operations:
* Implemented CRUD operations for BigQuery, including creating tables, inserting rows, querying, and updating rows. Added utility functions for converting Pydantic models to BigQuery schemas and saving schemas to JSON files. (`esp_data/db/bq.py`, `esp_data/db/utils.py`) [[1]](diffhunk://#diff-51cce12ffdd2900a048e6d5b7eb5e76172a3eb19cc88e7fb57f7e2ade96d3807R1-R116) [[2]](diffhunk://#diff-4960f25e1d895484f26f91e4cb1d78d1dfd1d5c3581657174dbaa21951daf164R1-R74)

These changes collectively enhance the project's functionality, improve documentation, and streamline data handling and integration with BigQuery.


GaganNarula:
CI tests dont work because of auth. Will have to change the python-ci workflow to use the github actions service account i think

mil-ad:
It might be a good idea to version the pydantic model and export a versioned [schema](https://docs.pydantic.dev/latest/concepts/json_schema/) in case we later make changes

mil-ad:
I haven't gone through everything but some random thoughts:

- logging
    - the `__init__.py` is setting up logging handlers. A "library" shouldn't do that, that's for the application to set up
    - I see that in a few places we do `logging.getLogger(__name__)` for logging. Given that this is meant to be used as a library we should probably be consistent and do `logging.getLogger("esp_data")` everywhere. This makes it easier for the downstream application to configure our logger.
- config
    - instantiating `default_shard_creator_cfg` and `default_webds_loader_cfg` and then using individual fields in other models doesn't look a good pattern to me.
    - `ShardCreatorConfig` and `WebDsLoaderConfig` are not used anywhere. What's the point of them? fame for `TextDataSample`?
    - we do things like `self.created_at.isoformat()` in our `to_json()` or `to_dict()` methods. While we're careful about reversing them in the `from_json()` and `from_dict()` this is not very robust code. I think we could potentially use `@field_serializer` decorator for this purpose?
- data module
    - Not convinced this library should do things like `gsutil_ls`, especially via calling gsutil from shell. I would generally stay away from `subprocess` via gsutil/gcloud in this library. Google has python APIs for these operations so _if_ we really want to support something we should do it that way. 
    - I still don't quite get the point of "functional.py" in the data module. I'm generally a big fan of functional codes but what does going stateless here give us? We're also duplicating a lot of the functionality that should already be in cloudpathlib and Pathlib, for example both Pathlib and cloudpathlib have `exists()` and we should mostly rely on them. I think we should go for something super light.
    - `DEBUG` as a key in `dotenv` is not very descriptive.

mil-ad:
also - should we consider switching to `src/` layout for this? https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

mil-ad:
is `make_storage_options()` meant to be exposed to the user? It looks like something we should be able to figure out behind the scenes

Agree! we shouldnt have to expose this, in fact the implementation takes care of this based on the path

mil-ad:
something I notice in a few places is using functions instead of relying on methods. We have different classes for each storage (`GSPath`, `S3Path`, etc) so we repeat the usual logic of `if is_gcs_path(path)`, `elif is_s3_path(path)` etc etc in each function

mil-ad:
We use `AnyPath` as a type in a few places but currently there's no way to instantiate `AnyPath`. The factory `__new__` only returns `Path | GSPath | S3Path`

Not sure what you mean here... do you mean it should return "cloudpathlib.AnyPath" ? but why should we do that ?

GaganNarula:
> I haven't gone through everything but some random thoughts:
> 
> * logging
>   
>   * the `__init__.py` is setting up logging handlers. A "library" shouldn't do that, that's for the application to set up
>   * I see that in a few places we do `logging.getLogger(__name__)` for logging. Given that this is meant to be used as a library we should probably be consistent and do `logging.getLogger("esp_data")` everywhere. 

This makes it easier for the downstream application to configure our logger.
Yes agreed. some things need a logger (and creating a child logger is fine as you say). i was somehow getting duplicate logging and this root logger was an attempt to solve it. Happy to leave it you if you want to fix this in your PR ?

> * config
>   
>   * instantiating `default_shard_creator_cfg` and `default_webds_loader_cfg` and then using individual fields in other models doesn't look a good pattern to me.

But weren't we doing that in NatureLM ? (model.from_config) I was trying to do the same.

>   * `ShardCreatorConfig` and `WebDsLoaderConfig` are not used anywhere. What's the point of them? fame for `TextDataSample`?

default_shard_creator_cfg is an instance of ShardCreatorConfig with default values. Same for the default_webds_loader_cfg. I suppose it might have been better to create ShardCreatorConfig and WebDsLoaderConfig with no default values and fill them in when running something like `default_cfg = Config(**default_params)`. I wanted to define default parameters for the so this was what I went for, but I'm fine if we want a different solution like a yaml file ? or just refactor this... 

>   * we do things like `self.created_at.isoformat()` in our `to_json()` or `to_dict()` methods. While we're careful about reversing them in the `from_json()` and `from_dict()` this is not very robust code. I think we could potentially use `@field_serializer` decorator for this purpose?

Ah this is actually a mistake in the code. I opted for "created_at" to be a string (that is validated to be a UTC datetime). These `self.created_at.isoformat()` shouldn't be there in `to_json` and `to_dict`

> * data module
>   
>   * Not convinced this library should do things like `gsutil_ls`, especially via calling gsutil from shell. I would generally stay away from `subprocess` via gsutil/gcloud in this library. Google has python APIs for these operations so _if_ we really want to support something we should do it that way.

Yes in general i agree with you, running things in shell can be dangerous. I haven't included gsutil_ls in the functional api. I included it here because I wanted to test the speed difference between cloudpathlib's "ls" method (as well as filesystem ls) vs running the cli via subprocess. Turns out cli via subprocess is still 3x faster than both the filesystem and cloudpathlib (which probably uses filesystem?)...

>   * I still don't quite get the point of "functional.py" in the data module. I'm generally a big fan of functional codes but what does going stateless here give us? We're also duplicating a lot of the functionality that should already be in cloudpathlib and Pathlib, for example both Pathlib and cloudpathlib have `exists()` and we should mostly rely on them. I think we should go for something super light.

Yeah I think this may have been a misunderstanding on my part. I think we were discussing the initial implementation I had (using object oriented code with classes like File and Bucket, which had the same api as cloudpathlib, e.g. File.open(".../some/path")). You said we should rather have some thing simple like read_file(...) which _sounded_ like be functional. RE this: `also duplicating a lot of the functionality that should already be in cloudpathlib and Pathlib` the other thing that's happening in this api is that if cloudpathlib fails (which btw it does in some cases and i haven't been able to successfully reproduce these errors sometimes), we use the filesystem approach to run the same thing. So the `exists` method in functional is not exactly duplicate..

However, I am happy to have just one way of doing things to make it simpler and if that means we just use "AnyPath" and have methods like `p = AnyPath(); p.open(); p.exists()` etc., then let's do that! We could extend it to have `p.rsync` etc.

>   * `DEBUG` as a key in `dotenv` is not very descriptive.
Agreed!

GaganNarula:
> also - should we consider switching to `src/` layout for this? https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

Oh.. hmm... i guess it wouldn't really break anything right ? as long as the package build code in pyproject.toml is correct ?

GaganNarula:
> It might be a good idea to version the pydantic model and export a versioned [schema](https://docs.pydantic.dev/latest/concepts/json_schema/) in case we later make changes

Haven't done this yet, if you'd like to do this in your PR

GaganNarula:
> something I notice in a few places is using functions instead of relying on methods. We have different classes for each storage (`GSPath`, `S3Path`, etc) so we repeat the usual logic of `if is_gcs_path(path)`, `elif is_s3_path(path)` etc etc in each function

Feel free to change, but i'm not sure which places in the code you're referring to..

GaganNarula:
Merging this branch so that we can start developing smaller PR's

PR #2, opened 2025-03-18
 Refactor cloud path operations and streamline R2 path handling

mil-ad:
🚚 renamed `_make_file_opener` -> `make_file_opener`. It wasn't local.
🗑️ Removed `get_cloud_prefix()`. There's _currently_ no need for it because Cloudpathlib classes have `cloud_prefix` class variables
🚚 renamed `is_cloudflarer2_path()` -> `is_r2_path()`. R2 is clear enough?
🔨 removed `_make_r2_path_with_auth()` and replaced it with the same mechanism we use for `GSPath` i.e. a cached client and a replacement class (`R2Path`)
🔨 Refactored `AnyPath`

### DIFF
--- a/esp_data/paths.py
+++ b/esp_data/paths.py
@@ -95,8 +95,8 @@ class GSPath(cloudpathlib.GSPath):
 
 class R2Path(cloudpathlib.S3Path):
     def __init__(self, cloud_path: str | cloudpathlib.S3Path, client=_get_r2_client()):
-        if isinstance(cloudpathlib, str):
-            cloud_path = cloudpathlib.replace("r2://", "s3://")
+        if isinstance(cloud_path, str):
+            cloud_path = cloud_path.replace("r2://", "s3://")
 
         super().__init__(cloud_path, client=client)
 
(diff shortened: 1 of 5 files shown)


esp_data/paths.py
```diff
@@ -99,38 +89,42 @@ class GSPath(cloudpathlib.GSPath):
...
         Solves the issue of disappearing // in the path string when using cloudpathlib.AnyPath.
 
         Args:
-            path (str | os.PathLike | Path | AnyPath): The path to a file or directory.
+            path (str | Path | GSPath | R2Path): The path to a file or directory.
 
         Returns:
             Path | GSPath | S3Path: The correct path object based on the path string.
         """
-        if isinstance(path, cls):
```
GaganNarula, line 119: Good catch!

GaganNarula:
There are tests for the paths module in tests/test_paths.py .. maybe we should extend that ? You could try running the current (very minimal) tests.

mil-ad:
thanks that test just helped me fix one bug! 🤯  I'll extend tests as soon as possible, maybe in the next PR

PR #3, opened 2025-03-20
More refactor

mil-ad:
Importing `esp_data` package was quite slow. It was mostly because we had calls to _get_client*() functions in the `__init__` default args which meant that we would instantiate the client when importing the package. I moved the clients to class variables where they only get instantiated at first use.

I also tidied up the `__init__.py`, wasn't sure why we needed config or logging there.

### DIFF
--- a/esp_data/__init__.py
+++ b/esp_data/__init__.py
@@ -1,33 +1,3 @@
-import logging
-import os
-import warnings
+from esp_data.paths import AnyPath, GSPath, R2Path
 
-from dotenv import load_dotenv
-
-from esp_data.paths import AnyPath
-
-load_dotenv()
-
-warnings.filterwarnings("ignore", "Your application has authenticated using end user credentials")
-CONFIG = {
-    "DEBUG": os.getenv("DEBUG", "False") == "True",
-}
-
-# make root logger
-logger = logging.getLogger("esp_data")
-
-# Set up a handler for this logger only
-if CONFIG["DEBUG"]:
-    level = logging.DEBUG
-else:
-    level = logging.INFO
-
-logger.setLevel(level)
-
-# Only add a handler if one doesn't exist yet
-if not logger.handlers:
-    handler = logging.StreamHandler()
-    handler.setFormatter(logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s"))
-    logger.addHandler(handler)
-
-__all__ = ["AnyPath", "CONFIG"]
+__all__ = ["AnyPath", "GSPath", "R2Path"]
--- a/esp_data/paths.py
+++ b/esp_data/paths.py
@@ -1,8 +1,6 @@
 """Path definitions that get rid of some issues with the AnyPath type hint."""
 
 import os
-import warnings
-from functools import lru_cache
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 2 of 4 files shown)


esp_data/__init__.py
```diff
@@ -1,33 +1,3 @@
-import logging
-import os
-import warnings
+from esp_data.paths import AnyPath, GSPath, R2Path
 
-from dotenv import load_dotenv
-
-from esp_data.paths import AnyPath
-
-load_dotenv()
```
GaganNarula, line 9: why not load the env variables here ? there are some env vars like the gcp project id and the gcp project number etc that are not secrets in the secret manager..

esp_data/paths.py
```diff
@@ -89,12 +73,30 @@ class GSPath(cloudpathlib.GSPath):
     For more details, see: https://github.com/drivendataorg/cloudpathlib/issues/390
     """
 
-    def __init__(self, cloud_path: str | cloudpathlib.GSPath, client=_get_gs_client()):
+    _client = None
+
+    def __init__(self, cloud_path: str | cloudpathlib.GSPath, client: Optional[GSClient] = None):
+        if not client:
+            if not GSPath._client:
+                GSPath._client = cloudpathlib.GSClient(storage_client=GS_Client_Official())
```
GaganNarula, line 79: I wonder if we'll need to point to credential files when we instantiate `GS_Client_Official`. . thinking about a case where we are using this on lambda labs and a service account cred file is needed .. unless a researcher is logging into their gcp account on those VMs..

GaganNarula (changes requested):
Removing dotenv here will break utils.py and paths.py (os.getenv calls). I propose extending this PR to do:

1. Find and add a google python client (google cloud resource manager ?) to dependencies, for determining things like "project_id" and "project_number". 
2. Call this resource manager in`__init__.py` and use the values in ->
3. The secrets get method (`utils.get_secret`) in the library `__init__` (or elsewhere as we find appropriate) to load secrets from google secrets API
3. Add these loaded secrets and extra vars into the env vars with os.environ[]

esp_data/paths.py
mil-ad: I don't think we need to do this in the library. All we can do is try/except and complain in a nice way. It's up to the host to set authentication up.

If you're on your local machine you need to do do [Application Default](https://cloud.google.com/docs/authentication/provide-credentials-adc) authentication for this to work. On a lambdalabs machine we need to put a service account and make sure `GOOGLE_APPLICATION_CREDENTIALS` is defined. On a GCP VM (with servcie account) this is is done automatically from the internal metadata server:

https://cloud.google.com/docs/authentication/application-default-credentials

mil-ad:
Done! I now read R2 secrets for initializing the client. I don't inject it back into the env vars because they're (slightly) safer in memory than env variables. Would we need them in the env? 

I also had to delete `validate_path_exists()` because it made utils module import `AnyPath` which would cause a circular dependency

esp_data/__init__.py
mil-ad: why should gcp project ID come from an env variable? We already know what it is by default

GaganNarula: ah so we're ok with putting the project_id hard coded as the default in `read_gcp_secret` ? maybe for later: we might want to build containers that use esp-data at a later date, which would first deploy and test in the "staging-xxx" environment so that id might actually change... anyhow I agree with removing dotenv from here..

GaganNarula (approved):

PR #4, opened 2025-03-26
refactoring

mil-ad:
- Started removing functions from `functional.py`. I think this complicates the library too much and doesn't bring enough value over the class methods. In this PR I have removed:
  - `create_file()` which is actually not used anywhere except tests
  - `exists()` which was used in a few places
- Removed `make_simple_logger()`. A library shouldn't really set up logging handlers
- Removed `gsutil_ls()`. It uses subprocess which which is not super reliable. It's not anywhere in the library. I think in general we should have a higher bar in this library for defining utility functions. We should focus on things that our users do a lot and is cumbersome.
- Removed `make_file_opener()`. Yet another place where we repeat our isLocal/isNotLocal logic and this function creates parent directories if they don't exist, which is not clear in the interface. It's only used in two places, in `wds.ShardWriter` and then in snippet below:
  ```
  opener = make_file_opener(path, mode="wb")
  file_obj = opener(path)
  ```
  We could instead just do: `file_obj = AnyPath(path).open("wb")`. For `wds.ShardWriter` I replaced it with a simple lambda function `lambda: AnyPath(shard_path).open("wb")`


GaganNarula (approved):
ok looks good. We might want to use some of that filesystem related code if are ditching cloudpathlib though,

mil-ad:
Agreed. If we run into any issues I'll turn them into tests and then patch inside AnyPath. I know what to look for after your commen

PR #5, opened 2025-03-26
refactor: remove open_file utility and replace with AnyPath usage in …

mil-ad:
- Removed `open_file()` from functional. Was only used in the naturelm script
- Removed `read_bytes()` and `read_text()`
- Removed `upload()`. This literally just calls `copy()` with the same args.
- Removed `download()`. This is again just a wrapper for `copy()`

### DIFF
--- a/esp_data/file_io/functional.py
+++ b/esp_data/file_io/functional.py
@@ -127,76 +127,6 @@ def yield_files(dir_path: str | os.PathLike | AnyPath, pattern: str = "*", use_f
         raise IOError(f"Failed to yield files in {dir_path} using both methods: {e}") from e
 
 
-def write_bytes(file_path: str | os.PathLike | AnyPath, data: bytes, use_fs: bool = False) -> bool:
-    """Write the data to the file.
-
-    Args:
-        file_path (str | os.PathLike | AnyPath): The path to the file, local or cloud.
-        data (bytes): The data to write to the file.
-        use_fs (bool, optional): If True, use the FileSystem approach. Defaults to False.
-
-    Returns:
-        bool: True if the data was written successfully.
-    """
-    file_path = AnyPath(file_path)
-
-    if is_local_path(file_path):
-        file_path.write_bytes(data)
-        return True
-
-    if not use_fs:
-        try:
-            file_path.write_bytes(data)
-            return True
-        except Exception as e:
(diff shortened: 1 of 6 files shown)


esp_data/file_io/functional.py
```diff
@@ -495,52 +329,6 @@ def copy(
...
-        bool: True if the download was successful.
-
-    Example:
-        # Download a file from a cloud bucket
-        download("gs://bucket_name/path/to/file.txt", "local_file.txt")
-
-        # Download a dir from a cloud bucket
-        download("gs://bucket_name/path/to/dir/", "local_dir/")
-    """
-    return copy(source, destination, use_fs)
```
GaganNarula, line 541: yeah this was just to make aliases but i see how its kinda useless 😂 will keep things cleaner and simpler going forward

esp_data/file_io/functional.py
```diff
@@ -495,52 +329,6 @@ def copy(
...
-
-
 def gcloud_rsync(
```
GaganNarula: probably also a candidate to remove because it uses subprocess, maybe you could remove it in this PR itself ? I dont think its used anywhere

tests/test_fileio_functional.py
```diff
@@ -2,19 +2,11 @@
 
 from esp_data import AnyPath
 from esp_data.file_io.functional import (
-    # create_file,
+    copy,
     delete_dir,
     delete_file,
-    download,
```
GaganNarula, line 8: We will probably just have one test suite for the AnyPath implementation right ? I guess we can keep these tests for future PR.

GaganNarula (changes requested):
`gcloud_rsync` can be removed in this PR already

GaganNarula:
Another change required:
in config/db_config, there's still one `open_file` call that needs to be replaced
Line 324:
```
with open_file(file_path, "w") as f:
    f.write(text)
```

esp_data/file_io/functional.py
mil-ad: done!

tests/test_fileio_functional.py
mil-ad: yeah I've been trying to keep the tests running! The number of failures have actually gone down now.

GaganNarula (approved):
LGTM!

PR #6, opened 2025-03-27
cached_class_property for dealing with storage options and clients

mil-ad:
- Added `cached_class_attribute` decorator. Python has `@cached_property` but they're for instance attributes not class attributes. 
- Removed `make_storage_options()` and instead used the new `cached_class_attribute` to have a `.storage_options` attr
- Do the client caching with the new decorator
- Removed `run_as_async`. I think it's out of scope of this library

### DIFF
--- a/esp_data/dataset/shard_creator.py
+++ b/esp_data/dataset/shard_creator.py
@@ -24,7 +24,7 @@ from tqdm.auto import tqdm
 import esp_data.file_io.functional as F
 from esp_data.config import DatasetConfig
 from esp_data.config.project_config import default_shard_creator_cfg
-from esp_data.paths import AnyPath, make_storage_options
+from esp_data.paths import AnyPath
 from esp_data.utils import make_id
 
 logger = logging.getLogger("esp_data")
@@ -522,7 +522,7 @@ def write_huggingface_shard(
     # Create shard path
     output_path = AnyPath(output_path)
     shard_path = output_path / (f"{shard_name}_{shard_id:06d}")
-    storage_options = make_storage_options(output_path) if storage_options is None else storage_options
+    storage_options = output_path.storage_options if storage_options is None else storage_options
 
     # Process the data using generator for memory efficiency
     total_samples = len(batch)
@@ -1005,7 +1005,7 @@ def create_sharded_dataset(
 
     # Save updated metadata if requested
     if save_metadata_as and len(metadata_df) > 0:
-        storage_options = make_storage_options(output_path)
+        storage_options = AnyPath(output_path).storage_options
         save_metadata(metadata_df, output_path / save_metadata_as, storage_options)
 
     # Report final statistics
--- a/esp_data/paths.py
+++ b/esp_data/paths.py
@@ -1,62 +1,16 @@
 """Path definitions that get rid of some issues with the AnyPath type hint."""
 
 import os
-from pathlib import Path
+from pathlib import PosixPath
 from typing import Optional
 
 import cloudpathlib
 from cloudpathlib import GSClient, S3Client, S3Path
 from google.cloud.storage.client import Client as GS_Client_Official
 
-from esp_data.utils import read_gcp_secret
+from esp_data.utils import cached_class_attribute, read_gcp_secret
 
-
-def is_gcs_path(path: str | Path | os.PathLike) -> bool:
-    return str(path).startswith("gs://")
-
-
-def is_s3_path(path: str | Path | os.PathLike) -> bool:
-    return str(path).startswith("s3://")
-
-
-def is_r2_path(path: str | Path | os.PathLike) -> bool:
-    # FIXME: This is a temporary solution
-    return "r2://" in str(path)
-
-
-def is_local_path(path: str | Path | os.PathLike) -> bool:
-    return not (is_gcs_path(path) or is_s3_path(path) or is_r2_path(path))
-
-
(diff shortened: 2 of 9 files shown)


esp_data/utils.py
```diff
@@ -90,21 +85,21 @@ def increment_version(version: str, mode: str = "patch") -> str:
     return f"{major}.{minor}.{patch}"
 
 
-async def run_as_async(func: Callable, new_event_loop: bool = False, **func_kwargs) -> Callable:
-    """Run the function asynchronously.
+# async def run_as_async(func: Callable, new_event_loop: bool = False, **func_kwargs) -> Callable:
```
GaganNarula: was there a reason to remove this function ?

esp_data/utils.py
```diff
@@ -129,3 +124,22 @@ def read_gcp_secret(secret_id: str, version_id: str = "latest", project_id: str
 
     payload = response.payload.data.decode("UTF-8")
     return payload
+
+
+# TODO (milad) add unit tests for this
```
GaganNarula, line 130: maybe lets add a test for this first before I merge

mil-ad: on it

esp_data/paths.py
```diff
@@ -1,62 +1,16 @@
...
-
-
-def make_storage_options(path: str | os.PathLike) -> dict | None:
-    if is_gcs_path(path):
-        return _make_gscs_storage_options()
-    elif is_s3_path(path) or is_r2_path(path):
-        return _make_s3_storage_options()
-    else:
-        return None
+_DEFAULT_GCP_PROJECT = "okapi-274503"
```
GaganNarula, line 13: again I think we'll have to deal with this differently if we want this lib to be part of some open source code (do we ?), or if its deployed in a container / app in some other project..

esp_data/dataset/shard_creator.py
```diff
@@ -522,7 +522,7 @@ def write_huggingface_shard(
     # Create shard path
     output_path = AnyPath(output_path)
     shard_path = output_path / (f"{shard_name}_{shard_id:06d}")
-    storage_options = make_storage_options(output_path) if storage_options is None else storage_options
+    storage_options = output_path.storage_options if storage_options is None else storage_options
```
GaganNarula, line 525: this change is great !

esp_data/utils.py
mil-ad: Done (thank you Cursor!)

esp_data/utils.py
mil-ad: It wasn't used anywhere and I'm not sure if we need it. I'm also not sure about the usage, is `func` an async function here? if yes how is this function different from `asyncio.run(func())`?

esp_data/paths.py
mil-ad: Agreed. The same thing applies to secrets. To me the "paths" part of this library is mostly meaningful and useful for internal esp people (at least for now) so if we can get away with not asking our users to define a bunch of environment variables we should do that

esp_data/utils.py
GaganNarula: well the idea was to use it as a decorator for a non-async "func" and turn that into a background task. `asyncio.run()` executes awaitables. You can then use it in classes where some of the methods are sequential and some concurrent etc. Would be nice to keep for now if thats ok with you.. I want to test it later for things like fetching data from multiple buckets (better than running multiple processes)

GaganNarula:
#8 
Couple of points I wanted to add:

1. In the last PR we removed a utility function `make_file_opener`. Now I find that it was necessary because: `pathlib.Path` doesn't allow writing to a file where the parent directories dont exist, while "GSPath" and "R2Path" do allow. To reproduce:
```
with AnyPath("/<local path that exists>/<new folder that doesn't exit>/temp").open("w") as f:
    f.write(b"peep")
# Raises FileNotFoundError
```

2. Some issues are appearing re `esp_data.dataset.shard_creator.write_arrow_shard`. The `pq.ParquetWriter` object created in `create_iterative_writer` method was failing to actually write the shard to a given bucket location, but it did not throw any error when the `writer.write_table(table)` statement was executed. I was not able to consistently reproduce this error, because it functioned correctly when writing to local disk (on my laptop). I found the issue while generating _some_ but not all components of the voxaboxen dataset with the output path being a cloud storage location. I think it has something to do with AnyPath, some client issue perhaps.

esp_data/utils.py
mil-ad: I'll keep it for now

esp_data/paths.py
mil-ad: I don't think I need to fix this in this PR. Have raised #11 so we can track

mil-ad:
> Some issues are appearing re esp_data.dataset.shard_creator.write_arrow_shard. The pq.ParquetWriter object created in create_iterative_writer method was failing to actually write the shard to a given bucket location, but it did not throw any error when the writer.write_table(table) statement was executed. I was not able to consistently reproduce this error, because it functioned correctly when writing to local disk (on my laptop). I found the issue while generating some but not all components of the voxaboxen dataset with the output path being a cloud storage location. I think it has something to do with AnyPath, some client issue perhaps.

This is on my radar. I think I'm going to switch to gcsfs (first the non FS variant) in my next PR and wait for your feedback?

GaganNarula (approved):
lgtm

PR #12, opened 2025-04-03
Milad refactor 4

mil-ad:
A LOT of renaming and moving things around to make the scope of the path/fs a bit more clearer. Getting ready to remove `use_fs` and actually use gcsfs/s3fs for most things


GaganNarula:
Could we remove a couple of notebooks in this PR ? (they were placed there for development) 
- [perf_benchmarking_beans0_hfdataset.ipynb](https://github.com/earthspecies/esp-data/blob/main/notebooks/perf_benchmarking_beans0_hfdataset.ipynb)
- [testing_fs_vs_cloudpathlib.ipynb](https://github.com/earthspecies/esp-data/blob/main/notebooks/testing_fs_vs_cloudpathlib.ipynb)

esp_data/io/__init__.py
```diff
@@ -0,0 +1,4 @@
+from esp_data.io.parsers import read_audio
```
GaganNarula, line 1: why is `read_audio` here ? the paths i understand

esp_data/io/functional.py
```diff
@@ -8,17 +8,39 @@
...
-from esp_data.paths import AnyPath, is_local_path, strip_cloud_prefix
-
-from .utils import make_fs
+from .paths import AnyPath, GSPath, R2Path, is_local_path, strip_cloud_prefix
 
 logger = logging.getLogger("esp_data")
 
 
+def _make_gcsfs() -> GCSFileSystem:
+    return GCSFileSystem(access="full_control")
```
GaganNarula, line 21: by the way, I set this access="full_control" here a while back for debugging issues... we might want to check if this allows for too many things like deleting buckets

esp_data/io/parsers.py
```diff
@@ -8,55 +8,45 @@
...
-def read_audio_from_bytes_sf(audio_bytes: bytes) -> tuple[np.ndarray, int]:
+def _read_audio_from_bytes_sf(audio_bytes: bytes) -> tuple[np.ndarray, int]:
     with io.BytesIO(audio_bytes) as audio_buffer:
         data, samplerate = sf.read(audio_buffer)
-
     return data, samplerate
 
 
-def read_audio_from_bytes_pydub(audio_bytes: bytes) -> tuple[np.ndarray, int]:
+def _read_audio_from_bytes_pydub(audio_bytes: bytes) -> tuple[np.ndarray, int]:
```
GaganNarula, line 25: (just realized) soundfile supports reading and writing mp3 so we can remove pydub

tests/test_fileio_functional.py
```diff
@@ -39,7 +33,8 @@ def test_upload_download_cloud(local_test_dir, cloud_path):
     download_target = local_test_dir / "cloud_test_download.bin"
     assert copy(cloud_path, str(download_target)) is True
     assert download_target.read_bytes() == b"Hello Cloud"
-    assert delete_file(cloud_path) is True
+    cloud_path.unlink()
```
GaganNarula, line 36: will we stick with "unlink" for the future.. i find this an unintuitive name but since Pathlib does unlink i suppose it makes sense

GaganNarula (changes requested):
Pydub can be removed and the read_audio can just be done with soundfile. Other comments are just comments, you can make changes as you like.

esp_data/io/functional.py
mil-ad: I think it defaults to full_control anyway but yeah it might make sense to set it to read_write.

esp_data/io/__init__.py
mil-ad: so that we can do `from esp_data.io import read_audio`. We don't have too many exposed functions here so we might as well put them in the `__init__`.

esp_data/dataset/shard_creator.py
```diff
@@ -21,10 +21,10 @@
 from datasets import Dataset
 from tqdm.auto import tqdm
 
-import esp_data.file_io.functional as F
+import esp_data.io.functional as F
```
TParcollet, line 24: This naming could very quickly become confusing if esp_data is used in a bigger PyTorch project. People tend to copy and past import. More generally, it's not really compliant with PEP8 mindset to have a F :p

esp_data/io/functional.py
```diff
@@ -8,17 +8,39 @@
...
+
+
+def _make_cloudflarer2fs() -> S3FileSystem:
+    return S3FileSystem(
```
TParcollet, line 25: Are these env variable properly checked by the S3FileSystem object? If not, it's going to be terrible to debug for users.

esp_data/io/parsers.py
```diff
@@ -8,55 +8,45 @@
...
-from esp_data.paths import AnyPath
-
-UNCOMPRESSED_AUDIO_FORMATS = ["wav", "flac", "ogg"]
-COMPRESSED_AUDIO_FORMATS = ["mp3"]
-
+from .paths import AnyPath
 
 logger = logging.getLogger("esp_data")
 
+UNCOMPRESSED_AUDIO_FORMATS = (".wav", ".flac", ".ogg")
```
TParcollet, line 15: Should be a single variable when we remove pydub.

esp_data/io/parsers.py
```diff
@@ -8,55 +8,45 @@
...
-
+        read_func = _read_audio_from_bytes_pydub
     else:
         raise ValueError(f"Unsupported audio format: {extension}")
 
     return read_func(audio_bytes)
 
 
-def read_audio_bytes_from_path(file_path: AnyPath, fs=None) -> tuple[np.ndarray, int]:
+def read_audio(file_path: AnyPath) -> tuple[np.ndarray, int]:
```
TParcollet, line 43: these two functions should support reading chunk (start / stop indexing) as it may be of use to people.

esp_data/io/parsers.py
mil-ad: fixed in #18

esp_data/dataset/shard_creator.py
mil-ad: you mean the `F` part or the `.io` part?

esp_data/io/functional.py
mil-ad: #18 gets rid of env vars. The whole idea of this io part is to make it easy for the user. We can read them from gcloud secrets so they don't need to define any env vars.

esp_data/io/parsers.py
mil-ad: I have got rid of pretty much all other readers in #18. We might need them at some point in the future but we'll cross that bridge when we get to it. I think it makes sense to keep the audio as it's our main thing.

raised #19 for chunking

cc @GaganNarula

esp_data/dataset/shard_creator.py
TParcollet: @mil-ad Well IN THEORY both, but F is much more problematic than io in this case imho.

mil-ad: we definitely don't want to have a `pip install io`, we're not huggingface 😁

This is just a module inside our package so should be okay?

PR #16, opened 2025-04-08
Apply template

TParcollet:
This is just a WIP PR to apply the template. It will most likely need to wait for Milad to trim the repository (to avoid documenting/formating things that will be removed). 

- [x] Merge with template
- [x] Make CI run
- [x] Fix the tests


requirements.txt
mil-ad, line 1: was this intentional?

TParcollet: Nope

TParcollet:
@mil-ad and specially @GaganNarula this PR is ready to be reviewed. For context, it is trying to remove everything that we don't want to release at first to then add them back one after the other in future PR (and adapting them to the new OSS rules as well). I have not touched the scripts and notebooks folder, but maybe we should do something about it? Also @mil-ad the automatic pypi upload seems to fail. Maybe we just want to deactivate this for now?

esp_data/io/filesystem.py
```diff
@@ -16,8 +18,49 @@
...
+    specified protocol. It supports Google Cloud Storage ('gcs', 'gs'),
+    Cloudflare R2 ('r2'), and the local filesystem ('local').
+
+    For the 'r2' protocol, it automatically retrieves the necessary credentials
+    (access key ID, secret access key, endpoint URL) from GCP Secret Manager.
+
+    The results are cached so subsequent calls with the same protocol and keyword
+    arguments will return the identical filesystem instance.
+
+    Parameters
```
GaganNarula: I thought we were using "Arguments" instead of "Parameters" ? at least that's whats in the repo template..

TParcollet: Good catch! I got trapped by an other example I took .... I'll fix this. thanks.

TParcollet: Fixed!

GaganNarula:
I have a different branch "apply_repo_template" which contains the dev code you are removing in this PR, so its fine from my side. We can also remove the scripts folder, but I would keep the notebooks for now. One of them is about huggingface datasets, and the other is the tutorial for the esp-data repo which we can update as we progress.

mil-ad:
> the automatic pypi upload seems to fail. Maybe we just want to deactivate this for now?

I think it's fine. It's failing because this version already exists in the repo. I'm going to change the trigger later anyway

mil-ad (approved):
Thank you! This is great

GaganNarula (approved):
LGTM

PR #18, opened 2025-04-08
simpler fs

mil-ad:
- add a `filesystem()` factory function that uses caching and secrets. I don't love mixing File-like and filesystem. Thinking if there's a simpler way to do this.
- Also added a `filesystem_from_path()` helper method method 
- Renamed `functional.py` to `filesystem.py`
- Removed non-audio parsers. Out of scope, at least for now
- Removed `yield_files()` 

Tests are still not fully passing but I'm leaving them in a better state than I found them:

image


mil-ad:
I ran into #21 in the mean time

mil-ad:
@GaganNarula can I merge this?

esp_data/io/filesystem.py
```diff
@@ -0,0 +1,54 @@
+"""This is a functional implementation of fileio using the FileSystem approach.
```
GaganNarula: this docstring is misleading, we can remove it i think (did we add having module docstrings to the repo template ? )

mil-ad: removed it. Not sure if the template mentions module docstrings but planning to update docstrings later when the API is a bit more final

scripts/zip_audios.py
GaganNarula, line 1: Can we delete this script ? It was buggy anyway

mil-ad: done

tests/test_fileio_functional.py
```diff
@@ -24,14 +23,17 @@ def test_upload_download_cloud(local_test_dir, cloud_path):
...
-    assert copy(local_file, cloud_path) is True
+
+    assert cloud_path.exists() is False
+    filesystem_from_path(cloud_path).put(str(local_file), str(cloud_path))
+    assert cloud_path.exists() is True
+
     # Download back to a different local file
     download_target = local_test_dir / "cloud_test_download.bin"
-    assert copy(cloud_path, str(download_target)) is True
+    filesystem_from_path(cloud_path).get(str(cloud_path), str(download_target))
```
GaganNarula, line 36: I think we want to simplify this.. especially having to call str() every time, as well as the name "get" and "put"

mil-ad: yeah I'm thinking about it. I don't like the str conversion.

`get()` and `put()` are not bad. It follows fsspec

GaganNarula (approved):
LGTM!

PR #20, opened 2025-04-08
Enable dynamic reading of audio files

### LINKED ISSUE #19: support chunks in read_audio()

TParcollet:
Fixes #19.

- [x] basic code
- [x] Tests (need one more with real audio file) 

@mil-ad / @GaganNarula can we not name the file having reading functions "parsers"? This sounds so confusing to me :/ (I can make the change in this PR if ok).


GaganNarula:
@TParcollet  what would you name it ? I chose parsers because it is "parsing" byte arrays -> some media format ... what do you propose ?

TParcollet:
@GaganNarula to me parsers == ArgParse ahah. This file litteraly only contains audio reading functionalities now, so why not just read_audio.py, unless we plan to have other reading capabilies here? Then it could be read_utils.py?

GaganNarula:
read_utils.py is better because we could extend this to more than just audio (like images and video and biologger files etc.)

mil-ad (approved):

PR #23, opened 2025-04-11
Added anypath and repurposed AnyPath

mil-ad:
* Added `anypath()` function factory and turned `AnyPath` into a type alias instead

### DIFF
--- a/esp_data/dataset/webds.py
+++ b/esp_data/dataset/webds.py
@@ -1,7 +1,7 @@
 import json
 import logging
 from functools import lru_cache, partial
-from typing import Any, Callable, Generator, Literal, Union
+from typing import Any, Callable, Generator, Literal
 
 import pandas as pd
 import webdataset as wds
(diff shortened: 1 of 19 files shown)


esp_data/dataset/webds.py
```diff
@@ -1,15 +1,15 @@
...
+from typing import Any, Callable, Generator, Literal
 
 import pandas as pd
 import webdataset as wds
 
 import esp_data.io.filesystem as F
 from esp_data.config import DatasetConfig
 from esp_data.config.project_config import default_webds_loader_cfg
-from esp_data.io import AnyPath
+from esp_data.io import AnyPath, anypath
```
GaganNarula: I wonder if the type hint and the actual function aren't confusing because one is just the lower case of the other ? Would it maybe be better to call the type hint "PathLike" ?

mil-ad: `PathLike` is a module in `os` but maybe `AnyPathT`?

mil-ad: or even `AnyPathType`

GaganNarula: not sure, though I prefer AnyPathT. Maybe i'm being too pedantic here

mil-ad: I'll change to `AnyPathT` for now. `T` is pretty common when programming generics

mil-ad: done

GaganNarula (approved):
looks good!

PR #24, opened 2025-04-11
Build and publish to private index

mil-ad:
Adds a github action to build the package and push it to our private index


TParcollet (approved):
LGTM! (We should not forget to have this only on release next, not sure why we want to keep it like that for development?)

GaganNarula (approved):
lgtm to me too!

mil-ad:
> We should not forget to have this only on release next, not sure why we want to keep it like that for development?

yeah I just added a task to asana to track this

PR #25, opened 2025-04-11
Setting up mkdocs

mil-ad:
- Starting to set up the docs

also:
- removed `scripts` dir
- push to esp pypi only when there's a release


GaganNarula (approved):
Looks good ! Docs is a major upgrade 😀

PR #26, opened 2025-04-16
docs versioning

mil-ad:
Started using mike for docs versioning.


.github/workflows/docs.yml
```diff
@@ -28,12 +34,19 @@ jobs:
...
-          path: site/
-          retention-days: 7
+      - name: Configure Git user
+        run: |
+          git config --local user.name "${{ github.actor }}"
+          git config --local user.email "${{ github.actor }}@users.noreply.github.com"
+
+      - name: Deploy main branch
+        if: github.ref == 'refs/heads/main'
+        run: uv run mike deploy -b docs-site --push --update-aliases dev
```
GaganNarula: @mil-ad just trying to understand this.. what's the difference between `uv run mike deploy -b docs-site --push --update-aliases dev`  and `uv run mike deploy -b docs-site --push --update-aliases $VERSION latest` ? one is the dev docs and the other is the stable docs ? .. also where is the $VERSION env variable set / updated ? is it the same as the package version in pyproject.toml ?

mil-ad: so I'm thinking of having docs versions for:
- all official releases (we'll tag then with `vX.Y.Z` and that's one scenario triggering the gh action` 
- the main branch under `dev`.

I'm not sure if having the main branch docs is useful. What do you think?

TParcollet: I do not think it is useful as long as the releases and dev are covered.

.github/workflows/docs.yml
```diff
@@ -1,16 +1,22 @@
...
+name: Build the docs
 
+# TODO: test this logic
 on:
   workflow_dispatch:
-  release:
-    types: [published]
+  push:
+    branches:
+      - main
```
TParcollet, line 7: Shouldn't we actually add dev here if the doc is covering releases and dev? (Also, should we start working with a dev branch as well ? like a proper 'develop' one?)

mil-ad: I think I should change that name in the mkdocs. by `dev` I basically meant `main`.

.github/workflows/docs.yml
mil-ad: by `dev` I really meant the `main` branch. But still not sure if there's much point in having docs include "unreleased" stuff. I thought it'd be useful when we're preparing the release but I guess we can always build and serve the docs locally for testing

mil-ad: I simplified the PR by removing this logic for now

GaganNarula (approved):
lgtm, (i'm guessing we dont deploy docs for the "develop" branch ?)

TParcollet:
@GaganNarula In theory, and if it was automatic ... we should. A simple drop down button in the doc can tell the user what version they are looking at. But since it's manual .... well certainly better to not do that ahah.

Issue #28, opened 2025-04-24
Inconsistent globing with `anypath`


GaganNarula:
It seems running `list(anypath(<some cloud path>).rglob("*txt"))` has different behavior for Cloudflare R2 paths vs  google cloud storage paths. For R2 it works correctly, (although using `**/*txt` as a pattern generates a different, strange result), for GCS it fails with error:

```
TypeError                                 Traceback (most recent call last)
Cell In[49], line 12
      7 folder.is_file()
      8 # False
      9 
     10 # Recursively list all files with the pattern "*txt" in the name
---> 12 list(folder.rglob("*txt"))

File [~/esp_projects/esp-data/.venv/lib/python3.10/site-packages/cloudpathlib/cloudpath.py:540](http://localhost:8888/lab/tree/notebooks/~/esp_projects/esp-data/.venv/lib/python3.10/site-packages/cloudpathlib/cloudpath.py#line=539), in CloudPath.rglob(self, pattern, case_sensitive)
    535 pattern_parts = PurePosixPath(pattern).parts
    536 selector = _make_selector(
    537     ("**",) + tuple(pattern_parts), _posix_flavour, case_sensitive=case_sensitive
    538 )
--> 540 yield from self._glob(selector, True)

File [~/esp_projects/esp-data/.venv/lib/python3.10/site-packages/cloudpathlib/cloudpath.py:501](http://localhost:8888/lab/tree/notebooks/~/esp_projects/esp-data/.venv/lib/python3.10/site-packages/cloudpathlib/cloudpath.py#line=500), in CloudPath._glob(self, selector, recursive)
    500 def _glob(self, selector, recursive: bool) -> Generator[Self, None, None]:
--> 501     file_tree = self._build_subtree(recursive)
    503     root = _CloudPathSelectable(
    504         self.name,
    505         [],  # nothing above self will be returned, so initial parents is empty
    506         file_tree,
    507     )
    509     for p in selector.select_from(root):
    510         # select_from returns self.name[/...](http://localhost:8888/...) so strip before joining

File [~/esp_projects/esp-data/.venv/lib/python3.10/site-packages/cloudpathlib/cloudpath.py:496](http://localhost:8888/lab/tree/notebooks/~/esp_projects/esp-data/.venv/lib/python3.10/site-packages/cloudpathlib/cloudpath.py#line=495), in CloudPath._build_subtree(self, recursive)
    493         continue
    495     nodes = (p for p in parts)
[...]

PR #29, opened 2025-04-25
io module docs

GaganNarula:
Feedback welcome!


docs/io.md
```diff
@@ -0,0 +1,289 @@
...
+
+The `io` module provides a set of functions for reading and writing data to and from various file formats and storage systems. It supports local files, Google Cloud Storage (GCS), and Cloudflare R2 buckets.
+
+You can use this module to:
+
+1.  Conduct file operations like "exists", "unlink" (delete), "open" (read/write), and "glob" (list files).
+2.  Download and upload files (and folders) to and from cloud storage locations
+3.  List the contents of buckets
+4.  Delete files
+5.  Read audio from a remote file directly into memory as a numpy array
```
TParcollet: Maybe add a precision to be 100% clear that this isn't like 'streaming'.

docs/io.md
```diff
@@ -0,0 +1,289 @@
+# The `esp_data.io` module
+
+The `io` module provides a set of functions for reading and writing data to and from various file formats and storage systems. It supports local files, Google Cloud Storage (GCS), and Cloudflare R2 buckets.
+
+You can use this module to:
+
+1.  Conduct file operations like "exists", "unlink" (delete), "open" (read/write), and "glob" (list files).
```
TParcollet: Can we add link to the parts of the documentation where exemples for these functions are given? People will not read the documentation as a tutorial (maybe a few of them will), most of them will come here looking for how to do what they want.

docs/io.md
```diff
@@ -0,0 +1,289 @@
...
+
+```python
+anypath("random.json").unlink()
+```
+
+#### Cloud files
+
+The above operations work identically for files in buckets.
+
+⚠️ One important difference between cloudpaths like `GSPath` and `S3Path` and local `PosixPath` is that buckets dont have actual "folders" like your local filesystem. So if you try to create a *local* file remember that parent folders of that file need to exist. This **NOT** true for a file like `gs://some_bucket/some_folder/some_file.txt`, you can directly write to it with `anypath("gs://some_bucket/some_folder/some_file.txt").open("w")`.
```
TParcollet: This **is** NOT ..

docs/io.md
```diff
@@ -0,0 +1,289 @@
...
+```
+    [S3Path('s3://esp-ci-cd-tests/esp-data-tests/find_files_tests/file_find_cloud.txt'),
+     S3Path('s3://esp-ci-cd-tests/esp-data-tests/temprandomfolder/file1.txt'),
+     S3Path('s3://esp-ci-cd-tests/esp-data-tests/temprandomfolder/file2.txt'),
+     S3Path('s3://esp-ci-cd-tests/esp-data-tests/temprandomfolder/random.txt'),
+     ...]
+
+
+
+Download all the contents in cloud "folder"
```
TParcollet: This is confusing to me, is it a download or an upload? It reads like an upload to the cloud

docs/io.md
```diff
@@ -0,0 +1,289 @@
...
+
+You can use this module to:
+
+1.  Conduct file operations like "exists", "unlink" (delete), "open" (read/write), and "glob" (list files).
+2.  Download and upload files (and folders) to and from cloud storage locations
+3.  List the contents of buckets
+4.  Delete files
+5.  Read audio from a remote file directly into memory as a numpy array
+
+There are two interfaces available:
```
TParcollet, line 15: I think that we could put a bit more context as to why these two interfaces. What are the differences and what are they intended for. It reads a bit like a list of things available but if people are not familiar with they won't get straight away what they can achieve (or when they should use it).

TParcollet (commented):
Thanks @GaganNarula this looks good overall to me! I only have a few minor comments.

docs/io.md
mil-ad: we can say if you're dealing with one file use one, and if you're anything that involves copying etc use the other.

mil-ad: we should also do a warning box saying this API might change in the next version and we're trying to consolidate it into one interface

GaganNarula: i'll make more changes thanks. One other thing: the globbing behavior of the `anypath` class has a bug (see issue #28 ) so I think i'd also add a note that the filesystem approach works is a better choice for that.

mil-ad (approved):

PR #38, opened 2025-05-30
esp_data_temp -> esp-data

mil-ad:
- 🐳 Copied/intgegrated `esp_data_temp` from [representation-learning repo](https://github.com/earthspecies/representation-learning) and adapt
- Copied pre-commit hooks from the template
- 🟢 made pre-commit happy
  - added `.yamllint` and `.dict-allowed.txt`
  - added dependency exports
- 🟢 Made CI happy
- Random fixes
  - `register_dataset` decorator and the registry had wrong type hints. `type[Dataset]` is different than `Dataset`
  - Removed `Self` in type-annotations. That needs a very recent python version and this library should be conservative about its assumptions

left to do:

- [x] lots of type errors in the code I copied. I'm surprised pre-commit didn't catch them. They need fixing
- [ ] we should use the io module for all the cloud/file operations
- [ ] Our pyproject doesn't fully match the template (e.g. 120 char lines instead of 88)


esp_data/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+    DatasetInfo,
+    dataset_from_config,
+    list_registered_datasets,
+    print_registered_datasets,
+    register_dataset,
+)
+
+__all__ = [
+    "dataset_from_config",
+    "Dataset",
```
GaganNarula, line 13: I thought we weren't adding `Dataset` to the package level ?

esp_data/datasets/animalspeak.py
```diff
@@ -0,0 +1,277 @@
...
+        license="unknown",
+    )
+
+    def __init__(
+        self,
+        split: str = "train",
+        output_take_and_give: dict[str, str] = None,
+        sample_rate: int = 16000,
+        audio_path_col: str = "gs_path",
+        postprocessors: Optional[List[Callable[[Dict[str, Any]], Dict[str, Any]]]] = None,
```
GaganNarula: lets add a TODO here to figure out what to do with these postprocessors because they dont work with `DatasetConfig` in `dataset_from_config`... seems to me that postprocessors are Transforms ?

GaganNarula (approved):
lgtm for merge

esp_data/__init__.py
mil-ad: Individual datasets go into `esp_data.datasets` but common classes and configs are at the package level. But I'm very open to this. Let me know what you think

esp_data/datasets/animalspeak.py
mil-ad: raised an issue. Might be more useful if we're moving from asana to github projects

mil-ad: I raised some tiny issues and assigned to you btw (they're like tiny TODOs)

esp_data/dataset_to_be_deprecated.py
```diff
@@ -0,0 +1,310 @@
+import pathlib
```
TParcollet: This should be removed imho. No contamination in esp_data.

esp_data/datasets/animalspeak.py
TParcollet: I think that we should remove it for now and add the feature back if requests arise.

esp_data/datasets/animalspeak.py
```diff
@@ -0,0 +1,277 @@
...
+import numpy as np
+import pandas as pd
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.io import GSPath, read_audio
+
+
+@register_dataset
+class AnimalSpeak(Dataset):
+    """AnimalSpeak dataset.
```
TParcollet, line 16: Let's adopt a formalism for all datasets to follow and put in the docstring. Otherwise the generated doc will be empty. We basically need to put the DatasetInfo in there, it's sub optimal due to redundancy but necessary for documentation.

esp_data/datasets/animalspeak.py
```diff
@@ -0,0 +1,277 @@
...
+            if pd.isna(v):
+                return []
+            elif isinstance(v, str):
+                return [item.strip() for item in v.split(",")]
+            else:
+                raise ValueError(f"Expected a string or NaN, but got {v} of type {type(v)}")
+
+        # TODO: Maybe we want to normalise the values even more? for instance apply
+        # .lower()?
+        self._data.background_species_sci = self._data.background_species_sci.apply(_to_list)
```
TParcollet: I don't think that we want that. This should be done during data generation and not at the Dataset level. Also why the hard coded path below?

esp_data/transforms/label_from_feature.py
```diff
@@ -0,0 +1,66 @@
...
+        self.label_map = label_map
+        self.override = override
+        self.output_feature = output_feature
+
+    @classmethod
+    def from_config(cls, cfg: LabelFromFeatureConfig) -> "LabelFromFeature":
+        return cls(**cfg.model_dump(exclude=("type")))
+
+    def __call__(self, df: pd.DataFrame) -> tuple[pd.DataFrame, dict]:
+        if self.output_feature in df and not self.override:
```
TParcollet, line 81: We should not have todos in released code imho. We either document it properly in an issue and remove it from the code, or just remove it from the code.

esp_data/transforms/multilabel_from_features.py
```diff
@@ -0,0 +1,122 @@
...
+    label_map : dict[str, int] | None, default=None
+        A mapping of unique values to integer IDs. If not provided, the transform will
+        generate a mapping based on the unique values in the specified feature columns.
+    output_feature : str, default="label"
+        The name of the output column to store the generated label lists.
+    override : bool, default=False
+        If False and the output_feature already exists in the dataset, an error is
+        raised. If True, the output_feature will be overwritten.
+
+    Methods
```
TParcollet, line 50: Doctests.

esp_data/transforms/label_from_feature.py
```diff
@@ -0,0 +1,66 @@
...
+
+class LabelFromFeatureConfig(BaseModel):
+    type: Literal["label_from_feature"]
+    feature: str
+    label_map: dict[str, int] | None = None
+    output_feature: str = "label"
+    override: bool = False
+
+
+class LabelFromFeature:
```
TParcollet, line 20: Docstring + doctest.

esp_data/transforms/subsample.py
```diff
@@ -0,0 +1,130 @@
...
+    @field_validator("ratios")
+    @classmethod
+    def is_in_range(cls, ratios: dict[str, float]) -> dict[str, float]:
+        if not all(0 <= r <= 1 for r in ratios.values()):
+            raise ValueError("All ratios must be in [0, 1]")
+        return ratios
+
+
+class Subsample:
+    """Subsample data based on property ratios."""
```
TParcollet: Docstring + doctest.

esp_data/transforms/subsample.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if "other" in self.ratios:
+            mask_other = ~df[self.property].isin(self.ratios.keys() - {"other"})
+            idx_other = df.index[mask_other].tolist()
+            chosen_other = self._choose_keys(idx_other, self.ratios["other"])
+            groups.append(df.loc[chosen_other])
+
+        return pd.concat(groups, ignore_index=True)
+
+    # def _subsample_dict(self, data: dict[str, Any]) -> dict[str, Any]:
```
TParcollet: remove.

esp_data/transforms/uniform_sample.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+# TODO (Gagan) I'm a bit confused by what UniformSample is ... it sounds like a subset
+# of Subsample just that the ratios "fixed" internally to make the dataset uniformly
+# distributed across a certain property. But here a scalar "ratio" is being provided
+# which would apply to all the unique groups in a property, and hence will not
+# distribute uniformly.
+
+
+class UniformSample:
```
TParcollet, line 33: Doctstring + doctests

esp_data/transforms/uniform_sample.py
```diff
@@ -0,0 +1,101 @@
...
+            n_samples = max(1, int(len(group) * self.ratio))
+            # TODO is this the right way to set up the random seed? Do we want to fix it
+            # here?
+            rng = np.random.default_rng(seed=42)
+            sampled_indices = rng.choice(len(group), size=n_samples, replace=False)
+            groups.append(group.iloc[sampled_indices])
+
+        return pd.concat(groups, ignore_index=True)
+
+    # def _uniform_sample_dict(self, data: dict[str, Any]) -> dict[str, Any]:
```
TParcollet: remove

TParcollet (changes requested):
Aside from the comments. We need to really carefully think about how we will document this to the others. This range from using transforms to requesting a new dataset. In particular the idea of adding things to this library will be a massive problem. I don't think that we can expect researchers which are immersed in their research to come and create an issue for adding x or y transform, same for the dataset, they may not want to wait. We must find a way to explain this process.

GaganNarula (approved):
🚀

TParcollet (approved):

PR #44, opened 2025-06-05
New datasets

GaganNarula:
New datasets:

+ InsectSet459
+ Beans
+ BarkleyCanyonDetection (this needs to be debugged though)

+ tests and datasets documentation.
+ `data_root` argument added to the datasets which allows a user to change the location of a dataset (e.g. if the dataset is moved from the bucket to a disk). This should be replaced with a `from_disk` method in v2.2.

Have not yet fixed the docstring issues in transforms.


esp_data/datasets/animalspeak.py
```diff
@@ -1,22 +1,38 @@
...
 
+    A part of NatureLM training and BioLingual, AnimalSpeak,
+    as over a million audio-caption pairs holding information on
+    species, vocalization context, and animal behavior
+    Original publication: https://arxiv.org/pdf/2308.04978
+
+    TRANSFERABLE MODELS FOR BIOACOUSTICS WITH HUMAN LANGUAGE SUPERVISION
+    Robinson et al 2023
+
+    Num samples = 2,045,492
```
TParcollet: I guess that this formalism is good for now, but i'm wondering how we should store these infos. Imho, we should have a webpage somewhere with these infos automatically obtained from requesting a database. Maybe we should not have hard values like that in the code, what do you think?

esp_data/datasets/barkley_canyon.py
```diff
@@ -1,29 +1,38 @@
...
+    Excerpt from the original Abstract:
+    "...This dataset, which is being made publicly available for further use,
+    includes strong-label annotations of phonations from blue whales, fin whales,
+    humpback whales, sperm whales, orcas, Pacific white-sided dolphins,
+    Risso's dolphins, and other delphinids that could not be identified to species.
+    All regional orca communities are represented within the dataset,
+    and phonations are labelled to ecotype and pod level when possible..."
+
+
+    References
```
TParcollet, line 30: See, here we already have a different doctstring between AnimalSpeak and this one. I think we should settle on a single formalism. Like: a paragraph describing the dataset, the reference, the authors, and that's it. I think we should not have anything that can somehow change (like number of rows / version / licence) in the docstring. Only immutable things.

esp_data/datasets/beans.py
```diff
@@ -0,0 +1,315 @@
...
+            # For JSONL files, read them directly into a DataFrame
+            self._data = pd.read_json(location, lines=True, orient="records")
+        else:
+            # Read CSV content
+            self._data = pd.read_csv(
+                location, keep_default_na=False, na_values=[""]
+            )  # This setting avoids setting 'None' to a pd.NA type
+
+    @classmethod
+    def from_config(cls, cfg: DatasetConfig) -> "Beans":
```
TParcollet, line 174: Well, maybe this one should be already implemented in Dataset?

TParcollet (approved):
Minor change to be done on the docstring of the datasets, maybe, but otherwise it can be merged IMHO

esp_data/datasets/beans.py
GaganNarula: i didnt follow what you meant here.. what should be already implemented ? this class method `from_config` ? i think one issue with putting this in Dataset directly is that the parameters for constructing different dataset objects may not always be the same.. (I think this allows for flexible structures)

esp_data/datasets/barkley_canyon.py
GaganNarula: yeah i agree, I'll try and clean it up and follow a structure for these four datasets

esp_data/datasets/beans.py
TParcollet: Yes, this is what I meant, but you are also correct (which is why it was not implemented). Maybe better like that after all.

esp_data/transforms/uniform_sample.py
```diff
@@ -31,7 +31,21 @@ def is_in_range(cls, ratio: float) -> float:
 
 
 class UniformSample:
-    """Uniformly sample data based on a property."""
+    """Uniformly sample data based on a property.
+
+    This transform uniformly samples a DataFrame based on a specified property and a
+    ratio. It is SUPPOSED 🚨 to ensure that the resulting DataFrame has a
```
TParcollet, line 37: This supposed is unclear --> Is it doing it, or not? The docstring should be clear on that.

TParcollet (approved):
@GaganNarula See my only comment. Once fixed, I think we can just press merge and move forward with the documentation (I will).

esp_data/transforms/uniform_sample.py
GaganNarula: i'm not sure we should change this docstring right now. I'll raise an issue in representation-learning (from where the logic of this transform comes) and when we actually have to use it we'll fix the docstring and add a rigorous test for this (which currently lack). The outcome of this transform _should_ be to balance a dataset distribution. The 🚨 symbol just helps highlight the uncertainty here... I suggest merging now and let Milad decide in his branch whether to remove this transform or we fix it

PR #45, opened 2025-06-06
Documentation for V2

TParcollet:
This PR contains cosmetic changes as well as new pages for the V2 documentation. It does not offer "tutorials" maybe we should think about that. 

To do:
- [x] transforms
- [x] datasets

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -1,0 +1,337 @@
+# `esp_data.datasets` module
+
+## What are ESP Datasets?
+
+The datasets module provides a collection of **datasets validated by the engineering team** for ESP projects. In short, this is the module to use to download, load and manipulate *official* ESP datasets. Each dataset is implemented as a class that inherits from the base `Dataset` class, providing a consistent interface for data loading and access.
+
+More technically an ESP Dataset is defined as such:
+- Inherits from the base `Dataset` class
+- Has a defined `DatasetInfo` containing metadata
+- Provides methods for loading and accessing data and splits
+- Can be configured through a `DatasetConfig`
+
+## How to Load Datasets?
+
+Datasets can be loaded following two different approaches:
+
+1. Direct instantiation:
+```python
+from esp_data.datasets import AnimalSpeak
+
+# Create a dataset instance
+dataset = AnimalSpeak(
+    split="validation"
+)
+
+# Access data
+sample = dataset[0]  # Get first sample
+print(len(dataset))  # Get dataset size
+```
+
+2. Using configuration:
+```python
+from esp_data import DatasetConfig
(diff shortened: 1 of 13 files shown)


pyproject.toml
```diff
@@ -50,7 +50,7 @@ python_files = [
 ]
 
 [tool.ruff]
-line-length = 88
+line-length = 100
```
GaganNarula, line 53: this seems to have changed again ? i thought we are sticking to 88 ?

esp_data/transforms/filter.py
```diff
@@ -67,14 +71,22 @@ def from_config(cls, cfg: FilterConfig) -> "Filter":
...
+        ----------
+        data : pd.DataFrame
+            The data dataframe to filter.
+
+        Returns
+        -------
+        tuple[pd.DataFrame, dict]
+            A tuple containing:
+            - The filtered data (same type as input)
+            - An empty dictionary for metadata
```
GaganNarula, line 84: Seems strange to return an empty dictionary because we have to follow the `transformed_data, metadata = transform(dataset)` pattern... also, `Filter` doesn't allow a metadata dict as an argument on construction so its a bit inconsistent there..

docs/datasets.md
```diff
@@ -0,0 +1,337 @@
...
+    property="source",
+    values=["xeno-canto", "iNaturalist"],
+    mode="include"
+)
+label_transform = LabelFromFeature(
+    feature="canonical_name",
+    output_column="label"
+)
+
+dataset.apply_transformations([filter_transform, label_transform])
```
GaganNarula, line 80: Issue here: currently we are supposed to pass in the TransformConfigs as a list not the transforms themselves.. I prefer the latter way (instantiate transforms outside the dataset and pass them in) but it doesnt work with the code as it looks right now.

esp_data/transforms/filter.py
mil-ad: returning an empty dict doesn't seem too strange to me. But yeah I don't like the inconsistency. I don't think it's a blocker for release though.

mil-ad: raised #48 for tracking this

pyproject.toml
mil-ad: I'm finding 88 a bit limiting but it's what we have in the template. Let's make the release and then change back to 88? It would need some manual intervention

docs/datasets.md
mil-ad: Noted. Let's merge and fix this separately

GaganNarula (approved):
good to go, lets remember to fix the doc issue for transforms in the milad/copy.. branch

PR #49, opened 2025-06-10
Birdset dataset

nkundiushuti:
Adding the birdset dataset


esp_data/datasets/birdset.py
```diff
@@ -0,0 +1,283 @@
...
+        if anypath(location).suffix == ".jsonl":
+            # For JSONL files, read them directly into a DataFrame
+            self._data = pd.read_json(location, lines=True, orient="records")
+        else:
+            # Read CSV content
+            self._data = pd.read_csv(
+                location, keep_default_na=False, na_values=[""]
+            )  # This setting avoids setting 'None' to a pd.NA type
+        if "train" in self.split:
+            self._data = self._data.sample(frac=0.8, random_state=42)
```
GaganNarula: i think you were describing this in the meeting today, but I forgot the purpose. Why are we taking a smaller fraction here ? Also, if its always 0.8, maybe just take the first 80% of the dataframe rather than sampling ? This assumes that the data is already pre-shuffled once..

esp_data/datasets/birdset.py
```diff
@@ -0,0 +1,283 @@
...
+        elif "validation" in self.split:
+            self._data = self._data.sample(frac=0.2, random_state=42)
```
GaganNarula: same here

docs/datasets.md
```diff
@@ -0,0 +1,343 @@
...
+
+# Create dataset configuration with transforms
+config = DatasetConfig(
+    dataset_name="animalspeak",
+    split="validation",
+    transformations=[filter_config, label_config]
+)
+
+# Create dataset with transforms
+dataset = AnimalSpeak.from_config(config)
```
GaganNarula, line 118: hmm i think this is a bug. As of now, the transformations are not applied automatically on instantiation. i'll fix this in my next patch

pyproject.toml
```diff
@@ -39,20 +44,18 @@ allow-direct-references = true
...
   ".", "esp_data",
 ]
+python_files = [
+    "test_*.py",
+    "check_*.py",
+]
 
 [tool.ruff]
-line-length = 120
+line-length = 88
```
GaganNarula: this is strange.. in `main` branch, the line-length is 100... why is this 88 ?

esp_data/datasets/birdset.py
```diff
@@ -0,0 +1,283 @@
...
+            raise IndexError(
+                f"Index {idx} out of bounds for dataset of length {len(self._data)}."
+            )
+
+        row = self._data.iloc[idx].to_dict()
+        # Ensure audio path is valid
+        if self.data_root:
+            audio_path = anypath(self.data_root) / row["path"]
+        else:
+            audio_path = anypath(row["path"])
```
GaganNarula, line 212: "path" is fine as the name, although i use "local_path" everywhere else. As long as the combination of `self.data_root` and `row["path"]` works, we're happy :)

esp_data/datasets/birdset.py
nkundiushuti: we are replicating the DT experimental setup in Section 4.1 in the paper https://arxiv.org/abs/2403.10380 they mention they take 20% of the data across different random seeds. maybe we just return the whole train set as the validation and we do the subsampling in the representation_learning because this is where we set the seeds. otherwise we have to give seed as a parameter to the dataset object.

pyproject.toml
nkundiushuti: maybe I have to rebase again. I haven't touched these files

esp_data/datasets/birdset.py
GaganNarula: Yes i agree, lets not do this here, would be better to use the UniformSample or Subsample transforms in representation-learning directly

pyproject.toml
GaganNarula: Can you change this to 100 ? Thanks!

GaganNarula (changes requested):
Couple of things to change:

1. Remove subsampling in BirdSet
2. Line-length pyproject.toml

GaganNarula (approved):
good to merge!

PR #51, opened 2025-06-11
allowing negatives in multilabel

david-rx:
During evaluation on multilabel datasets, for instance on BEANS detection, we need to include the negative examples, i.e. those which have no labels. I'm assuming this will be true for other multilabel cases as well, though excluding them by e.g. a transform may be common. Examples were previously being dropped by .dropna() since this _row_to_ids previously returned None.

### DIFF
--- a/esp_data/transforms/multilabel_from_features.py
+++ b/esp_data/transforms/multilabel_from_features.py
@@ -46,6 +46,9 @@ class MultiLabelFromFeatures:
     override : bool, default=False
         If False and the output_feature already exists in the dataset, an error is
         raised. If True, the output_feature will be overwritten.
+    allow_missing_labels : bool, default=True
+        If True, rows with no labels will be included in the output. If False, rows with
+        no labels will be dropped.
 
     Methods
     -------
@@ -63,11 +66,13 @@ class MultiLabelFromFeatures:
         label_map: dict[str, int] | None = None,
         output_feature: str = "label",
         override: bool = False,
+        allow_missing_labels: bool = True,
     ) -> None:
         self.features = features
         self.label_map = label_map
         self.override = override
         self.output_feature = output_feature
+        self.allow_missing_labels = allow_missing_labels
 
     @classmethod
     def from_config(cls, cfg: MultiLabelFromFeaturesConfig) -> "MultiLabelFromFeatures":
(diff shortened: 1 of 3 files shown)


esp_data/transforms/multilabel_from_features.py
```diff
@@ -98,10 +98,7 @@ def _row_to_ids(row: pd.Series) -> list | None:
                     v = [row[f]]
                 row_labels.extend(map(lambda x: label_map[x], v))
 
-            if row_labels:
-                return sorted(row_labels)
-            else:
-                return None
+            return sorted(row_labels)
```
GaganNarula, line 107: hi david, is this still a draft PR ? If not then i'm not sure how this diff is changing the behavior of this transform... anyway, do add a unittest for this behavior (of keeping negative labels) in the `tests/test_multilabel_from_features.py`. 

Also, I suggest that keeping negative labels should be a config parameter and used in construction of MultiLabelFromFeatures

mil-ad:
We should probably be explicit about this behaviour and choose it via a config

esp_data/transforms/multilabel_from_features.py
david-rx: Not a draft -> this changes a None return to return an empty list, causing the row to be no longer dropped by the subsequent .dropna(). I am happy to add a test for it and make it a config parameter

GaganNarula: ok sorry there's obviously a difference there! :D Could you do the following:

1. Add a parameter to the MutliLabelFromFeatures transform and MutliLabelFromFeaturesConfig that allows returning empty lists. I'm not sure what to call it, `allow_missing_labels` ? (negative labels has a different connotation I feel because that could also mean labels from other samples in the data that do not correspond to this sample (e.g. like in SimCLR)
2. Add a test for it in the corresponding unit test file.
3. Add a note in the docs/transforms.md in the section for `MultiLabelFromFeatures transform` about allowing this behavior.

GaganNarula (changes requested):
Need to add a parameter to control for this behavior

esp_data/transforms/multilabel_from_features.py
david-rx: Great, will do

GaganNarula (approved):
LGTM!

Issue #46, opened 2025-06-09
Query sr remotely in .io module


benjaminsshoffman:
is it easy to add a function to the `.io` module to grab an audio file’s sr remotely? The ability to read a subset of frames is great but I don’t know the sr’s of all my files a priori.

GaganNarula:
Added support for reading a subset of frames using start_time and end_time without explicitly knowing sample_rate. this is in `esp_data.io.read_utils.read_audio`

PR #52, opened 2025-06-12
New audio read methods and fixes to dataset.from_config

### LINKED ISSUE #39: Obsolete fields in DatasetConfig
Do we need `audio_path_col`, `multi_label`, `metrics` field in `DatasetConfig`? any point in them?

GaganNarula:
Release v1.0.1

- [x] `Dataset.from_config` now applies transforms if they are included in the DatasetConfig.
- [x]  Added support for reading audio between start and end_times and getting audio info #46 
- [x]  Updated docs
- [x] Small fixes #39 #49 (BirdSet is added from a previous PR)


esp_data/io/read_utils.py
```diff
@@ -155,3 +155,135 @@ def audio_stereo_to_mono(
         return np.mean(audio, axis=0)
     else:
         raise ValueError(f"Unsupported mono conversion method: {mono_method}")
+
+
+def get_audio_info(
```
TParcollet, line 176: I am not sure why we need this function while read_audio() already can do that? I find it confusing to have multiple read_audio functions. Do you see any specific use case for not using the read_audio functionality for that?

TParcollet (commented):
Thanks @GaganNarula ! Honestly nothing to say aside from the remark. I've tested, it works. Tests are ok.

TParcollet (approved):
LGTM!

Issue #59, opened 2025-06-16
Allow int:int label mapping


nkundiushuti:
In some cases the values we want to map are not strings but ints, so in these cases( egyptian fruit bats from BEANS) we would like to have an int to int mapping
```label_map: dict[str, int] | None = None```

GaganNarula:
solved in #62

PR #62, opened 2025-06-16
Small fixes

GaganNarula:
- [x] An indexing error was wrongly raised for negative python array indexing (e.g. `x = [1,2,3]; x[-1]` is allowed in python
- [x] The split_path validation in `DatasetInfo` is now removed. Because of how datasets get "registered" when one is imported, all split_path validations occur which slows down imports a lot. The validation is now moved to unit tests.
- [x] Small doc string fixes that were causing problems when running build_docs.yml CI
- [x] Fix for issue #59

### DIFF
--- a/esp_data/dataset.py
+++ b/esp_data/dataset.py
@@ -5,7 +5,6 @@ from typing import Any, Dict, Iterator, Optional
 import semver
 from pydantic import BaseModel, ConfigDict, Field, field_validator
 
-from esp_data.io import anypath
 from esp_data.transforms import RegisteredTransformConfigs, transform_from_config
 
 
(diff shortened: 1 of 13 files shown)


esp_data/dataset.py
```diff
@@ -65,7 +64,7 @@ def convert_none(cls, v: Any) -> Any:  # noqa: ANN401
 class DatasetInfo(BaseModel):
     """A Pydantic base model for the info (cfg) of a dataset.
 
-    Parameters
+    Attributes
```
TParcollet, line 67: Parameters

TParcollet (approved):
@GaganNarula Only one minor issue (see Arguments -> Parameters)

esp_data/dataset.py
GaganNarula: actually keeping `Parameters` here (in DatasetInfo and DatasetConfig) is causing an issue with building the docs. Its fixed by changing it to `Attributes`

Issue #55, opened 2025-06-16
Concatenate datasets


GaganNarula:
a function that concatenates two or more datasets like so:

```
from esp_data.dataset import Dataset, concatenate_datasets

ds1 = AnimalSpeak()
ds2 = BarkleyCanyon()
new_ds  = concatenate_datasets([ds1, ds2], merge_strategy=....)

assert type(new_ds) == Dataset
assert len(new_ds) == len(ds1) + len(ds2)

```

PR #68, opened 2025-06-19
Concatenate datasets 

GaganNarula:
- [x] #55 We can now concatenate datasets using the `esp_data.concat` module and `concatenate_datasets` function. This is v1 of this approach and has some disadvantages, most obviously extra memory use.
- [x] DatasetConfig allows extra parameters. This gives a little more freedom to the use to pass parameters for experiments that are dataset specific.
- [x] Added a small transform `Deduplicate` to remove duplicates from datasets.
- [x] Docs for concatenate_datasets and deduplicate.
- [x] Unit tests

### DIFF
--- a/docs/concatenate.md
+++ b/docs/concatenate.md
@@ -1,17 +1,26 @@
 # `esp_data.concat` Module
 
-## What is Dataset Concatenation?
+## What is Dataset concatenation?
 
 The `concat` module provides utilities for **combining multiple ESP datasets** into a single unified dataset. This is particularly useful when you want to train models on data from multiple sources or combine different splits of related datasets while maintaining proper data handling and metadata.
 
 More technically, dataset concatenation:
+
 - Combines multiple `Dataset` objects into a single `ConcatenatedDataset`
 - Preserves original dataset functionality through source dataset references
 - Handles column mismatches through configurable merge strategies
 - Maintains proper metadata and configuration merging
 - Tracks data provenance for debugging and analysis through a merged `DatasetInfo`
 
-## How to Concatenate Datasets?
+### A note on dataset merging
+
+In v1 of this approach, we have are three [merge strategies](#merge-strategies). Currently, because we use `pandas` DataFrames under the hood, the merge strategies are:
+
+1. **Soft Merge**: Keeps all columns from all datasets, filling missing values with NaN.
+2. **Overlap Merge**: Keeps only columns that exist in all datasets.
+3. **Hard Merge**: Requires all datasets to have identical columns, raising an error if they differ.
+
+## How can I concatenate datasets?
 
 Datasets can be concatenated using the `concatenate_datasets` function with different merge strategies:
 
@@ -43,13 +52,13 @@ print(f"First sample: {sample.keys()}")
 print(f"Last sample: {combined_dataset[-1].keys()}")
 ```
 
-!!! warning
+!!! remark
     The `ConcatenatedDataset` class *currently* cannot be instantiated from a `DatasetConfig`. It is designed to be created directly from existing `Dataset` objects. So trying `ConcatenatedDataset.from_config(config)` will raise an error.
 
 
 #### Apply transformations before / after concatenation
 
-You can apply transformations to the individual datasets before concatenation as shown in [transforms.md](transforms.md). This allows you to treat the data as needed, but you can also apply transformations after concatenation if you want to operate on the combined dataset as a whole. Here is an example of applying a filter transformation *after* concatenation:
+You can apply transformations to the individual datasets *before* concatenation as shown in [transforms.md](transforms.md). This allows you to treat the data as needed, but you can also apply transformations *after* concatenation if you want to operate on the combined dataset as a whole. Here is an example of applying a filter transformation *after* concatenation:
 
 ```python
 from esp_data.datasets import AnimalSpeak, BarkleyCanyon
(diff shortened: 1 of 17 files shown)


docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
+# `esp_data.concat` Module
+
+## What is Dataset Concatenation?
+
+The `concat` module provides utilities for **combining multiple ESP datasets** into a single unified dataset. This is particularly useful when you want to train models on data from multiple sources or combine different splits of related datasets while maintaining proper data handling and metadata.
+
+More technically, dataset concatenation:
+- Combines multiple `Dataset` objects into a single `ConcatenatedDataset`
+- Preserves original dataset functionality through source dataset references
+- Handles column mismatches through configurable merge strategies
```
TParcollet, line 11: Maybe an anchor to the documentation about the merge strategy

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+- Maintains proper metadata and configuration merging
+- Tracks data provenance for debugging and analysis through a merged `DatasetInfo`
+
```
TParcollet, line 14: Maybe a short sentence about merging strategies at the top (and anchors)

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+filter_config = FilterConfig(
+    type="filter",
+    property="species_common",
+    values=["American Robin", "Bottle-nosed Dolphin"],
+    mode="include"
+)
+
+# Run the transformation on the combined dataset
+transform_metadata = combined_dataset.apply_transformations([filter_config])
+```
```
TParcollet, line 84: If i understand correctly, this will just pick one dataset samples if the other ones does not have the column named 'species_common' as their value will be unset, right?

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+
+```python
+# Hard merge - strictest option
+combined_dataset = concatenate_datasets(
+    [dataset1, dataset2],
+    merge_level="hard"
+)
+```
+
+### Advanced Usage with Transforms
```
TParcollet: This should be with the doc on Transforms after concatenation, right?

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+# This will raise MergeException if sample rates differ
+try:
+    incompatible_dataset = concatenate_datasets([audio_16k, audio_44k])
+except MergeException as e:
+    print(f"Sample rate mismatch: {e}")
+```
+
+### Output Column Mapping
+
+The `output_take_and_give` mappings are merged and validated:
```
TParcollet, line 235: think that this could use a longer text to explain what this is exactly, I dont think we have a piece of documentation for that yet.

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+
+```python
+from esp_data.datasets import AnimalSpeak
+
+# Create datasets with compatible column mappings
```
TParcollet, line 240: You wanted to write 'incompatible' I believe?

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+    output_take_and_give={"canonical_name": "different_name"}  # Conflict!
+)
+
+try:
+    bad_combined = concatenate_datasets([dataset1, dataset3])
+except MergeException as e:
+    print(f"Mapping conflict: {e}")
+```
+
+## Error Handling
```
TParcollet: Not sure if that is necessary.

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+
+### 1. Choose the Right Merge Strategy
+
+- Use **soft merge** when datasets have different but complementary columns
+- Use **overlap merge** when you only need common features across datasets
+- Use **hard merge** when datasets should have identical schemas
+
+### 2. Validate Before Concatenation
+
+```python
```
TParcollet, line 286: I think that this should be done automatically before merging right? Why do we ask the user to do it?

docs/concatenate.md
```diff
@@ -0,0 +1,366 @@
...
+    # Process in chunks to manage memory
+    for i in range(0, len(combined), chunk_size):
+        chunk = [combined[j] for j in range(i, min(i + chunk_size, len(combined)))]
+        # Process chunk
+        yield chunk
+```
+
+### 4. Preserve Data Lineage
+
+```python
```
TParcollet: I can't understand what this is about from the documentation

esp_data/concat.py
```diff
@@ -0,0 +1,431 @@
...
+        split_paths=merged_split_paths,
+        version=merged_version,
+        description=merged_description,
+        sources=merged_sources,
+        license=merged_license,
+        changelog=merged_changelog,
+    )
+
+
+def concatenate_datasets(
```
TParcollet, line 253: We want an exemple in the docstring for testing  (and documentation purpose).

esp_data/concat.py
```diff
@@ -0,0 +1,431 @@
...
+    finally:
+        # Restore original output_take_and_give mappings to source datasets
+        for dataset, original_otag in zip(datasets, original_otags, strict=False):
+            dataset.output_take_and_give = original_otag
+
+
+class ConcatenatedDataset(Dataset):
+    """A dataset created by concatenating multiple datasets.
+
+    This dataset maintains references to the original datasets to enable
```
TParcollet, line 352: I'd say -> docstring exemple as well here.

esp_data/dataset.py
```diff
@@ -46,6 +46,13 @@ class DatasetConfig(BaseModel):
 
     """
 
+    model_config = ConfigDict(
```
TParcollet, line 49: What is this for?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "esp-data"
-version = "1.0.2"
+version = "1.1.0"
```
TParcollet: Should this be done in this PR?

TParcollet (changes requested):
Thanks for the work! I cannot test the code properly the code while in the train, so i'll test it next week, but I reviewed it.

pyproject.toml
GaganNarula: well otherwise will have to make another PR for bumping the version for the release

esp_data/dataset.py
GaganNarula: I believe ConfigDict is a parameter set used by pydantic.BaseModel. I added this to have the `extra="allow"` parameter set.

pyproject.toml
TParcollet: Which is good practice I think!

docs/concatenate.md
GaganNarula: Added some more info about `output_take_and_give` to `datasets.md`..

TParcollet (approved):
tests are passing locally and the changes look good! Thanks

PR #70, opened 2025-06-20
adding stowell datasets

david-rx:
Datasets from: https://royalsocietypublishing.org/doi/10.1098/rsif.2018.0940

- Chiffchaff ID
- Pipit ID
- Little Owl ID


GaganNarula (changes requested):
more documentation on split paths, then we're good!

GaganNarula (approved):
🚀

PR #71, opened 2025-06-20
Add zebra finch julie elie dataset

nkundiushuti:
Added the dataset from here: https://figshare.com/articles/dataset/Vocal_repertoires_from_adult_and_chick_male_and_female_zebra_finches_Taeniopygia_guttata_/11905533/1


scripts/zebra_finch_julie_elie_010.py
GaganNarula, line 1: I think we wanted to keep scripts/ in a different branch for now

.dict-allowed.txt
```diff
@@ -3,8 +3,7 @@
 ### Common Words for use in Compound Words ###
 ### Compound Words With 1 or 2 letter Words ###
 ### Jargon ###
+Te
```
GaganNarula: Where is this used ?

scripts/zebra_finch_julie_elie_010.py
nkundiushuti: ah true, sorry

.dict-allowed.txt
nkundiushuti: it's one of the call types in the dataset, it should also be part of the other branch

nkundiushuti:
I excluded the preprocessing scripts, if you want to have another look

.dict-allowed.txt
GaganNarula, line 1: @nkundiushuti oh is "Te" not necessary ? it looks like you removed it ?

esp_data/datasets/zebra_finch_julie_elie.py
```diff
@@ -0,0 +1,263 @@
...
+    >>> dataset = ZebraFinchJulieElie(
+    ...     split="test",
+    ...     output_take_and_give={"label": "label"},
+    ...     sample_rate=16000,
+    ...     data_root="gs://esp-ml-datasets/zebra_finch_julie_elie/v0.1.0/raw/"
+    ... )
+    """
+
+    info = DatasetInfo(
+        name="Zebra Finch Julie Elie",
```
GaganNarula: can we change this to lower case ? `zebra_finch_julie_elie`. The reason is that the `name` attribute is used for choosing datasets in a config. So its more like a parameter than just a name..

GaganNarula (changes requested):
looks good, just needs one change in the name

esp_data/datasets/zebra_finch_julie_elie.py
nkundiushuti: ok I renamed it

GaganNarula (approved):
Lgtm!

PR #75, opened 2025-06-23
bengalese finch repertoire

david-rx:
Add the bengalese finch songs dataset https://figshare.com/articles/media/BirdsongRecognition/3470165/1

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -175,6 +175,13 @@ The list of available dataset will grow over time. Please refer to the next sect
             - "!^.*$"  # Exclude everything
         show_source: true
 
+::: esp_data.datasets.BengaleseFinchCalls
+    handler: python
+    options:
+        filters:
+            - "!^.*$"  # Exclude everything
+        show_source: true
+
 ## Using your own dataset
 
 First of all, you must answer an important question: is this new dataset relatively stable and time and potentially useful to others? If yes, then you **should talk to the engineering team** to add it as an official ESP Dataset. If not, you can just follow the next steps!
--- a/esp_data/datasets/bengalese_finch_calls.py
+++ b/esp_data/datasets/bengalese_finch_calls.py
@@ -3,7 +3,8 @@
 This dataset contains individual calls from Bengalese finches. Each row in the
 metadata CSV corresponds to a *single call* extracted as an audio snippet.
 The dataset is organized by individual birds, with each bird having its own
-split containing its vocal repertoire.
+split containing its vocal repertoire. Note that the repertoires
+are per-bird, so labels should not be compared across splits.
 
 The data is hosted in the `esp-ml-datasets` GCS bucket in folder bengalese_finch with:
 - Individual CSV files per bird (Bird0.csv, Bird1.csv, etc.)
@@ -20,7 +21,7 @@ Examples
 >>> ds = BengaleseFinchCalls(split="Bird0", sample_rate=16000)
 >>> first = ds[0]
 >>> first.keys()
- dict_keys(['local_path', 'call_type', 'individual_id', 'audio'])
+dict_keys(['local_path', 'call_type', 'individual_id', 'audio'])
 """
 
 from typing import Any, Dict, Iterator, Optional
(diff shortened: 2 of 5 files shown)


docs/datasets.md
```diff
@@ -183,13 +183,20 @@ The list of available dataset will grow over time. Please refer to the next sect
             - "!^.*$"  # Exclude everything
         show_source: true
 
+::: esp_data.datasets.BengaleseFinchCalls
+    handler: python
+        options:
```
GaganNarula: hi david, this identation here at `options` is throwing an error in doc building. the `options` field should be at the same indentation as `handler`, while the `filters` and `source` should be one tab more (like it is).

docs/datasets.md
```diff
@@ -183,13 +183,20 @@ The list of available dataset will grow over time. Please refer to the next sect
...
+            filters:
+                - "!^.*$"  # Exclude everything
+            show_source: true
 
 ::: esp_data.datasets.GiantOtters
     handler: python
         options:
```
GaganNarula: same here, was fixed in main already so you have a conflict

GaganNarula (changes requested):
A couple of minor changes needed (see comments) and a conflict needs to be resolved.

esp_data/datasets/bengalese_finch_calls.py
```diff
@@ -0,0 +1,253 @@
...
+            "Bird1": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird1.csv",
+            "Bird2": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird2.csv",
+            "Bird3": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird3.csv",
+            "Bird4": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird4.csv",
+            "Bird5": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird5.csv",
+            "Bird6": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird6.csv",
+            "Bird7": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird7.csv",
+            "Bird8": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird8.csv",
+            "Bird9": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird9.csv",
+            "Bird10": "gs://esp-ml-datasets/bengalese_finch/v0.1.0/raw/Bird10.csv",
```
GaganNarula, line 55: would it make sense to have a full split where each birds call is present with the name attached to it ? e.g. Bird0_call_type_1

david-rx: Would that be a split for each label? How would you see it being used?

GaganNarula: I mean just one split called "full" (or "eval" or something), which combines all splits, so that the repertoire size is the union of the sets of individual repertoires and the sample size is large ? would also avoid individually loading each bird split as a separate dataset during eval

david-rx: Oh got it. Yeah I could see that being useful if you wanted to train on everything. I'm not sure how much overlap there is between the repertoires - i.e. it may make sense only to learn to distinguish between Bird0_call_type_1 and Bird0_call_type_2, but possibly Bird0_call_type_1 is roughly Bird1_call_type_5 and we don't want to attempt learning to distinguish them?

That said, I could see it being useful down the line, happy to add the split if you think so too

GaganNarula: actually lets table this for another time when the use case appears. You're right, it may be that the bird calls do overlap and learning distinctions between them would be counterproductive. I will merge as is!

GaganNarula:
One more change needed: add the dataset import to the `esp_data/__init__.py`

david-rx:
> One more change needed: add the dataset import to the `esp_data/__init__.py`

added

GaganNarula (approved):
lgtm

PR #79, opened 2025-06-25
Orchive dataset

nkundiushuti:
the annotated part of orchive is quite small (I was not aware)
they provide, however, a quite large unannotated dataset with cropped calls

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -191,6 +191,13 @@ The list of available dataset will grow over time. Please refer to the next sect
             - "!^.*$"  # Exclude everything
         show_source: true
 
+::: esp_data.datasets.Orchive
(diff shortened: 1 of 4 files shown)


david-rx:
Hey @nkundiushuti @GaganNarula @mil-ad - what do we think of including this in the end? For vocal repertoire, we don't need a training set, clustering/retrieval are the primary evals for many repertoire datasets anyways. And, this would be our hardest, most diverse repertoire task and only marine one. Maybe we could just add a split for "all data" since we will probably not train on it and we want to use everything, and maybe a split with all data but having at least 3 examples per class?

Lmk what you think and happy to help with that if you agree

EDIT - hold that thought, I think there is a much better one

PR #80, opened 2025-06-26
Bug fix in dataset registry

GaganNarula:
Need to import all datasets otherwise registry doesn't work and can't run `dataset_from_config` in another application like representation-learning

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "esp-data"
-version = "1.1.0"
+version = "1.1.1"
 description = "Data tools for ESP projects"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "esp-data"
-version = "1.1.0"
+version = "1.1.1"
```
TParcollet, line 3: If doing this you need to make a release on github, right?

TParcollet (approved):

pyproject.toml
GaganNarula: yup

PR #82, opened 2025-06-29
Small transform fix

GaganNarula:
1. Fixes a bug in `label_to_feature` when a feature with the same name as `output_feature` is being overwritten by a given label map
2. Assertion error change in `multilabel_from_feature`. (Does this do what its supposed to do ?)

### DIFF
--- a/esp_data/transforms/label_from_feature.py
+++ b/esp_data/transforms/label_from_feature.py
@@ -93,7 +93,13 @@ class LabelFromFeature:
         else:
             label_map = self.label_map
 
-        df_clean.loc[:, [self.output_feature]] = df_clean[self.feature].map(label_map)
+        if self.output_feature in df_clean:
+            # This will overwrite the existing feature if it exists
+            df_clean.loc[:, self.output_feature] = df_clean[self.feature].map(label_map)
+        else:
+            # Doing this avoids the SettingWithCopyWarning, i.e.
+            # "A value is trying to be set on a copy of a slice from a DataFrame."
+            df_clean.loc[:, [self.output_feature]] = df_clean[self.feature].map(label_map)
 
         metadata = {
             "label_feature": self.feature,
--- a/esp_data/transforms/label_from_feature.py
+++ b/esp_data/transforms/label_from_feature.py
@@ -93,6 +93,10 @@ class LabelFromFeature:
         else:
             label_map = self.label_map
 
+        # We use assign here because it allows us to create the
+        # self.output_feature column if it doesn't exist, or overwrite
+        # it if it does, without needing to use .copy() or
+        # .loc with different syntax's depending on overwrite or not.
         df_clean = df_clean.assign(**{self.output_feature: df_clean[self.feature].map(label_map)})
 
         metadata = {
--- a/esp_data/transforms/label_from_feature.py
+++ b/esp_data/transforms/label_from_feature.py
@@ -93,13 +93,7 @@ class LabelFromFeature:
         else:
             label_map = self.label_map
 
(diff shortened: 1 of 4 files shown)


esp_data/transforms/label_from_feature.py
```diff
@@ -93,7 +93,13 @@ def __call__(self, df: pd.DataFrame) -> tuple[pd.DataFrame, dict]:
...
             label_map = self.label_map
 
-        df_clean.loc[:, [self.output_feature]] = df_clean[self.feature].map(label_map)
+        if self.output_feature in df_clean:
+            # This will overwrite the existing feature if it exists
+            df_clean.loc[:, self.output_feature] = df_clean[self.feature].map(label_map)
+        else:
+            # Doing this avoids the SettingWithCopyWarning, i.e.
+            # "A value is trying to be set on a copy of a slice from a DataFrame."
+            df_clean.loc[:, [self.output_feature]] = df_clean[self.feature].map(label_map)
```
mil-ad: isn't it better to change `df_clean = df.dropna(subset=[self.feature])` to `df_clean = df.dropna(subset=[self.feature]).copy()` and then simply do `df_clean[self.output_feature] = df_clean[self.feature].map(label_map)` instead of the `loc` thing? what do you think? I think I did the loc because .dropna() returns a view

mil-ad: also whatever we end up doing we should copypasta to multilabel too?

GaganNarula: i can try that .. but does calling copy() actually create a copy ? twice the memory use ?

mil-ad: valid concern.

mil-ad: but I still don't understand why you have to do it like that in line 98. What is different when `output_feature` is in `df_clean`? 🤔

GaganNarula: oh i should have added this to the PR description. If the `output feature` is in `df_clean`, and we run : `df_clean.loc[:, [output_feature]] = df_clean[self.feature].map(label_map)` it throws a `ValueError: setting an array element with a sequence`. If instead we keep it as `df_clean.loc[:, self.output_feature] = df_clean[self.feature].map(label_map)` for both cases (i.e. `output_feature` being overwritten or not) it raises a SettingWithCopyWarning.

GaganNarula: The reason for `ValueError: setting an array element with a sequence` is that it tries to replace a dataframe with a series essentially.

GaganNarula: so it seems pandas copy is default a deep copy `.copy(deep=True)` https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.copy.html#pandas.DataFrame.copy.

mil-ad: ah that makes a lot more sense now thank you. Could you actually add this as a comment before the if-statement.

one last thing we could try to avoid copy() and this if-else is pandas's `assign()`? https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html

probably something like 

```py
df_clean = df_clean.assign(**{self.output_feature: df_clean[self.feature].map(label_map)})
```

GaganNarula: yes this works well, just tested ! looks clean too. I'll switch to this

mil-ad (approved):
Looks good to me. Could you just add a comment above this assign() magic describing why we're doing this instead of something more trivial. I'm sure we're going to forget next week.

PR #85, opened 2025-07-07
Fixes for AudioSet

david-rx:
- Adds documentation for AudioSetCaps captions which were added to AudioSet
- Removes mock data from AudioSet tests
- Converts an added multilabel json string into Python list for labels
- Removes loading by start and end time because the clips are already cropped

### DIFF
--- a/esp_data/datasets/audioset.py
+++ b/esp_data/datasets/audioset.py
@@ -1,5 +1,6 @@
 """AudioSet dataset"""
 
+import json
 from io import StringIO
 from typing import Any, Dict, Iterator, Optional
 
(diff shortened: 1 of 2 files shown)


esp_data/datasets/audioset.py
```diff
@@ -122,7 +129,14 @@ def _load(self) -> None:
         location = self.info.split_paths[self.split]
         # Read CSV content
         csv_text = anypath(location).read_text(encoding="utf-8")
-        self._data = pd.read_csv(StringIO(csv_text))
+
+        # Converter to parse JSON-encoded labels into Python lists
+        def parse_label(value: str) -> list:
```
GaganNarula, line 134: nice! i didnt know you could do that!

tests/test_audioset.py
```diff
@@ -178,314 +118,247 @@ def test_length(dataset: Dataset) -> None:
...
+    sample = dataset_with_output_mapping[0]
+    assert "audio_label" in sample
+    assert "labels" not in sample  # Original column should be filtered out
 
-        # Get a sample and check the mapping
-        sample = dataset_with_output_mapping[0]
-        assert "audio_label" in sample
-        assert "label" not in sample  # Original key should not be present
+    # Audio is NOT automatically included in AudioSet when using output_take_and_give
+    # This is different from some other datasets like BengaleseFinchCalls
```
GaganNarula, line 213: so i think this means we have to change the BengaleseFinchCalls dataset implementation. The idea of `output_take_and_give` was that this is the _exact_ output dictionary expected by the user. If provided, and "audio" is not included in the mapping, then its not there ..

GaganNarula (approved):
looks good!

PR #87, opened 2025-07-09
Adding Voxaboxen datasets

GaganNarula:
Adding two forms of Voxaboxen with the original datasets from the paper:

1. `Voxaboxen`: very simple, returns full `audio` and `selection_table` as dataframe.
2. `VoxaboxenEvents`: follows the process in `DetectionDataset` in https://github.com/earthspecies/voxaboxen/blob/main/voxaboxen/data/data.py with some performance improvements.

🚨This dataset has a *lot* more parameters than other datasets, importantly `scale_factor` which changes the length of audio returned based on the rate at which an embedding model produces sound event annotation output.

### DIFF
--- a/esp_data/__init__.py
+++ b/esp_data/__init__.py
@@ -22,6 +22,8 @@ from .datasets import (
     LittleOwlId,
     MacaquesCooCalls,
     PipitId,
+    Voxaboxen,
+    VoxaboxenEvents,
     ZebraFinchJulieElie,
 )
 
@@ -48,4 +50,6 @@ __all__ = [
     "ZebraFinchJulieElie",
     "BengaleseFinchCalls",
     "MacaquesCooCalls",
+    "Voxaboxen",
+    "VoxaboxenEvents",
 ]
--- a/esp_data/datasets/__init__.py
+++ b/esp_data/datasets/__init__.py
@@ -10,6 +10,7 @@ from .insectset_459 import InsectSet459
 from .littleowl_id import LittleOwlId
 from .macaques_coo_calls import MacaquesCooCalls
 from .pipit_id import PipitId
+from .voxaboxen import Voxaboxen, VoxaboxenEvents
 from .zebra_finch_julie_elie import ZebraFinchJulieElie
 
 __all__ = [
@@ -27,4 +28,6 @@ __all__ = [
     "PipitId",
     "AudioSet",
     "MacaquesCooCalls",
+    "Voxaboxen",
+    "VoxaboxenEvents",
 ]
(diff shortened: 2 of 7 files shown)


david-rx (approved):
Looks great. Good with me if you merge.

A followup discussion: for representation-learning and generally people evaluating on this, I'm now thinking we might use both modes: for model inputs and training, we'll use the "VoxaboxenEvents" version. To score during evaluation, we'll also need the "raw" annotation for each entire file, not the frame-based annotation. To get this, we'll use the "Voxaboxen" form, and connect it back to our predictions made based on the "VoxaboxenEvents." Does that sound right to you?

GaganNarula:
> Looks great. Good with me if you merge.
> 
> A followup discussion: for representation-learning and generally people evaluating on this, I'm now thinking we might use both modes: for model inputs and training, we'll use the "VoxaboxenEvents" version. To score during evaluation, we'll also need the "raw" annotation for each entire file, not the frame-based annotation. To get this, we'll use the "Voxaboxen" form, and connect it back to our predictions made based on the "VoxaboxenEvents." Does that sound right to you?

That's sounds good to me,  you're the expert on usage. One thing that I left out of the VoxaboxenEvents dataset, which seemed relevant for training, was this bit of code in the `__init__` of voxaboxen's DetectionDataset :

```
self.rng = default_rng(seed=self.seed)
....

if self.train:
            self.omit_empty_clip_prob = args.omit_empty_clip_prob
            self.clip_start_offset = (
                self.rng.integers(0, np.floor(self.clip_hop * self.sr)) / self.sr
            )
else:
        self.omit_empty_clip_prob = 0
        self.clip_start_offset = 0
```
it seems that for training, clips with no annotations can be skipped by some probability set by the user, and the clip start can be offset by a random amount. Also, this requires the `seed` to be set.

@david-rx wdyt? do you need this ?

david-rx:
> That's sounds good to me, you're the expert on usage. One thing that I left out of the VoxaboxenEvents dataset, which seemed relevant for training, was this bit of code in the `__init__` of voxaboxen's DetectionDataset :
> 
> ```
> self.rng = default_rng(seed=self.seed)
> ....
> 
> if self.train:
>             self.omit_empty_clip_prob = args.omit_empty_clip_prob
>             self.clip_start_offset = (
>                 self.rng.integers(0, np.floor(self.clip_hop * self.sr)) / self.sr
>             )
> else:
>         self.omit_empty_clip_prob = 0
>         self.clip_start_offset = 0
> ```
> 
> it seems that for training, clips with no annotations can be skipped by some probability set by the user, and the clip start can be offset by a random amount. Also, this requires the `seed` to be set.
> 
> @david-rx wdyt? do you need this ?

I've just asked Benj, and they did not include this in the paper. It could be a nice convenience feature for speed, but not critical for now

PR #88, opened 2025-07-14
Refactor read_audio() function parameter types in read_utils.py

mil-ad:
small fixes in `read_audio()`:
  - switched hints from the legacy `Optional[int]` to `int | None`
  - fixed types for `start_time`, `end_time`, `input_str`, I don't think they can handle `None`

I also raised [an issue](https://github.com/orgs/earthspecies/projects/16/views/2?pane=issue&itemId=119754877) because I think this function's interface is not clear enough


GaganNarula (approved):
👌

Issue #74, opened 2025-06-21
Add exporters for datasets


GaganNarula:
It seems that file IO performance is a bottleneck for training when several experiment runs are trying to access the nfs disk. We might want to first create `tar` or `parquet` or `TFRecord` shards from a training/eval dataset when experiment launches. This could take some time initially, but might be faster when several experiments need to iterate over the same set of shards.

PR #90, opened 2025-07-17
Export to tar and load from tar

GaganNarula:
Issue #74 

- [x] This PR adds a functional api in `esp_data.exporters` that has this public function : `export_as_tar(ds: Dataset, output_path: str | AnyPathT, ... export_options)`
- [x] Needed to add some utility functions for this so created a `dataset_utils.py` module. 
- [ ] Tests for these two ^^ in `tests/test_exporters.py`
- [x] Adds `NatureLMAudio` tar dataset as *Iterable only* (untested, because iterating over the data is tricky since the shards are 1.5 gb each)
- [ ] Seperate refactor issue: moved a lot of the dataset specific unit tests into a separate folder, can revert if its not useful / should be done in another PR.

Once we discuss this design, I can also add a `.save_to_disk(self, output_path, format="tar")` method to `Dataset` class


esp_data/dataset_utils.py
```diff
@@ -0,0 +1,190 @@
...
+
+import numpy as np
+import soundfile as sf
+import webdataset as wds
+
+from esp_data.io import AnyPathT, anypath
+
+
+def audio_decoder(data: dict, dtype: str = "float32", format: str = "FLAC") -> dict[str, Any]:
+    """Decode audio data from a WebDataset sample.
```
TParcollet, line 13: If it is web dataset specific, should we make the name of the function for explicit? wds_audio_decoder?

esp_data/dataset_utils.py
```diff
@@ -0,0 +1,190 @@
...
+        Size of the shuffle buffer.
+    batch_size: int, optional
+        Batch size for processing audio files.
+    shard_shuffle: bool, optional
+        Whether to shuffle the shards.
+    shard_shuffle_size: int, optional
+        Size of the shuffle buffer for shards.
+    split_by_worker: bool, optional
+        Whether to split the dataset by worker.
+    batch_collate_fn: Callable, optional
```
TParcollet, line 216: I am really wondering if we really want to integrate any batching logic to a load function? It sounds a bit counter intuitive to me. Batching should be a dataloader thing, not a dataset thing.

esp_data/exporters.py
```diff
@@ -0,0 +1,521 @@
...
+from esp_data import Dataset
+from esp_data.io import AnyPathT, anypath, filesystem_from_path
+
+from .dataset_utils import audio_encoder
+
+# Configure logging
+logger = logging.getLogger("esp_data")
+
+# Initialize colorama for cross-platform colored terminal output
+colorama.init()
```
TParcollet: Do we really want that? If so, this is something that should be defined at the package level to have something homogeneous across the board. But not sure about it :p

esp_data/exporters.py
```diff
@@ -0,0 +1,521 @@
...
+        logger.error(f"Exception traceback:\n{traceback.format_exc()}")
+    elif error_handling == "raise":
+        raise e
+    elif error_handling == "ignore":
+        pass
+    if isinstance(e, KeyboardInterrupt):
+        raise e
+
+
+def _make_file_opener(
```
TParcollet: Again, if we support multiple format as suggested by docstring, I propose that function names are explicit "webdataset or wds"

esp_data/exporters.py
```diff
@@ -0,0 +1,521 @@
...
+        if compression not in ["gz", "bz2", "xz"]:
+            raise ValueError(
+                f"Unsupported compression format: {compression}. "
+                "Supported formats are 'gz', 'bz2', 'xz'."
+            )
+        # append compression extension if specified
+        shard_path = f"{str(shard_path)}.{compression}"
+
+    # Initialize shard writer
+    sink = wds.ShardWriter(
```
TParcollet: Note here: I think that shards should be based on size. E.g. user can specify the max size in MB and we go from there e.g. 100 shards of 500MB. (Which does not exclude batched writing, it's just to mention that we should support this).

esp_data/exporters.py
```diff
@@ -0,0 +1,521 @@
...
+        shard_name=shard_name,
+        compression=compression,
+    )
+
+
+def export_as_tar(
+    ds: Dataset,
+    output_path: str | AnyPathT,
+    *,
+    num_samples_per_shard: int = 1000,
```
TParcollet, line 310: sample based vs size based, we should talk about that.

esp_data/exporters.py
```diff
@@ -0,0 +1,521 @@
...
+        output_path=output_path,
+        sample_prep_function=sample_prep_function,
+        log_every=log_every,
+        error_handling=error_handling,
+        shard_name=shard_name,
+        compression=compression,
+    )
+
+
+def export_as_tar(
```
TParcollet, line 306: Question, can people shuffle afterwards (at sampling time)? I think it's only possible within the samples in a shard, right?

TParcollet:
@GaganNarula thanks for the work. I like the design. I think it must stay like this and simply receive a dataset object rather than being a dataset class function. I am wondering however if we should make the "map" of shards more easily accessible to people. Huggingface has this nice way of defining the splits where you give the list of shards for each split. This could come handy if someone wants to create smaller splits without using transforms, like just saying take [shard-000  to shard-100] and name it ds["my_small_split"]. It could also help creating subsets very easily without re exporting everything. Like, exporting initially takes the subset mapping of the original dataset, but then one can easily modify it. This would of course means that the dataset class may break afterwards, so maybe not that urgent, but I liked this possibility when creating the different subset of the loquacious set.

esp_data/dataset_utils.py
GaganNarula: agreed

esp_data/exporters.py
GaganNarula: yes its possible to shuffle both within the shard (`shuffle_size` parameter in`esp_data.dataset_utils.load_webdataset`) and shuffle shard order (`shard_shuffle`) at sampling time

esp_data/exporters.py
GaganNarula: yeah i dont really care to include or not, just makes the terminal output pretty while sharding 😄  i do agree that these should be set at a project level or something

esp_data/exporters.py
GaganNarula: i agree this is a good discussion to have. The issue i was facing was this: if we use the `wds.ShardWriter` and specify a max size, it will stop writing to a file after the size is reached, and then we would move on to the next file.. however, in the parallel / chunked write setting which i've done now (otherwise its really too slow for large datasets), its easier to split a dataset up into batch indices before initializing sharding using the `len(ds)` call. If we knew the size in Gb (on disk / bucket) of an entire dataset *before* sharding, and the user decided to evenly split it, we could compute a sensible value for `num_samples_per_shard` and split the indices. 

In a parallelized setting it seems a bit tricky to split by an arbitrary size defined by the user... do you have an idea i could try ? for e.g. create `num_worker` splits, and then within each split let the `wds.ShardWriter` decide the number of shard files to write based on size of each,.. but then there is an issue of communicating the shard ids across workers so there's no overlap..

GaganNarula:
p.s. does anyone have an idea why this test is failing in the CI runner but not on my laptop ? 

in `tests/unittests/datasets_tests/test_naturelm_audio.py`
```
    def test_iteration_over_one_sample(dataset) -> None:
        t0 = time()
>       sample = next(iter(dataset))
```
Error:
```
Traceback:
def wait_for_child(self):
        """Check the status variable and raise an exception if necessary."""
        verbose = int(os.environ.get("GOPEN_VERBOSE", 0))
        if self.status is not None and verbose:
            # print(f"(waiting again [{self.status} {os.getpid()}:{self.proc.pid}])", file=sys.stderr)
            return
        self.status = self.proc.wait()
        if verbose:
            print(
                f"pipe exit [{self.status} {os.getpid()}:{self.proc.pid}] {self.args} {info}",
                file=sys.stderr,
            )
        if self.status not in self.ignore_status and not self.ignore_errors:
>           raise IOError(f"{self.args}: exit {self.status} (read) {info}")
E           OSError: ("((['gsutil', 'cat', 'gs://esp-ml-datasets/naturelm/processed/v0.1.0/tar/train/shard_000000.tar'],), {'bufsize': 8192}): exit 1 (read) {}", <webdataset.gopen.Pipe object at 0x7f287db55ae0>, 'gs://esp-ml-datasets/naturelm/processed/v0.1.0/tar/train/shard_000000.tar')

.venv/lib/python3.10/site-packages/webdataset/gopen.py:92: OSError
----------------------------- Captured stderr call -----------------------------
ServiceException: 401 Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object. Permission 'storage.objects.get' denied on resource (or it may not exist).
```

Seems to run fine on my laptop, but it uses my credentials here..

esp_data/dataset_utils.py
GaganNarula: agreed.. maybe we can keep these params in the `load_webdataset` function but not include them in a `Dataset` init ?

esp_data/datasets/naturelm_audio.py
mil-ad, line 1: why are we testing this with our biggest dataset when testing it tricky anyway? Can't we test this with one of the datasets that repr learning project is using so that we can benchmark the performance?

esp_data/dataset_utils.py
mil-ad, line 1: can we rename this whole file to webdataset_utils.py or something?

mil-ad, line 1: I'm not familiar with webdataset so some noob questions.

- What does a batch mean here? is it different from a shard?
- Does this file have be audio/json aware? can't we make it lower level and audio-agnostic?

esp_data/exporters.py
mil-ad: I don't think we want this, especially in the middle of some random file.

We could have a `logging_utils.py` where we create logging Formatter objects that are prettier 🤷

esp_data/exporters.py
```diff
@@ -0,0 +1,521 @@
...
+        or a file object if the path is local.
+    """
+    file_path = anypath(file_path)
+
+    if isinstance(file_path, Path):
+        parent_dir = file_path.parent
+        parent_dir.mkdir(parents=True, exist_ok=True)
+        return open(file_path, mode=mode)
+
+    fs = filesystem_from_path(file_path)
```
mil-ad: doesn't this cause the same bug I had? should be str

mil-ad (commented):
left some comments

esp_data/dataset_utils.py
```diff
@@ -0,0 +1,190 @@
...
+    Returns
+    -------
+        dict: Dictionary containing the decoded audio data and metadata.
+
+    Raises
+    ------
+    ValueError
+        If the sample does not contain an audio key ending with .flac, .wav, etc.
+    """
+    audio_key = next((k for k in data if k.endswith(f".{format.lower()}")), None)
```
mil-ad, line 31: IMO this makes too many assumptions for the input and output. Maybe instead of passing a dictionary we could do this at a different level

esp_data/dataset_utils.py
```diff
@@ -0,0 +1,190 @@
...
+        The sample to encode
+    indent: int
+        Indentation level for JSON (default: 2)
+
+    Returns
+    -------
+        dict: Dictionary containing the encoded sample in JSON format.
+    """
+    json_data = json.dumps(sample, indent=indent).encode("utf-8")
+    return {"sample.json": json_data}
```
mil-ad, line 119: why do we need this?

esp_data/dataset_utils.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def load_webdataset(
+    path: str | AnyPathT,
+    file_pattern: str = "shard*tar",
+    data_processor: Callable = None,
```
mil-ad: wrong hint

esp_data/dataset_utils.py
```diff
@@ -0,0 +1,190 @@
...
+    shuffle_size: int | None = None,
+    batch_size: int | None = None,
+    shard_shuffle: bool = False,
+    shard_shuffle_size: int = 1000,
+    split_by_worker: bool = False,
+    batch_collate_fn: Callable = None,
```
mil-ad:
```suggestion
    batch_collate_fn: Callable | None = None,
```

mil-ad: wrong hint

esp_data/dataset_utils.py
```diff
@@ -0,0 +1,190 @@
...
+    shard_shuffle: bool, optional
+        Whether to shuffle the shards.
+    shard_shuffle_size: int, optional
+        Size of the shuffle buffer for shards.
+    split_by_worker: bool, optional
+        Whether to split the dataset by worker.
+    batch_collate_fn: Callable, optional
+        Function to collate the batch.
+    seed Union[int, bool, None]:
+        Seed for shuffling. Defaults to True, random seed. If None, means no shuffling!
```
mil-ad, line 219: I find `int | bool | none` confusing. Why mix seed and shuffling?

esp_data/datasets/naturelm_audio.py
GaganNarula: yes but i haven't created a tar archive for any of the repr-learning datasets.. was thinking we'd merge this and then i do it ?

esp_data/dataset_utils.py
GaganNarula: > What does a batch mean here? is it different from a shard?

yes`batch` is different from shard (batch_size can be  >= or <= the size (in samples) of a single shard), `batch` is the same as the idea of a batch in a pytorch DataLoader. As Tituoan said, we dont really want it here, so i dont mind removing these..

> Does this file have be audio/json aware? can't we make it lower level and audio-agnostic?

The data needs to be encoded to binary, and I thought we could have two default ways of doing it: one is agnostic because it just serializes the whole data dict (`json_encoder`) and then encodes it (the decoder just reverses that). We could instead use `pickle` ?

and the other one is where the audio is binarized by soundfile (dtype float32 and flac format save space) while the rest of the data is json dumped..

esp_data/dataset_utils.py
GaganNarula: the input assumption (dict) is based off the `webdataset` library's examples in their docs. Also, our dataset objects return dicts too so it kind of works ? These encoders (`audio_encoder` and `json_encoder`) where meant as reasonable default for encoding the data, assuming that every dataset in esp-data either has an "audio" field (so use `audio_encoder`) or doesn't (`json_encoder`)

GaganNarula:
> @GaganNarula thanks for the work. I like the design. I think it must stay like this and simply receive a dataset object rather than being a dataset class function. I am wondering however if we should make the "map" of shards more easily accessible to people. Huggingface has this nice way of defining the splits where you give the list of shards for each split. This could come handy if someone wants to create smaller splits without using transforms, like just saying take [shard-000 to shard-100] and name it ds["my_small_split"]. It could also help creating subsets very easily without re exporting everything. Like, exporting initially takes the subset mapping of the original dataset, but then one can easily modify it. This would of course means that the dataset class may break afterwards, so maybe not that urgent, but I liked this possibility when creating the different subset of the loquacious set.

I am for this! especially to make a dev split for testing or developing an idea. One way to do this is to specify a regex pattern in the `split_path`, something like `"dev": "gs://path/to/tar/files/shard_001[0-9][0-9][0-9].tar"` would specify files in range `shard_001000.tar` to `shard_001999.tar`. This works directly with the way the `load_webdataset` function works

GaganNarula:
this is now too old

PR #91, opened 2025-07-18
switch to fs for reading

mil-ad:
Another step in moving from cloudpathlib to gcsfs

### DIFF
--- a/esp_data/io/read_utils.py
+++ b/esp_data/io/read_utils.py
@@ -7,6 +7,7 @@ from typing import Any, Literal, Optional
 import numpy as np
 import soundfile as sf
 
+from . import filesystem_from_path
 from .paths import AnyPathT, anypath
 
 logger = logging.getLogger("esp_data")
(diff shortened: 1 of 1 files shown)


esp_data/io/read_utils.py
```diff
@@ -269,7 +271,8 @@ def read_audio_by_time(
         raise ValueError(f"Unsupported audio format: {extension}")
 
     try:
-        with file_path.open("rb") as f:
+        fs = filesystem_from_path(file_path)
+        with fs.open(str(file_path), "rb") as f:
```
TParcollet, line 275: I find it a bit funny that you need to first instantiate an object already with the path and then open it again by specifying the path. I understand that changing that would require making it an actual class, but hum ... ok for now though.

TParcollet (approved):
Tests pass, and code ok.

esp_data/io/read_utils.py
mil-ad: agreed. I think it makes sense for us to eventually define a `open()` method for `AnyPath` that does this filesystem dance under the hood and also deals with this funny conversion

PR #92, opened 2025-07-22
adding WABAD dataset

david-rx:
Adding the WABAD dataset.

Note this is a strongly-labeled dataset, but for now we're not adding a Voxaboxen-style dataset. Instead we iterate through vocalizations with random windows on either side. Planning to add a frame-level variant in a followup PR.

### DIFF
--- a/esp_data/datasets/wabad.py
+++ b/esp_data/datasets/wabad.py
@@ -1,0 +1,499 @@
+"""Wabad dataset"""
+
+from io import StringIO
+from typing import Any, Dict, Iterator, Optional
+
+import librosa
+import numpy as np
+import pandas as pd
+from numpy.random import default_rng
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+
+@register_dataset
+class Wabad(Dataset):
+    """WABAD dataset as individual events.
+
+    Description
+    -----------
+    A global PAM dataset with temporally strong labels for bird vocalizations.
+    Includes 90,662 boxed vocalisations from 1,147 bird species recorded at 70
+    recording sites in 27 countries and distributed across 13 biomes.
+
+    For this dataset variant, each sample corresponds to an individual boxed vocalization.
+    The audio window around each event can be randomly sampled, potentially including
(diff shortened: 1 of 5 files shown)


esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,500 @@
...
+        sources=["Global PAM recordings"],
+        license="CC BY-NC 4.0",
+    )
+
+    def __init__(
+        self,
+        split: str = "master",
+        output_take_and_give: dict[str, str] = None,
+        sample_rate: int = 16000,
+        data_root: Optional[str | AnyPathT] = None,
```
mil-ad, line 113:
```suggestion
        data_root: str | AnyPathT | None = None,
```

No `Optional` please

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,500 @@
...
+        mono_method: str = "average",
+        window_duration: float = 5.0,
+        random_window: bool = True,
+        min_padding: float = 0.0,
+        max_padding: float = 2.0,
+        unknown_label: str = "Unknown",
+        seed: Optional[int] = None,
```
mil-ad, line 120: same here

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,500 @@
...
+            output_take_and_give=cfg.get("output_take_and_give", None),
+            data_root=cfg.get("data_root"),
+            sample_rate=cfg.get("sample_rate", 16000),
+            mono_method=cfg.get("mono_method", "average"),
+            window_duration=cfg.get("window_duration", 5.0),
+            random_window=cfg.get("random_window", True),
+            min_padding=cfg.get("min_padding", 0.0),
+            max_padding=cfg.get("max_padding", 2.0),
+            unknown_label=cfg.get("unknown_label", "Unknown"),
+            seed=cfg.get("seed", None),
```
mil-ad, line 371: we don't want any defaults here. They should be handled in Config class or dataset `__init__`

mil-ad (changes requested):
small comments

david-rx:
I'm closing this for now given the randomness issue. Let's discuss again after the deadline, I think we could support it in a similar way to Voxaboxen and it's a very valuable dataset for the team.

PR #93, opened 2025-07-22
support partial loading of audio from disk

david-rx:
For BEANs and AnimalSpeak, support only loading a segment from disk to reduce IO.

For BEANs we load from the first N seconds by default: all models evaluated on BEANs must truncate to less  than ten seconds anyway, and consider only the start.

For AnimalSpeak we default to a random ten-seconds, which is also used by all models trained on it so far.

### DIFF
--- a/esp_data/datasets/animalspeak.py
+++ b/esp_data/datasets/animalspeak.py
@@ -6,9 +6,10 @@ from typing import Any, Dict, Iterator, Optional
 import librosa
 import numpy as np
 import pandas as pd
+from numpy.random import default_rng
 
 from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
-from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, get_audio_info, read_audio
 
 
 @register_dataset
@@ -30,12 +31,22 @@ class AnimalSpeak(Dataset):
     Examples
     -------
     >>> from esp_data.datasets import AnimalSpeak
+    >>> # Random window with efficient loading
     >>> dataset = AnimalSpeak(
     ...     split="validation",
+    ...     max_duration=5.0,
+    ...     random_window=True,
+    ...     seed=42,
     ...     output_take_and_give={"species_common": "comm"}
     ... )
     >>> print(dataset.info.name)
     animalspeak
+    >>> # Fixed window from start
+    >>> dataset_fixed = AnimalSpeak(
+    ...     split="train",
+    ...     max_duration=10.0,
+    ...     random_window=False
+    ... )
     """
 
     info = DatasetInfo(
@@ -57,6 +68,9 @@ class AnimalSpeak(Dataset):
         output_take_and_give: dict[str, str] = None,
         sample_rate: Optional[int] = None,
         data_root: Optional[str | AnyPathT] = None,
+        max_duration: float = 10.0,
+        random_window: bool = True,
+        seed: Optional[int] = None,
     ) -> None:
         """Initialize the AnimalSpeak dataset.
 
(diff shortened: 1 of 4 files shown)


mil-ad:
do tests pass? 
image

esp_data/datasets/animalspeak.py
```diff
@@ -57,6 +68,9 @@ def __init__(
         output_take_and_give: dict[str, str] = None,
         sample_rate: Optional[int] = None,
         data_root: Optional[str | AnyPathT] = None,
+        max_duration: float = 10.0,
+        random_window: bool = True,
```
mil-ad, line 72: do we want to default to random_window = True?

esp_data/datasets/animalspeak.py
```diff
@@ -57,6 +68,9 @@ def __init__(
...
+        seed: Optional[int] = None,
```
mil-ad, line 73: This synatx is obsolete. Use the new one
```suggestion
        seed: int | None = None,
```

esp_data/datasets/animalspeak.py
```diff
@@ -148,6 +175,9 @@ def from_config(cls, dataset_config: DatasetConfig) -> tuple["AnimalSpeak", dict
             output_take_and_give=cfg.get("output_take_and_give", None),
             data_root=cfg.get("data_root"),
             sample_rate=cfg["sample_rate"],
+            max_duration=cfg.get("max_duration", 10.0),
+            random_window=cfg.get("random_window", True),
+            seed=cfg.get("seed", None),
```
mil-ad, line 180: I don't think there should be any default fallbacks in these `from_config()` factories. If there are any defaults they should be set in the config class. @GaganNarula I don't know why we do `output_take_and_give=cfg.get("output_take_and_give", None)`. We should fix that to `output_take_and_give=cfg.output_take_and_give`.

esp_data/datasets/animalspeak.py
```diff
@@ -201,7 +231,26 @@ def __getitem__(self, idx: int) -> dict[str, Any]:
         else:
             audio_path = anypath(row["local_path"])
 
-        audio, sr = read_audio(audio_path)
+        # Read audio with time limiting for efficiency
+        # Get audio info to determine available duration for both cases
```
mil-ad, line 235: unnecessary comment. code is very clear

mil-ad (changes requested):
left some comments

esp_data/datasets/animalspeak.py
```diff
@@ -73,13 +87,26 @@ def __init__(
...
         """
         super().__init__(output_take_and_give)  # Initialize the parent Dataset class
         self.split = split
         self._data: pd.DataFrame = None
         self._load()  # Load the dataset (fills self._data)
         self.sample_rate = sample_rate
         self.data_root = data_root
+        self.max_duration = max_duration
+        self.random_window = random_window
+        self.seed = seed
```
mil-ad, line 108: is `seed` used for anything other this windowing? if not can we make sure it doesn't have any side effects for other random operations in the codebase and also rename to `random_window_seed` or something to make that super clear.

esp_data/datasets/animalspeak.py
david-rx: I think so? So far, this is the only way the dataset has been used downstream.

GaganNarula: I think default should be False. The issue is we might be using these Dataset classes for things other than training, so I dont think that would be the only way to use them downstream.

GaganNarula:
@david-rx i think we should close this PR because its quite old and esp-data has changed a bit. Also, we don't know yet if want to really allow for random segmentation *within* a dataset. It might be better to create an audio processing transform and apply it as a hook to the Dataset, and use call it while iterating

PR #99, opened 2025-07-23
v1.2 patch release

GaganNarula:


mil-ad (commented):
If we're adding new datasets that's at least a minor release? How about 1.2.0?

mil-ad:
also I noticed that `latest` tag doesn't point to latest release. Maybe fix that in this release too?

mil-ad (approved):
LGTM

PR #102, opened 2025-07-23
David repertoire

david-rx:
Adding proper splits for bengalese finch dataset.


GaganNarula (approved):
looks fine!, lets just be a bit more conservative in the future about the number of splits .. this is a very large number, and I'm sure it has a use case but we should discuss a bit more

Issue #98, opened 2025-07-23
Filtering before concatenating


david-rx:
When using a concatenated dataset, if I run filter or drop_duplicates on one individual dataset prior to concatenating, I get an indexing error. It looks like this:

```
        ^^^^^^^^^^^^^^^^
Loader worker process 0.
Original Traceback (most recent call last):
  File "/scratch/david_earthspecies_org/venvs/11138/lib/python3.12/site-packages/esp_data/concat.py", line 438, in __getitem__
    source_item = source_dataset[source_row_idx]
                  ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/scratch/david_earthspecies_org/venvs/11138/lib/python3.12/site-packages/esp_data/datasets/animalspeak.py", line 194, in __getitem__
    raise IndexError(f"Index {idx} out of bounds for dataset of length {len(self._data)}.")
IndexError: Index 1039279 out of bounds for dataset of length 818136.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/scratch/david_earthspecies_org/venvs/11138/lib/python3.12/site-packages/torch/utils/data/_utils/worker.py", line 351, in _worker_loop
    data = fetcher.fetch(index)  # type: ignore[possibly-undefined]
           ^^^^^^^^^^^^^^^^^^^^
  File "/scratch/david_earthspecies_org/venvs/11138/lib/python3.12/site-packages/torch/utils/data/_utils/fetch.py", line 52, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
            ~~~~~~~~~~~~^^^^^
  File "/home/david_earthspecies_org/rep5/representation_learning/data/dataset.py", line 115, in __getitem__
    sample = self.ds[idx]
             ~~~~~~~^^^^^
  File "/scratch/david_earthspecies_org/venvs/11138/lib/python3.12/site-packages/esp_data/concat.py", line 444, in __getitem__
    raise RuntimeError(
RuntimeError: Failed to load item 1039279 from source dataset 0: Index 1039279 out of bounds for dataset of length 818136.

sattach: error: task 0 exited with exit code 1
david_earthspecies_org@slurm-login:~/rep5$
```
My suspicion is that there is a need need to call reset_index somewhere in the case that we concat. Config that can reproduce:

```
train_datasets:
  # AnimalSpeak dataset for training
  - dataset_name: animalspeak
    dataset_version: 0.0
    split: train
    balance: false
[...]

PR #107, opened 2025-07-28
Fix for issue #98, adding esp_raincoast dataset

GaganNarula:
- Reproduced issue #98  with a test in `test_concat.py`
- To fix ^^, added a `ignore_index=True` parameter that resets the index after operations like `drop_na`, `drop_duplicates` etc. In `filter.py` we explicitly call `.reset_index(drop=True)` to do the same thing. But we should discuss if this is the right way ? Currently the "index" of any esp-data dataset doesnt really carry any information so its fine to reset, but maybe it would be for certain use cases ? should we add info on this as a warning to the Docs ?
- Added `esp_raincoast` dataset. This is a v1, might require a second PR to fix some things as required by Emmanueal and co.

### DIFF
--- a/esp_data/__init__.py
+++ b/esp_data/__init__.py
@@ -17,6 +17,7 @@ from .datasets import (
     BengaleseFinchCalls,
     BirdSet,
     ChiffchaffId,
+    ESPRaincoast,
     GiantOtters,
     InsectSet459,
     LittleOwlId,
@@ -52,4 +53,5 @@ __all__ = [
     "MacaquesCooCalls",
     "Voxaboxen",
     "VoxaboxenEvents",
+    "ESPRaincoast",
 ]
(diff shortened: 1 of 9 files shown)


mil-ad:
> Currently the "index" of any esp-data dataset doesnt really carry any information so its fine to reset, but maybe it would be for certain use cases ? should we add info on this as a warning to the Docs ?

I don't think we need a warning. The fact the we're using a dataframe is an internal implementation detail that the user shouldn't need to know and definitely shouldn't access. 

There are scenarios where having [multi-dimensional index](https://pandas.pydata.org/docs/user_guide/advanced.html) but I doubt we'll need those.

mil-ad (approved):
Looks good. Please rebase the PR before merging

PR #108, opened 2025-08-04
💫 improved docs

mil-ad:
- Small improvement to doc generation. We don't need to manually add datasets to the .md file anymore.
- Added some make recipes for easier testing
- fix a numpy docstring issue
- Added a note for keyring on slurm


GaganNarula (approved):
good to go

Issue #112, opened 2025-08-07
VoxaboxenEvents sets arbitrary fields in the config


mil-ad:
In `VoxaboxenEvents` class we do this:

```
ds = cls(
    split=split,
    output_take_and_give=cfg.get("output_take_and_give", None),
    data_root=cfg.get("data_root"),
    sample_rate=cfg.get("sample_rate", 16000),
    mono_method=cfg.get("mono_method", "average"),
    clip_duration=cfg.get("clip_duration", 10.0),
    clip_hop=cfg.get("clip_hop", 5.0),
    clip_start_offset=cfg.get("clip_start_offset", 0.0),
    omit_empty_clip_prob=cfg.get("omit_empty_clip_prob", 0.0),
    scale_factor=cfg.get("scale_factor", 1),
    segmentation_based=cfg.get("segmentation_based", True),
    unknown_label=cfg.get("unknown_label", "Unknown"),
)
```

This has a few issues:

- We're using fields in that are not defined in `DatasetConfig`. The reason pydantic is not complaining is that for some reason we have `extra=allowed` in our model config:

```
model_config = ConfigDict(
    arbitrary_types_allowed=True,
    validate_assignment=True,
    str_strip_whitespace=True,
    extra="allow",
)
```

Maybe we need a specific pydantic model for this dataset but we should use `arbitrary_types_allowed=False` and `extra="forbid"` in general. 

- This also suffers from #94, lots of default values that should be in the config and `__init__`. I have fixed that for other datasets in #113 but leaving this one to @GaganNarula because I'm not super familiar with voxaboxen.

PR #113, opened 2025-08-07
make from_config()s dumb again

### LINKED ISSUE #94: remove any defaults from `from_factory()` methods
I [noticed](https://github.com/earthspecies/esp-data/pull/93#discussion_r2221752349) that we set fallback default values in some dataset's `from_config()` factory methods. We should not make a habit of this. These factory methods should be pretty dumb. If there are defaults they should be handled in configs and __init__ methods. Let's review all datasets and remove these fallbacks

mil-ad:
fixes #94

This should be followed up by #112

### DIFF
--- a/esp_data/datasets/littleowl_id.py
+++ b/esp_data/datasets/littleowl_id.py
@@ -95,20 +95,38 @@ class LittleOwlId(Dataset):
 
     @classmethod
     def from_config(cls, dataset_config: DatasetConfig) -> tuple["LittleOwlId", dict[str, Any]]:
+        """Create a Dataset instance from a configuration dictionary.
+
+        Parameters
+        ----------
+        dataset_config : DatasetConfig
+            Configuration dictionary containing dataset parameters
+
+        Returns
+        -------
+        tuple[Dataset, dict[str, Any]]
+            A tuple containing the dataset instance and metadata.
+            If the dataset_config contains transformations, they will be applied
+            and the metadata will be returned as dict, otherwise empty dict.
+        """
+
         cfg = dataset_config.model_dump(exclude=("dataset_name", "transformations"))
-        split = cfg.get("split", "train_across_year")
-        if split not in cls.info.split_paths:
-            available_splits = ", ".join(cls.info.split_paths.keys())
-            raise LookupError(f"Invalid split '{split}'. Available splits: {available_splits}")
-        ds = cls(
-            split=split,
-            output_take_and_give=cfg.get("output_take_and_give"),
-            data_root=cfg.get("data_root"),
-            sample_rate=cfg["sample_rate"],
-        )
+
+        # Do not include split in kwargs if not defined and let __init__ use the default
+        kwargs = {
+            "output_take_and_give": cfg["output_take_and_give"],
+            "data_root": cfg["data_root"],
+            "sample_rate": cfg["sample_rate"],
+        }
+        if cfg["split"]:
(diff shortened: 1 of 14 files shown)


esp_data/datasets/littleowl_id.py
```diff
@@ -95,20 +95,38 @@ def _load(self) -> None:
...
-            sample_rate=cfg["sample_rate"],
-        )
+
+        # Do not include split in kwargs if not defined and let __init__ use the default
+        kwargs = {
+            "output_take_and_give": cfg["output_take_and_give"],
+            "data_root": cfg["data_root"],
+            "sample_rate": cfg["sample_rate"],
+        }
+        if cfg["split"]:
```
GaganNarula, line 121: seems like this bit of code here is different from the template followed in the other datasets in this PR ? 

also, should we make the `split` parameter in the `DatasetConfig` have a default `None` value ? Right now it goes `split: str = "train"`

GaganNarula (changes requested):
can we add a `from_sources` method to the DatasetConfig ? to allow loading from yaml ? or another PR for that ?

mil-ad:
> can we add a from_sources method to the DatasetConfig ? to allow loading from yaml ? or another PR for that ?

Oh yeah I think that's been a successful recipe in repr-learning. We can probably use the same `from_config()`. It's pretty easy to detect if the argument is a dict/config or a path to a yaml file. I'll do it in another PR.

mil-ad:
raised #114 to track that

esp_data/datasets/littleowl_id.py
mil-ad: Yeah I had to handle this one differently because of the default value for split in the config. I think there was another dataset that has a similar issue. pipid maybe?

> also, should we make the split parameter in the DatasetConfig have a default None value ? Right now it goes split: str = "train"

I think train is a sensible default. I think the main question is how do to we want to handle datasets that don't fit DatasetConfig. It could be different default values (like this one) or extra fields like voxaboxen

mil-ad: how do you propose I change this in this PR?

GaganNarula: >  I think the main question is how do to we want to handle datasets that don't fit DatasetConfig

Well the DatasetConfig is quite restrictive right now.. it only covers the very basic datasets.  if we want all datasets to be able to be instantiated from yaml files, we would need extra fields and bypass validation by pydantic for those fields... this is especially important when users create datasets _externally_ using the `register_dataset` hook. They'll might want to implement some adhoc logic in their "unofficial" datasets and still use yaml to configure whatever fields they like

GaganNarula: I think we can call this PR done for now

GaganNarula (approved):
all good for now

esp_data/datasets/littleowl_id.py
mil-ad: For transforms we expect the user to add a config class and transform class. We already deal with from_config() there too. We could do the same here? 

I think subclassing DatasetConfig is reasonable.

GaganNarula: hmm yes I guess its fine to subclass it

mil-ad: I mean, I'm subclassing's enemy number one 😄 but in this case I think it makes sense

PR #116, opened 2025-08-13
Improvements to from_config + fixes

mil-ad:
- allow specifying yaml paths directly
- When using @abstractmethod it doesn't make sense to raise exceptions
- Made some @properties follow @abstractmethod to force the child class to implement them
  - these actually triggered some fails in CI. The problem was in the test

- We're doing this in most datasets `from_config()s`:

```
  cfg = dataset_config.model_dump(exclude=("dataset_name", "transformations"))
```

- We were using the wrong type when passing `exclude` in `model_dump()`. Fixed by providing a set

### DIFF
--- a/esp_data/dataset.py
+++ b/esp_data/dataset.py
@@ -248,6 +248,7 @@ class Dataset(ABC):
         self.output_take_and_give = output_take_and_give
 
     @property
+    @abstractmethod
     def available_splits(self) -> Sequence[str]:
         """Get the available splits of the dataset.
 
@@ -256,9 +257,10 @@ class Dataset(ABC):
         Sequence[str]
             A sequence of split names available in the dataset.
         """
-        raise NotImplementedError
+        pass
 
     @property
+    @abstractmethod
     def columns(self) -> Sequence[str]:
         """Get the columns of the dataset.
 
@@ -267,7 +269,7 @@ class Dataset(ABC):
         Sequence[str]
             A sequence of column names in the dataset.
         """
-        raise NotImplementedError
+        pass
 
     @abstractmethod
     def _load(self) -> Optional[Sequence[Any]]:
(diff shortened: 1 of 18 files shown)


esp_data/dataset.py
```diff
@@ -256,9 +259,10 @@ def available_splits(self) -> Sequence[str]:
         Sequence[str]
             A sequence of split names available in the dataset.
         """
-        raise NotImplementedError
```
GaganNarula, line 259: question, do we want this to be an abstract class or a base class ? i had initially considered it an abstract class, but then there was this method `apply_transformations` that actually does some computation (so more like a base class from which the child datasets inherit behavior)

GaganNarula: if its a base class i'd leave the `raise NotImplementedError` in there

esp_data/dataset.py
```diff
@@ -457,7 +462,13 @@ def dataset_from_config(
     KeyError
         If the dataset is not registered
     """
+
+    if isinstance(dataset_config, (Path, str)):
+        with Path(dataset_config).open("r") as fp:
+            dataset_config = DatasetConfig.model_validate(yaml.safe_load(fp))
+
```
GaganNarula, line 469: It might be better to load the yaml into a dict first... because we would like to use keys like "concat" etc to distinguish between DatasetConfig and ConcatConfig etc (TBD)

mil-ad: safe_load() returns a dict right? or do you mean store it in an intermediate variable? We can do that we doing DatasetConfig vs ConcatConfig etc

esp_data/dataset.py
mil-ad: are you talking about the class or this particular method?

I think it's fine for the class to be `abc` and have some methods with implementations and some `@abstractmethod` methods

GaganNarula: I'm talking about this class. If it's a base class, it makes sense to raise an error if a subclass doesn't implement a method no? Otherwise it's just an interface

GaganNarula (approved):
looks good!

PR #118, opened 2025-08-14
Dataset performance benchmarks

GaganNarula:
Simple benchmarks for iterating over a raw dataset, with / without a pytorch dataloader and from nfs or bucket.


mil-ad (approved):
LGTM

Issue #120, opened 2025-08-25
`dataset_from_config` for concatenating datasets


GaganNarula:
We want that the function `dataset_from_config` can read a yaml file and either:
1. Return a single Dataset
2. Return a concatenated Dataset
3. Return a collection of Datasets

Adds a `key` parameter to the function so that if a yaml file contains more than one dataset configuration, the user must provide a value for `key` which chooses one of the configurations

PR #121, opened 2025-08-25
Extend dataset_from_config

GaganNarula:
#120 `dataset_from_config` now loads:
1. Single Dataset 
2. ConcatenatedDataset
3. A collection (list) of Dataset / ConcatenatedDataset

Added a `key` parameter to choose which configuration to load from a yaml that has several configs.

### DIFF
--- a/esp_data/concat.py
+++ b/esp_data/concat.py
@@ -6,7 +6,13 @@ from typing import Any, Dict, Iterator, Literal, Optional
 import pandas as pd
 import semver
 
-from esp_data.dataset import Dataset, DatasetConfig, DatasetInfo
+from esp_data.dataset import (
+    ConcatConfig,
+    Dataset,
+    DatasetInfo,
+    dataset_from_config,
+    register_dataset,
+)
 
 logger = logging.getLogger("esp_data")
 
@@ -346,6 +352,7 @@ def concatenate_datasets(
             dataset.output_take_and_give = original_otag
 
 
+@register_dataset
 class ConcatenatedDataset(Dataset):
     """A dataset created by concatenating multiple datasets.
 
@@ -379,7 +386,15 @@ class ConcatenatedDataset(Dataset):
         "Concatenated dataset length should match sum of source datasets lengths"
     """
 
-    # No class-level info attribute to avoid overwriting issues
+    info = DatasetInfo(
+        name="concatenated_dataset",
+        owner="ESP Data Team",
+        split_paths={"concatenated": "virtual://concatenated_dataset"},
+        version="0.1.0",
+        description="A dataset created by concatenating multiple datasets.",
+        sources=["Multiple datasets"],
+        license="CC0-1.0",
+    )
 
     def __init__(
         self,
@@ -391,7 +406,7 @@ class ConcatenatedDataset(Dataset):
     ) -> None:
         super().__init__(output_take_and_give)
         self._data = data
-        self.info = dataset_info  # Instance attribute, not class attribute
+        self.info = dataset_info  # Overwrite info with provided metadata
         self._source_datasets = source_datasets  # Private attribute
         self.sample_rate = sample_rate
         self.split = "concatenated"
@@ -409,11 +424,39 @@ class ConcatenatedDataset(Dataset):
         pass  # Data is already loaded
 
     @classmethod
-    def from_config(cls, dataset_config: DatasetConfig) -> None:
-        raise NotImplementedError(
-            "ConcatenatedDataset cannot be instantiated from config. "
-            "Use concatenate_datasets function instead."
-        )
+    def from_config(
+        cls, concat_config: ConcatConfig
+    ) -> tuple["ConcatenatedDataset", Dict[str, Any]]:
+        """Create a ConcatenatedDataset from a ConcatConfig object.
+
+        Parameters
+        ----------
+        concat_config : ConcatConfig
+            Configuration object specifying the datasets to concatenate
+            and how to merge them.
+
+        Returns
+        -------
+        tuple[ConcatenatedDataset, dict]
+            A tuple containing the ConcatenatedDataset instance
+            and metadata about transformations applied.
+
+        Raises
+        -------
+        ValueError
+            If the concatenated dataset is empty after merging.
+            This can happen if no datasets are provided or if all datasets are empty,
+            or if the merge level results in an empty dataset.
+        """
+        datasets = [dataset_from_config(cfg)[0] for cfg in concat_config.datasets]
+        ds = concatenate_datasets(datasets, merge_level=concat_config.merge_level)
+        if len(ds) == 0:
+            raise ValueError("Concatenated dataset is empty. Check input datasets or merge level.")
+        if concat_config.transformations:
+            transform_metadata = ds.apply_transformations(concat_config.transformations)
+            return ds, transform_metadata
+
+        return ds, {}
 
     def __len__(self) -> int:
         return len(self._data)
--- a/esp_data/concat.py
+++ b/esp_data/concat.py
@@ -338,12 +338,12 @@ def concatenate_datasets(
         dataset_infos = [ds.info for ds in datasets]
         merged_info = _merge_dataset_info(dataset_infos)
 
-        return ConcatenatedDataset(
-            data=concatenated_df,
-            dataset_info=merged_info,
-            source_datasets=datasets,
-            sample_rate=combined_sample_rate,
-            output_take_and_give=combined_otag,
+        return (
+            concatenated_df,
+            merged_info,
+            datasets,
+            combined_sample_rate,
+            combined_otag,
         )
 
     finally:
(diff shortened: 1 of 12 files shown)


esp_data/concat.py
```diff
@@ -346,6 +352,7 @@ def concatenate_datasets(
             dataset.output_take_and_give = original_otag
 
 
+@register_dataset
```
mil-ad, line 355: is there a usecase for registering concat'd dataset? The `name` is also not unique anyway

esp_data/concat.py
```diff
@@ -409,11 +424,39 @@ def _load(self) -> None:
...
+        -------
+        ValueError
+            If the concatenated dataset is empty after merging.
+            This can happen if no datasets are provided or if all datasets are empty,
+            or if the merge level results in an empty dataset.
+        """
+        datasets = [dataset_from_config(cfg)[0] for cfg in concat_config.datasets]
+        ds = concatenate_datasets(datasets, merge_level=concat_config.merge_level)
+        if len(ds) == 0:
+            raise ValueError("Concatenated dataset is empty. Check input datasets or merge level.")
```
mil-ad: we should follow the same pattern in this `from_config()` that we do (or should do) in other datasets, i.e. this should ultimately call `ConcatenatedDataset`'s `__init__`. 

I'm not sure what's the point of `__init__`'s arguments right now. They don't do anything. Can't we have `concatenate_datasets()` arguments in the `__init__` and then `from_config()` just instantiates the datasets and calls it? This is why I find inheritance weird, it's forcing `ConcatenatedDataset` to have an interface that's not making sense.

esp_data/dataset.py
```diff
@@ -98,6 +99,50 @@ def delay_importing_reg(cls, v: Any) -> Any:  # noqa: ANN401
...
+            from esp_data.transforms.registry import RegisteredTransformConfigs
+
+            adapter = TypeAdapter(RegisteredTransformConfigs)
+
+            validated = []
+            for t in v:
+                validated.append(adapter.validate_python(t))
+            return validated
+        else:
+            return None
```
mil-ad, line 150: if we remove `register_dataset()` from concat dataset this bit can go right?

esp_data/dataset.py
```diff
@@ -416,8 +461,15 @@ def register_dataset(cls: type[Dataset]) -> type[Dataset]:
...
         The registered dataset class
+
+    Raises
+    -------
+    ValueError
+        If the dataset name is already registered
     """
     name = cls.info.name
+    if name in _dataset_registry:
+        raise ValueError(f"Dataset '{name}' is already registered.")
```
mil-ad, line 482: this check should happen in `register_dataset()`. Look at `register_transform()` for similar checks

esp_data/dataset.py
```diff
@@ -439,36 +491,114 @@ def print_registered_datasets() -> None:
...
     return _dataset_class.from_config(dataset_config)
+
+
+def _make_single_dataset_or_concat(
+    cfg: dict[str, Any],
+) -> tuple[Dataset, dict[str, Any]]:
+    try:
+        return _make_dataset_from_config(DatasetConfig.model_validate(cfg))
+    except Exception:
+        return _make_dataset_from_config(ConcatConfig.model_validate(cfg))
```
mil-ad: is there enough here to have a function?

esp_data/dataset.py
```diff
@@ -439,36 +491,114 @@ def print_registered_datasets() -> None:
...
+
+    # Last option is that this is a collection:
+    # list of single DatasetConfigs / ConcatConfigs
+    if isinstance(data[key], list):
+        output_datasets = []
+        output_transform_metadata = []
+        for item in data[key]:
+            ds, transform_metadata = _make_single_dataset_or_concat(item)
+            output_datasets.append(ds)
+            output_transform_metadata.append(transform_metadata)
```
mil-ad: it looks like this bit is trying too hard to be flexible. We either expect to see `dataset_name` key or `concat` key in root (or root[key] if key is provided)

tests/test_concat.py
mil-ad, line 1: we need a test for nested concat datasets I think

mil-ad (changes requested):

esp_data/dataset.py
GaganNarula: but this line is already in the `register_dataset` function.. ?

esp_data/dataset.py
GaganNarula: the issue is `dataset_from_config` doesnt work without `register_dataset` so concat config will also not work, and hence `ConcatenatedDataset.from_config` wont work

esp_data/concat.py
GaganNarula: same answer as below

esp_data/dataset.py
GaganNarula: had to repeatedly call these few lines in the `dataset_from_config` so wrote it as a helper

esp_data/dataset.py
GaganNarula: oh so this is what i didnt understand.. so "dataset_name" and "concat" are special keys ? i wasnt sure..

esp_data/dataset.py
mil-ad: oops yeah. The diff view chopped it at a weird place

esp_data/dataset.py
mil-ad: yeah I think being explicit makes things easier for us, wdyt? We should probably also rename `dataset_name` to just `name` imo.

esp_data/dataset.py
mil-ad: can't we move things a bit around. dataset_from_config() could be calling ConcatDataset's `from_config()` _or_ use registry

mil-ad: we can discuss in our chat today, I'm also wondering what happens when things are nested 🙃

esp_data/concat.py
GaganNarula: we could have the arguments of `concatenate_datasets` as the `__init__` arguments of the class but then that changes what `concatenate_datasets` returns..will be breaking some representation-learning code ? its ok by me

mil-ad: I think that's ok. We're introducing breaking changes anyway with the config. They'll have to update that too

esp_data/dataset.py
GaganNarula: hey i'm running into a circular import issue trying to implement what you suggest here ^^. The problem is `concat.py` imports from `dataset.py`, so `dataset.py` cant import `ConcatenatedDataset`

esp_data/datasets/voxaboxen.py
mil-ad, line 1: can we do this in a separate PR?

esp_data/dataset.py
```diff
@@ -439,36 +519,124 @@ def print_registered_datasets() -> None:
...
+            # If the dict contains a single dataset/concat config, we can create it
+            return _make_single_dataset_or_concat(data)
+        else:
+            raise ValueError(
+                "Multiple / Invalid dataset configurations found. "
+                "Please provide a specific key to select one."
+            )
+
+    # An option is that this is a collection:
+    # list of single DatasetConfigs / ConcatConfigs
```
mil-ad: do we want to support this?

GaganNarula: i think we discussed it last week and decided yes.. only that it cant be nested.. it makes sense to support it, for e.g. when doing evaluation on a set of test datasets, we dont want to concatenate them (for whatever reason, like different metrics or different data preps etc) and still want to define them in a single config file

esp_data/datasets/voxaboxen.py
GaganNarula: it was really as an integration test for the `_config_registry` and `register_config` decorator.. i can remove it if its better in a separate PR

esp_data/dataset.py
mil-ad: isn't that already supported with the `key` thing?

GaganNarula: the `key` chooses between several different definitions in a single yaml. The collection is a list of definitions of either `DatasetConfig` or `ConcatConfig`, so a `key` cant really choose amongst the list if that was what you were asking ? (an `index` could but we dont want that i think)..

If this is getting too complicated we can drop it, but i do know that there is a need for it in repr-learning project

mil-ad: I'd drop it for now. We can change the schema to support this later if everyone needs it

esp_data/datasets/voxaboxen.py
mil-ad: 👍🏼

esp_data/dataset.py
mil-ad: is this still an issue?

GaganNarula: currently the solution is that `ConcatenatedDataset` has a default `info` attribute where the `dataset_name` is `concatenated_dataset` such that it's a special case in the dataset registry

mil-ad: I'm still not convinced we want concat in the registry but we can change that in a separate PR. I was referring to the cyclic dependency, if that's fixed we should merge

GaganNarula: thats what i meant, its not an issue (solved) _because_ concat its currently in the registry

mil-ad: then let's go 🚀

mil-ad (approved):

PR #123, opened 2025-09-04
Another attempt at removing cloudpathlib

mil-ad:
This PR removes cloudpathlib and just uses filesystem.

The main complexity is that fsspec doesn't work nicely with Path-like objects. There has been quite a few attempts in the community to add support to fsspec but it's complicated. 

But we quite like Path-like objects so this PR extends PurePaths from python to work with cloud Paths. This is not pretty but let's us at least do foo / bar style operations and we explicitly use filesystem or util functions that use filesystems for io operations.

- The way were processing and saving "data_root" made certain assumptions and tests started failing after the changes in the `io` module. Fixed a few issues when while I was at it:
  - Noticed that the default data_root for audioset dataset was wrong. Fixed and updated the tests. @GaganNarula we probably need new tests to make sure that custom data_root works
  - also noticed that some of data_root args in `__init__` had the old style type hints. Have fixed some of those in this PR too.


esp_data/io/filesystem.py
GaganNarula, line 21: should we continue caching this ? I dont know if we ever discussed this

mil-ad: ah this is related to what @Paulchen-git found right? My guess is that the problem goes away with changing spawn method even with this `@cache` in place. Let's discuss that in a new issue. We have to fix it somehow :D

esp_data/io/paths.py
```diff
@@ -1,198 +1,265 @@
...
-    client : cloudpathlib.S3Client, optional
-        An explicit `cloudpathlib.S3Client` instance to use. If None, the
-        default cached client (`__client`) configured with R2 credentials
-        is used. Defaults to None.
-    """
-
-    is_cloud: bool = True
-    is_local: bool = False
+    @property
+    def object_path(self) -> str:
```
GaganNarula: are these methods like `with_bucket`, `relative_to_bucket`, `object_path` etc needed ?

esp_data/io/file_utils.py
```diff
@@ -0,0 +1,17 @@
...
+def exists(path: str | AnyPathT) -> bool:
+    return filesystem_from_path(path).exists(str(path))
+
+
+def rm(
+    path: str | AnyPathT,
+    recursive: bool = False,
+    maxdepth: int | None = None,
+    **kwargs: dict[str, Any],
+) -> None:
```
GaganNarula, line 28: there are no docstrings here.. i think it would help, at least with `rm`

esp_data/datasets/barkley_canyon.py
```diff
@@ -83,9 +83,11 @@ def __init__(
         self._load()  # Load the dataset (fills self._data)
         self.sample_rate = sample_rate
         self.data_root = data_root
-        if self.data_root is None:
-            # we assume that parent dir of the split path is the data root
+
+        if data_root is None:
             self.data_root = anypath(self.info.split_paths[self.split]).parent
+        else:
+            self.data_root = data_root
```
GaganNarula, line 89: i dont think we need this `else` because of line 85 `self.data_root = data_root`

esp_data/datasets/barkley_canyon.py
```diff
@@ -185,13 +187,10 @@ def __getitem__(self, idx: int) -> dict[str, Any]:
...
-        # Ensure audio path is valid
-        if self.data_root:
-            audio_path = anypath(self.data_root) / row["local_path"]
-        else:
-            audio_path = anypath(row["local_path"])
 
-        with audio_path.open("rb") as f:
+        audio_path = anypath(self.data_root) / row["local_path"]
+
+        with filesystem_from_path(audio_path).open(str(audio_path), "rb") as f:
```
GaganNarula, line 192: we can replace this with the `get_audio_info` function in `read_utils`

esp_data/datasets/esp_raincoast.py
```diff
@@ -110,8 +112,7 @@ def _load(self) -> None:
         else:
             from io import StringIO
 
-            fs = filesystem_from_path(location)
-            csv_text = fs.read_text(str(anypath(location).no_prefix), encoding="utf-8")
+            csv_text = read_text(location, encoding="utf-8")
```
GaganNarula, line 115: just fyi, i think i will be replacing all this later in the databackends PR, but this is good right now

esp_data/datasets/insectset_459.py
```diff
@@ -59,8 +59,8 @@ def __init__(
         self,
         split: str = "train",
         output_take_and_give: dict[str, str] = None,
-        sample_rate: Optional[int] = None,
-        data_root: Optional[str | AnyPathT] = None,
```
GaganNarula, line 63: i wonder if we can configure these type hint format detectors in pre-commit or something ? :D

esp_data/datasets/voxaboxen.py
```diff
@@ -371,7 +368,7 @@ def __len__(self) -> int:
     def __getitem__(self, idx: int) -> dict[str, Any]:
         """Get a specific sample from the dataset.
         Parameters
-        ----------
+        ----------2
```
GaganNarula: is the "2" for something ?

esp_data/io/__init__.py
```diff
@@ -1,15 +1,25 @@
...
+)
 
 __all__ = [
     "anypath",
     "AnyPathT",
-    "GSPath",
-    "R2Path",
+    "PureGSPath",
+    "PureR2Path",
+    "PureS3Path",
```
GaganNarula, line 16: should we just keep this private to the `io` module ? i think we only need `anypath` and `AnyPathT`

tests/test_barkley_canyon.py
```diff
@@ -326,7 +326,7 @@ def test_detection_info_property(dataset_detection: Dataset) -> None:
     assert dataset_detection.info.version == "0.1.0"
     assert "train" in dataset_detection.info.split_paths
     assert "validation" not in dataset_detection.info.split_paths
-    assert anypath(dataset_detection.info.split_paths["train"]).exists()
+    assert exists(dataset_detection.info.split_paths["train"])
```
GaganNarula, line 329: what about having a single test that checks for all available_split for all registered datasets ?

esp_data/datasets/barkley_canyon.py
mil-ad: good catch, made it consistent

esp_data/datasets/insectset_459.py
mil-ad: Ha! There's one! https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/

I'll turn this on in a separate PR!

mil-ad: #139

esp_data/datasets/voxaboxen.py
mil-ad: Oops, fixed!

esp_data/io/paths.py
mil-ad: No! I used to use them but got rid of them. Will clean up

esp_data/datasets/esp_raincoast.py
mil-ad: sounds good!

tests/test_gcs_paths.py
```diff
@@ -0,0 +1,240 @@
...
+
+    # Test invalid protocol
+    with pytest.raises(ValueError, match="Path must start with 'gs://': s3://bucket/file"):
+        PureGSPath("s3://bucket/file")
+
+
+def test_edge_cases():
+    """Test edge cases and special scenarios."""
+    # Test path with multiple consecutive slashes
+    path = PureGSPath("gs://my-bucket//folder///file.txt")
```
GaganNarula, line 118: really nice!

tests/test_gcs_paths.py
```diff
@@ -0,0 +1,240 @@
...
+def test_joinpath_equivalence():
+    b = PureGSPath("gs://bucket/base")
+    assert str(b.joinpath("sub", "file.txt")) == "gs://bucket/base/sub/file.txt"
+    # joining an absolute subpath (bucket-rooted)
+    assert str(b / "/abs/thing") == "gs://bucket/abs/thing"
+
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
```
GaganNarula: what is this for ?

tests/test_gcs_paths.py
```diff
@@ -0,0 +1,240 @@
...
+    rel = p.relative_to("gs://bucket/")
+    assert isinstance(rel, PureGSPath)
+    assert str(rel) == "f1/f2/file.txt"
+
+
+def test_joinpath_equivalence():
+    b = PureGSPath("gs://bucket/base")
+    assert str(b.joinpath("sub", "file.txt")) == "gs://bucket/base/sub/file.txt"
+    # joining an absolute subpath (bucket-rooted)
+    assert str(b / "/abs/thing") == "gs://bucket/abs/thing"
```
GaganNarula, line 185: i thought this would turn out to be `gs://bucket/base/abs/thing` ?

pyproject.toml
```diff
@@ -5,7 +5,6 @@ description = "Data tools for ESP projects"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "cloudpathlib[all]>=0.20.0",
```
GaganNarula, line 8: 🎉

GaganNarula (changes requested):
a few things to improve, see comments

tests/test_gcs_paths.py
mil-ad: leakage from cursor. I'll remove 🙈

pyproject.toml
mil-ad: 😌

tests/test_barkley_canyon.py
mil-ad: I actually quite like that I can run a test for a particular dataset. Maybe it's possible to define one test and instantiate it for each dataset?

esp_data/io/paths.py
mil-ad: gone

tests/test_gcs_paths.py
mil-ad: gone

tests/test_gcs_paths.py
mil-ad: well, if we half-close our eyes and pretend that these are unix-ish paths then `/abs/thing` is different from `abs/thing`

similar to normal files:

```
>>> f = Path("/foo/bar")
>>> f
PosixPath('/foo/bar')
>>> f / "/boom/blah/q.txt"
PosixPath('/boom/blah/q.txt')
>>> f / "boom/blah/q.txt"
PosixPath('/foo/bar/boom/blah/q.txt')
```

tests/test_barkley_canyon.py
GaganNarula: yeah, we could add a fixture in the another PR !

esp_data/io/file_utils.py
mil-ad: 🫡
fixed

esp_data/datasets/barkley_canyon.py
mil-ad: yes! can we do in a separate PR?

GaganNarula: 👍

tests/test_gcs_paths.py
```diff
@@ -0,0 +1,240 @@
...
+    # Test stem property
+    assert path.stem == "file"
+
+    # Test parent property
+    parent = path.parent
+    assert str(parent) == "gs://my-bucket/folder"
+
+    # Test parts property
+    parts = path.parts
+    assert len(parts) >= 2  # At least gs://bucket and the object path
```
GaganNarula: maybe lets assert exactly what the parts will be `gs://bucket` and `folder` or `gs://`, `bucket` and `folder` ?

tests/test_gcs_paths.py
```diff
@@ -0,0 +1,240 @@
...
+    joined = base / "file.txt"
+    assert str(joined) == "gs://my-bucket/folder/file.txt"
+
+    # Test joining with absolute path
+    joined = base / "/absolute/path"
+    assert str(joined) == "gs://my-bucket/absolute/path"
+
+    # Joining from bucket-only should insert root separator
+    bucket_only = PureGSPath("gs://my-bucket")
+    assert str(bucket_only / "bar/text.text") == "gs://my-bucket/bar/text.text"
```
GaganNarula, line 77: ah so i see a key difference here compared to the comment later: joining a bucket to an absolute path is allowed, but joining a path like `gs://bucket/base` to an absolute path `/abs/path` is not right ? it might be important to document that

GaganNarula: I was referring to the line 119 and 120

GaganNarula: oh hmm, so doing `joined = base / "/absolute/path"` on line 119 actually drops the `folder/subfolder/file.txt` and replaces it ... its a bit confusing

esp_data/io/__init__.py
mil-ad: I personally mind it being public.

mil-ad: I'll remove if you feel strongly about it :D

tests/test_gcs_paths.py
mil-ad: done

tests/test_gcs_paths.py
mil-ad: I think it is allowed. See comment [here](https://github.com/earthspecies/esp-data/pull/123#discussion_r2435766988)

mil-ad: it treats `/` as the root

esp_data/datasets/insectset_459.py
mil-ad: ready for review in #140!!

GaganNarula (approved):
i think its good to go!

Issue #117, opened 2025-08-13
streaming=True in esp-data datasets


GaganNarula:
We want to be able to:

1. Import and instantiate datasets without delay
2. Only iterate rather than index into datasets, with a config param `streaming=True`

GaganNarula:
This is now part of the PR #131 as a part of the DataBackend

PR #124, opened 2025-09-10
Attempting streaming with pandas

GaganNarula:
#117 
`streaming=True` will not load the `self._data` dataframe at init, and won't allow indexing only iteration.


GaganNarula:
Streaming handling is done better in #131

Issue #125, opened 2025-09-12
Error message when first use of esp-data library


Paulchen-git:
Problem come from google cloud configuration.

<img width="946" height="274" alt="Image" src="https://github.com/user-attachments/assets/5d31fbb2-457f-41a3-a42a-e1d63ea45a54" />

Error message when running Direct Instanciation of the Datasets part in esp-data documentation: `OSError: Project was not passed and could not be determined from the environment.`

PR #126, opened 2025-09-12
update esp-data documentation with Google Cloud set up

Paulchen-git:
Updated documentation with required steps to set up Google Cloud for the first time and access esp-data package. #125

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -29,7 +29,66 @@ uv tool install keyring --with keyrings.google-artifactregistry-auth
 !!! tip
     `uv tool` allows you to install Python packages that provide command-line interfaces for system-wide use. The dependencies are installed in an isolated virtual environment, separate from your current project.
 
-### 2. Configure your project to use the private index
+### 2. Set up Google Cloud to access `esp-data` package
+
+This step is required if you haven't set up Google Cloud on your device yet. If Google Cloud isn't properly set up the following steps will fail.
+
+#### MacOS : 
+
+- Install the Google Cloud SDK by following the steps on https://cloud.google.com/sdk/docs/install or using Homebrew with the following command :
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -29,7 +29,66 @@ uv tool install keyring --with keyrings.google-artifactregistry-auth
 !!! tip
     `uv tool` allows you to install Python packages that provide command-line interfaces for system-wide use. The dependencies are installed in an isolated virtual environment, separate from your current project.
 
-### 2. Configure your project to use the private index
+### 2. Set up Google Cloud to access `esp-data` package
+
+This step is required if you haven't set up Google Cloud on your device yet. If Google Cloud isn't properly set up the following steps will fail.
+
+#### MacOS :
```
GaganNarula: I think we dont need a macos specific setup, its better to remove this section and instead start with the next "Install the Google Cloud SDK..." bit

docs/index.md
```diff
@@ -29,7 +29,66 @@ uv tool install keyring --with keyrings.google-artifactregistry-auth
...
+    ```sh
+    gcloud auth application-default login
+    ```
+    You will have to authenticate again and then you should get the following output :
+    ```sh
+    Your browser has been opened to visit:
+
+	https://accounts.google.com/...
+
+    Credentials saved to file: [/Users/username/.config/gcloud/application_default_credentials.json]
```
GaganNarula: agan this is MacOs specific, so either state both `/Users/...` and the Linux and Windows paths or just remove it for now..

GaganNarula (changes requested):
looking good, just a few small changes!

GaganNarula (approved):
lgtm!

PR #131, opened 2025-10-01
Pandas + polars as backends

GaganNarula:
We need to reduce the dependency on pandas, so we abstract away the concrete backend of the data (for now just the dataframes).

- Introduces the `DataBackend` protocol which defines an interface but doesn't enforce inheritance
- Polars and Pandas backend implementations follow this protocol 
- Switch between backends in each Dataset, adds `backend` parameter to Dataset and DatasetConfig
- `transforms` now accept `DataBackend`s instead of a particular implementation like pandas dataframe (current)


esp_data/backends/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+    -------
+    list[str]
+        List of registered backend names
+
+    Examples
+    --------
+    >>> list_backends()
+    ['pandas', 'polars']
+    """
+    return list(_BACKEND_REGISTRY.keys())
```
mil-ad: I don't think these belong in the `__init__`

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,649 @@
...
+
+from __future__ import annotations
+
+from typing import Any, Iterator, Literal, Protocol, TypeVar, overload
+
+# Type alias for backend return type (allows chaining)
+DataFrameBackendT = TypeVar("DataFrameBackend")
+
+
+class DataFrameBackend(Protocol):
```
mil-ad: Do we expect this to go beyond pandas and polars? if yes maybe call this `DataBackend`?

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,649 @@
...
+    --------
+    >>> backend = PandasBackend.read_csv("data.csv")
+    >>> row = backend[0]  # Get first row as dict
+    >>> filtered = backend.filter_isin("species", ["cat", "dog"])
+    >>> for row in filtered:
+    ...     print(row)
+    """
+
+    @classmethod
+    def read_csv(
```
mil-ad: looks like a factory method. rename to `from_csv()`

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,649 @@
...
+
+        Examples
+        --------
+        >>> backend = PandasBackend.read_csv("data.csv")
+        >>> backend = PolarsBackend.read_csv("gs://bucket/data.csv", streaming=True)
+        """
+        ...
+
+    @classmethod
+    def read_json(
```
mil-ad: from_json()

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,649 @@
...
+            Backend instance wrapping the loaded DataFrame
+
+        Examples
+        --------
+        >>> backend = PandasBackend.read_json("data.jsonl", lines=True)
+        """
+        ...
+
+    @classmethod
+    def read_parquet(
```
mil-ad: from_parquet

esp_data/dataset.py
```diff
@@ -297,9 +305,16 @@ def __init__(self, output_take_and_give: dict[str, str] = None) -> None:
...
+        backend : BackendType, optional
+            The backend to use for DataFrame operations ("pandas" or "polars"), by default "polars"
+        streaming : bool, optional
+            Whether to use streaming mode for dataset processing, by default False
 
         """
         self.output_take_and_give = output_take_and_give
+        self._backend_type = backend
+        self._streaming = streaming
+        self._backend_class = get_backend(backend)
```
mil-ad, line 316: why save this as an attribute?

esp_data/dataset.py
```diff
@@ -439,13 +454,28 @@ def apply_transformations(
         -------
         RuntimeError
             If the dataset's data is not loaded yet.
+            If using pandas backend in streaming mode (not supported).
+
+        Notes
+        -----
+        Transformations in streaming mode are only supported with polars backend
+        (using LazyFrame). Pandas backend requires eager evaluation.
```
mil-ad: ideally the comments here should not include backend-specific details. That's the main point of having a protocol.

We could just check if the protocol implements this and if not error out.

esp_data/dataset.py
```diff
@@ -439,13 +454,28 @@ def apply_transformations(
...
         """
         if self._data is None:
             raise RuntimeError("No data loaded. Call load() first.")
 
+        # Check if pandas backend in streaming mode
+        if self._streaming and self._backend_type == "pandas":
```
mil-ad: same as comment above, it's strange to have `self._backend_type == "pandas"` here

esp_data/dataset.py
```diff
@@ -70,6 +71,8 @@ class DatasetConfig(BaseModel):
     output_take_and_give: dict[str, str] | None = None
     split: str = "train"
     data_root: Optional[str] = None
+    streaming: bool = False
+    backend: BackendType = "polars"
```
mil-ad, line 75: Can we import and use `PandasBackend` directly. That allows us to get rid of `get_backend()` and the entire backend registry.

esp_data/dataset.py
```diff
@@ -288,7 +291,12 @@ class Dataset(ABC):
 
     info: DatasetInfo
 
-    def __init__(self, output_take_and_give: dict[str, str] = None) -> None:
+    def __init__(
+        self,
+        output_take_and_give: dict[str, str] = None,
+        backend: BackendType = "polars",
```
mil-ad, line 297: [thinking out loud] this is kind of like the comment I had in the design doc for esp-toolkit. Does it make sense for all datasets to define a backend? isn't it better to instantiate a backend outside and pass it to this?

mil-ad (commented):
left some comments

esp_data/backends/__init__.py
GaganNarula: shall we put this in `registry.py` like in `transforms` ?

GaganNarula: or in protocol.py ?

esp_data/dataset.py
GaganNarula: good point!

esp_data/dataset.py
GaganNarula: the issue is that we want the Dataset to load the csv (either into memory or as streaming like LazyFrame) when instantiated .. otherwise we'd have to call the `Backend.from_csv` outside of the dataset and the user would have to know the split path. This is simpler, especially if a dataset has multiple backends (like arrow / parquet etc)

esp_data/dataset.py
GaganNarula: idk, just some Claude bug

esp_data/dataset.py
GaganNarula: Or do you mean the user should do:

```
from esp_data.backends import PandasBackend

backend = PandasBackend()
ds = SomeDataset(.., backend=backend)
```

This wont work currently because these dataframe backends actually need the dataframe on construction (`backend = PandasBackend(df)`).. also even if they didnt (null by default), this would be a problem when constructing a Dataset from a yaml config

esp_data/dataset.py
GaganNarula: actually this isnt a bug.. the `_backend_class` is used to load the csv with `_backend_class.from_csv`. But now that i think of it, we could just have `get_backend(backend)` call also in the individual datasets without saving an attibute..

esp_data/backends/pyarrow_backend.py
mil-ad, line 1: is this file empty? maybe remove for now

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+    -------
+    list[str]
+        List of registered backend names
+
+    Examples
+    --------
+    >>> list_backends()
+    ['pandas', 'polars']
+    """
+    return list(_BACKEND_REGISTRY.keys())
```
mil-ad: sorry I don't remember if we discussed this or not but why have a registry? do we expect users to register their own backends? I did a quick search and `list_backends() and `get_backends()` are not used anywhere (expect a test)

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+        Returns
+        -------
+        DataBackend
+            New backend wrapping the result of the function application
+        """
+        ...
+
+
+# Type alias for supported backend names
+BackendType = Literal["pandas", "polars"]
```
mil-ad: not sure if this belongs here. Protocol shouldn't know or care about implementations. Maybe belongs to `dataset.py`?

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+
+from __future__ import annotations
+
+from typing import Any, Callable, Iterator, Literal, Protocol, Type, TypeVar, overload
+
+from .pandas_backend import PandasBackend
+from .polars_backend import PolarsBackend
+
+# Type alias for backend return type (allows chaining)
+DataBackendT = TypeVar("DataBackend")
```
mil-ad: I don't think this is what you want. Using `TypeVar` mostly makes sense with generics when you want to support multiple types but also want to enforce the same type for example between arguments and the return type.

for example:


```py
T = TypeVar('T')

def foo(blah: list[T]) -> T:
    return blah[0]
```

mil-ad: you can probably just use `Self`?

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+    instance attribute, making the API more Pythonic and cleaner.
+    """
+
+    @classmethod
+    def from_csv(
+        cls,
+        path: str,
+        *,
+        streaming: bool = False,
+        **kwargs: Any,  # noqa ANN401
```
mil-ad: is the noqa still necessary? We added this to pyproject:

```
[tool.ruff.flake8-annotations]
allow-star-arg-any = true
```

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+    This protocol uses the Adapter pattern where the backend wraps a data
+    and provides a unified interface. The wrapped data is stored as an
+    instance attribute, making the API more Pythonic and cleaner.
+    """
+
+    @classmethod
+    def from_csv(
+        cls,
+        path: str,
+        *,
```
mil-ad, line 25: 👌

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+
+    This protocol uses the Adapter pattern where the backend wraps a data
+    and provides a unified interface. The wrapped data is stored as an
+    instance attribute, making the API more Pythonic and cleaner.
+    """
+
+    @classmethod
+    def from_csv(
+        cls,
+        path: str,
```
mil-ad, line 24: shouldn't we support AnyPathT | str? if yes let's do it in a separate PR

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+        """Check if backend is in streaming mode.
+
+        Returns
+        -------
+        bool
+            True if in streaming mode, False otherwise
+        """
+        ...
+
+    @overload
```
mil-ad, line 123: I didn't know about `@overload`, cool! Does this mean the implementation should implement _one_ of these or _all_ of these?

esp_data/backends/protocol.py
```diff
@@ -0,0 +1,569 @@
...
+    def __getitem__(self, key: list[int]) -> DataBackendT:
+        """Get multiple rows by list of indices."""
+        ...
+
+    @overload
+    def __getitem__(self, key: slice) -> DataBackendT:
+        """Get rows by slice."""
+        ...
+
+    def __getitem__(self, key: int | list[int] | slice) -> dict[str, Any] | DataBackendT:
```
mil-ad: why not have an `@overload` here?

mil-ad (changes requested):
Have only looked at two files but submitting anyway so that I feel less bad about review delay 😬

esp_data/backends/protocol.py
GaganNarula: you're right, this isnt necessary, will remove `list_backends` and `get_backends`. Should we instead have an `available_backends` property in each dataset that the user can call to find out which backends exist ?

mil-ad: I think eventually yes but maybe let's do that when this whole idea of "backends" have matured a bit

esp_data/backends/polars_backend.py
```diff
@@ -0,0 +1,962 @@
...
+        **kwargs : Any
+            Additional polars-specific arguments
+
+        Returns
+        -------
+        PolarsBackend
+            Backend instance wrapping the loaded DataFrame or LazyFrame
+        """
+        # Filter out kwargs for any non-polars argument
+        if streaming:
```
mil-ad, line 102: fancy! Do we want to complain if there are kwargs we don't recognize?

mil-ad: it _might_be possible to leverage pydantic's `validate_call()` here instead. 

https://docs.pydantic.dev/latest/api/validate_call/#pydantic.validate_call_decorator.validate_call

It's meant to be used as a decorator but you can just do:

``` py
def from_csv(**kwargs):
   fn = validate_call(pd.from_csv)
   fn(kwargs)
```

not a big deal , just an idea

esp_data/backends/protocol.py
GaganNarula: all of them, its just a clean way of having several input types and return types implemented in one method.

esp_data/backends/protocol.py
GaganNarula: I've removed `list_backends` , but we do need the `get_backend` and `_BACKEND_REGISTRY` . `get_backend` is used in the `Dataset` classes to load the data based on the `backend` parameter. I've refactored a bit to create a new module for some type information and two keep these two objects there.

esp_data/backends/protocol.py
GaganNarula: Somehow ruff is ignoring this and keeps generating the `ANN003` error for `**kwargs` .. i tried moving this setting to the linter section as `[tool.ruff.lint.flate8-annotations] ...` to avoid a warning from the pre-commit, but it didnt help solve the issue. For now I had to put `**kwargs: object` in the code

esp_data/backends/protocol.py
GaganNarula: i thought since the `Datasets` are supporting `AnyPathT` at some point someone has to do a `str(path)` type statement, might as well do it in a dataset ? we can change it though, no biggie

esp_data/backends/protocol.py
GaganNarula: this was actually wrongly implemented. The non-overloaded method actually doesn't need any input argument or return type hints or the `overload` decorator. Corrected.

esp_data/backends/protocol.py
GaganNarula: I removed the typevar

esp_data/backends/polars_backend.py
GaganNarula: We dont complain because we want to allow specific arguments for specific backends to be allowed together. for e.g. in `animalspeak.py`:
```
self._data = self._backend_class.from_csv(
            location,
            streaming=self._streaming,
            infer_schema_length=10_000,
            keep_default_na=False,
            na_values=[""],
        )
```
Here, `infer_schema_length` is a polars parameter but `keep_default_na` and `na_values` are pandas params.

Good point for discussion.. is this an anti-pattern ? it could even more messy with different backends .. the other way to do it would be to use the `_backend_class` type / name to do:
```
if isinstance(self._backend_type, PandasBackend):
    ... do from_csv(**pandas_only_kwargs)
 elif  isinstance(self._backend_type, PolarsBackend):
    ... from_csv(**polars_only_kwargs)
    ....
```

this also seems ugly though

mil-ad: how about having something like `backend_kwargs` arg? I think that's better than having conflicting args

esp_data/backends/protocol.py
mil-ad: fair enough

esp_data/backends/protocol.py
mil-ad: ok I think this is fine for now

mil-ad (approved):

PR #132, opened 2025-10-02
Paul/benchmarking

Paulchen-git:
- Implementing saving process into CSV file on gs bucket
- changes made into a separate file named benchmark_latency.py


Paulchen-git:
There is an error in Pytests on raincoast dataset, I guess it has something to do with Maddie's PR.

Paulchen-git:
list[int] is not supported in click_option, I'll correct it
Also I take a look to the closing time of threads

Paulchen-git:
I need to solve the problem with bucket and num_worker before updating the benchmark_latency file

scripts/benchmarks/loading_time.py
```diff
@@ -0,0 +1,56 @@
...
+    type=str,
+    default="bucket",
+    help="Data location to benchmark (e.g., 'nfs' or 'bucket')",
+    show_default=True,
+)
+def main(
+    data_location: str,
+    config_path: Path,
+) -> None:
+    start_time = time.perf_counter()
```
mil-ad: We'll be doing this timing operations quite a lot. It might not be a bad idea to create some utility functions for them. This bit in particular lends itself well to be a decorator or context like:

```py
with timer() as t:
  build_raw_dataset(config_path, data_location)
```

It might also not be bad idea to look at [timeit](https://docs.python.org/3/library/timeit.html) module which implements a lot of these things

scripts/benchmarks/loading_time.py
mil-ad, line 1: This is great. Some ideas for future (not needed in this PR) regarding "loading times":

- We should slowly think about a way to run this experiment one by one on all datasets. And we should make sure we do them all with a cold start.
  - I'm not sure if esp_data has a method to give you list of datasets (we might already have a method that gets this from our dataset registry) but if not we can add that
- I'm also interested in benchmarking import times. It's easy but we just haven't done it
- Also it would be cool to measure "time to first sample". i.e. just do what you're doing in this file followed by a `next(dataset)`

scripts/benchmarks/loading_time.py
```diff
@@ -0,0 +1,56 @@
...
+        [
+            {
+                "data_location": data_location,
+                "loading_time_seconds": elapsed_time,
+                "timestamp": pandas.Timestamp.now(),
+            }
+        ]
+    )
+
+    save_results(all_results, csv_path, logger)
```
mil-ad: we should probably save the actual config yaml file too?

scripts/benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,216 @@
...
+    # Get bucket metadata
+    bucket = client.bucket(bucket_name)
+    bucket.reload()  # Ensure we have the latest metadata
+
+    return bucket.location
+
+
+def get_machine_location(
+    logger: logging.Logger,
+) -> str:
```
mil-ad: maybe return zone and region separately

scripts/benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,216 @@
...
+    client = storage.Client()
+
+    # Get bucket metadata
+    bucket = client.bucket(bucket_name)
+    bucket.reload()  # Ensure we have the latest metadata
+
+    return bucket.location
+
+
+def get_machine_location(
```
mil-ad: this is very GCP VM specific so I'd rename to something like `get_gcp_instance_location()`

scripts/benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,216 @@
...
+        persistent_workers=(num_workers > 0),
+    )
+
+    return loader
+
+
+def save_results(
+    results: pandas.DataFrame,
+    output_path: Path,
+    logger: logging.Logger,
```
mil-ad: you don't need to pass logger object around. In Python logging module you can get access to the same object by passing the same name to `logging.getLogger()`. 

In the actual esp_data library we always do `logger = logging.getLogger("esp_data")`. This is because python logging is hierarchical and as a library we should only configure our logger and let the application deal with it. 

The benchmark code lives outside the library so you can pretty much do whatever you want. You already configure a root logger so you can do 

`logger = logging.getLogger("benchmark_dataloader")` anywhere you need it

scripts/benchmarks/benchmark_latency.py
```diff
@@ -0,0 +1,378 @@
...
+
+    # Define the CSV file path in the bucket
+    csv_path = "gs://esp-ci-cd-tests/esp-data-tests/benchmark_dataset/benchmark_dataset.csv"
+
+    # Save all results to the single CSV file in the bucket
+    save_results(all_results, csv_path, logger)
+
+
+if __name__ == "__main__":
+    mp.set_start_method("spawn", force=True)
```
mil-ad, line 398: add a lot of comments here describing _why_ we're doing this

mil-ad (changes requested):
Good start! Left some small comments

scripts/benchmarks/loading_time.py
GaganNarula: agree

scripts/benchmarks/benchmark_prefetch.py
```diff
@@ -0,0 +1,322 @@
...
+    plt.ylabel("Samples per Second")
+    plt.grid(True)
+    plt.subplot(2, 1, 2)
+    plt.plot(elapsed_time_list, n_batches_list, marker="o")
+    plt.title("Batches vs Elapsed Time")
+    plt.xlabel("Elapsed Time (s)")
+    plt.ylabel("Batches")
+    plt.grid(True)
+    plt.tight_layout()
+    plt.savefig(f"benchmark_{name}_prefetch_{prefetch_factor}.png")
```
GaganNarula: this figure will be saved locally to the repo location, its better to save it in the same place as the csv..  
```
with filesystem_from_path(csv_bucket_path / 'benchmark_figure.png') as f:
    plt.savefig(f, ... <other options>)
```
generally, keep the computation / tasks of a single function limited. It would be better to move this figure plotting code outside this function, in the main() or somewhere else

scripts/benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,216 @@
...
+    output_path : Path
+        Path to save the CSV file.
+    logger : logging.Logger
+        Logger for logging messages.
+    """
+
+    try:
+        existing = pandas.read_csv(output_path)
+        results = pandas.concat(
+            [existing, pandas.DataFrame([{}]), pandas.DataFrame([{}]), results], ignore_index=True
```
GaganNarula: i'm confused by the concat of two empty dataframes in `pandas.DataFrame([{}]), pandas.DataFrame([{}]` bit.. what does this do ?

Paulchen-git: That was a mistake — I forgot to remove those empty dataframes. I had used them in two consecutive test jobs when experimenting with the CSV.

scripts/benchmarks/benchmark_latency.py
```diff
@@ -0,0 +1,392 @@
...
+                shard_download_time = t0.elapsed()
+                t0.start = time.perf_counter()  # Reset start time after first sample
+
+            with Timer("sample_time", store=None) as t1:
+                # Log stats every log_interval samples
+                if n_samples > 1 and (n_samples % log_interval == 0):
+                    elapsed_since_start = t0.elapsed()
+                    elapsed_since_last = t1.start - last_log_time
+
+                    running_avg_samples_per_sec = 0.99 * running_avg_samples_per_sec + 0.01 * (
```
GaganNarula: We can remove this running_avg thingy... we're not storing the result anyway

scripts/benchmarks/benchmark_latency.py
```diff
@@ -0,0 +1,392 @@
...
+
+    # Save all results to the single CSV file in the bucket
+    save_results(all_results, csv_path)
+
+
+if __name__ == "__main__":
+    # Use 'spawn' to be compatible with DataLoader multiprocessing.
+    # When running the benchmark with data_location set on 'bucket'
+    # if we don't use the following line we get a freeze for num_workers > 0.
+    mp.set_start_method("spawn", force=True)
```
GaganNarula, line 398: Let's add a link to torch multiprocessing documentation here. Also, another start method is "fork_server" that we haven't looked at

scripts/benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,233 @@
...
+    Returns
+    -------
+    Dataset
+        The raw dataset.
+    """
+    with config_path.open("r", encoding="utf-8") as fh:
+        raw = yaml.safe_load(fh)
+
+    dataset_config = DatasetConfig.model_validate(raw[data_location])
+    dataset, _ = dataset_from_config(dataset_config)
```
GaganNarula: I think in the latest esp-data main branch you don't need to first load the DatasetConfig .. you can directly give the path to the config to the dataset_from_config function

scripts/benchmarks/benchmark_prefetch.py
```diff
@@ -0,0 +1,359 @@
+"""Benchmark DataLoader Prefetch Parameter.
+
+This script benchmarks the performance of PyTorch DataLoaders with different
+prefetch settings.
+
```
GaganNarula: Why do we need a separate benchmark script for prefetch testing ? Can the logic be combined in the benchmark_latency.py script?

GaganNarula (changes requested):
Almost there! :) A few more comments

scripts/benchmarks/benchmark_latency.py
Paulchen-git: It's now removed !

scripts/benchmarks/benchmark_latency.py
Paulchen-git: I added the link. I'll take a look at "fork_server" method.

scripts/benchmarks/benchmark_utils.py
Paulchen-git: Yes, I looked at the code and I've modified the function to take this into account

scripts/benchmarks/benchmark_prefetch.py
Paulchen-git: I'll try to combined it but it's not corrected yet.
Should I add a click options to make it possible to choose the specific parameter we want to test ?
Because if I run a loop over each combination of many parameters it takes a long time to run. (That's why I've separated it first)

Paulchen-git: It is now combined with a click option to choose the prefetch factor

mil-ad (approved):
LGTM

GaganNarula (approved):
Looks great, i think we'll need to do some more digging on the multiprocessing start method, but we can add those comments in another PR

PR #135, opened 2025-10-15
Wabad detection

benjaminsshoffman:
Here's a first pass at including WABAD. Running the file starts a quality check, which the dataset passes.

Note that there is some funny business with the common names from the GBIF app, see slack message.


pyproject.toml
```diff
@@ -17,6 +17,9 @@ dependencies = [
     "s3fs>=2024.9.0",
     "semver",
     "soundfile>=0.13.1",
+    "torch>=2.8.0",
+    "torchaudio>=2.8.0",
+    "tqdm>=4.67.1",
```
mil-ad: This seems to be only used in the `__main__` which is not how the dataset is used so I'd remove it as a dependency.

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,436 @@
...
+        if np.any(np.isnan(audio)):
+            print(f"nan present in {audio_path}")
+        if np.all(audio == 0):
+            print(f"clip is all zeros {audio_path}")
+        st_end = st["Begin Time (s)"].max()
+        audio_end = len(audio) / ds.sample_rate
+        if st_end > audio_end:
+            print(f"events happen after audio in {audio_path}")
+
+    print("Done checking dataset")
```
mil-ad: remove and add unit tests

pyproject.toml
mil-ad: That part should be done as unit tests and I don't think you'd need tqdm when running unit tests (because pytest usually hides stdout) but if needed add it in the `dev` group

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,436 @@
...
+        if self.sample_rate is not None and sr != self.sample_rate:
+            audio = torchaudio.functional.resample(
+                torch.tensor(audio),
+                sr,
+                self.sample_rate,
+                lowpass_filter_width=64,
+                rolloff=0.9475937167399596,
+                resampling_method="sinc_interp_kaiser",
+                beta=14.769656459379492,
+            ).numpy()
```
mil-ad: is this meant to always run on the CPU or do we want to support GPU versions?

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,436 @@
...
+
+        if self.sample_rate is not None and sr != self.sample_rate:
+            audio = torchaudio.functional.resample(
+                torch.tensor(audio),
+                sr,
+                self.sample_rate,
+                lowpass_filter_width=64,
+                rolloff=0.9475937167399596,
+                resampling_method="sinc_interp_kaiser",
+                beta=14.769656459379492,
```
mil-ad: could you document this magic values? it might make sense to put this whole resampling logic into utils? is it reusable or very specific to this dataset?

mil-ad (changes requested):

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,436 @@
...
+    """
+
+    info = DatasetInfo(
+        name="wabad",
+        owner="benjamin",
+        split_paths={"all": "gs://esp-ml-datasets/wabad/v0.1.0/raw/all_info.csv"},
+        version="0.1.0",
+        description="[MISSING]",  # Redundant with docstring
+        sources="zenodo.org",
+        license="Creative Commons Attribution 4.0 International",
```
GaganNarula: better just to keep a code like "CC-BY-4.0"

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,436 @@
...
+        if os.path.exists(info_fp):
+            print("Reloading GBIF taxonomy from cache")
+            with open(info_fp, "r") as f:
+                json.load(f)
+
+        else:
+            print("Checking against GBIF/animalspeak taxonomy")
+            import requests
+
+            base_url = "http://gagan-dev:8000"
```
GaganNarula: this kind of preprocessing code to get taxonomy shouldn't be here. These Dataset classes are just meant to access data thats already prepared. Best thing to do:

1. Create a preprocessing script, add this bit and the full validation loop in __main__ to it.
2. Push the script to the scripts/data_preprocessing_scripts/ folder in the `esp-data_ingestion` branch. I know its kind of ugly but we don't have a place for all the preprocessing scripts.

esp_data/datasets/wabad.py
GaganNarula: I think benj mentioned that these are default values of the `librosa.resample` method so i think just that comment is fine.

I would be in favor of adding resampling methods as functions that are called from a factory function to the io/read_utils.py. This one here can be one of them. But lets do it in a following PR?

esp_data/datasets/wabad.py
GaganNarula: ideally it should be in `tests/test_wabad.py`.  Also, doing a full loop in the unittests is probably going to slow down the CI a lot, so I suggest just a random sample of 5 indices or so. This relates to my comment earlier about the data preprocessing part.

pyproject.toml
benjaminsshoffman: Ok, removed in next commit

esp_data/datasets/wabad.py
benjaminsshoffman: I was following the previous datasets, which output numpy arrays and operate on the CPU. So, the answer to your question is it was meant to run on CPU.

Probably copying to GPU isn't efficient unless we're doing batched resampling (I think?), but that would be a big change in how things are set up.

esp_data/datasets/wabad.py
benjaminsshoffman: Yeah these values are meant to match librosa.resample, they are taken from https://docs.pytorch.org/audio/stable/tutorials/audio_resampling_tutorial.html. I have added this documentation to the next commit.

I'll assume `read_utils.py` are your purview, but happy to help.

esp_data/datasets/wabad.py
benjaminsshoffman: Fixed in next commit

esp_data/datasets/wabad.py
benjaminsshoffman: I created a preprocessing script and data validation loop, and pushed it to esp-data_ingestion

esp_data/datasets/wabad.py
benjaminsshoffman: ok, done in new commit

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,298 @@
...
+import torchaudio
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, filesystem, read_audio
+
+LABEL_MAPPING_PATH = "gs://esp-ml-datasets/wabad/v0.1.0/raw/wabad_label_mapping_15_10_2025.json"
+
+
+@register_dataset
+class WABAD(Dataset):
```
mil-ad, line 20: can we rename to `Wabad` to follow pep-8?

https://peps.python.org/pep-0008/#class-names

benjaminsshoffman: WABAD is an acronym (World Annotated Bird Acoustic Dataset), so current naming follows pep-8:

"Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError."

https://peps.python.org/pep-0008/#descriptive-naming-styles

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,298 @@
...
+        self._load()
+
+        # Default label mapping path
+
+        label_mapping_fn = os.path.basename(LABEL_MAPPING_PATH)
+        label_mapping_path_local = Path(__file__).parent / label_mapping_fn
+        if not os.path.exists(label_mapping_path_local):
+            print(f"Getting {LABEL_MAPPING_PATH} from GCP")
+            fs = filesystem("gcs")
+            fs.get(LABEL_MAPPING_PATH, label_mapping_path_local)
```
GaganNarula: is there a reason to copy the file to local and not just load it into memory with Pandas ? Is it too large to fit in memory.. use `pd.read_json` to load it if needed.

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,298 @@
...
+        data_root : str | AnyPathT | None
+            Optional root directory to prepend to each row['audio_fp'].
+        """
+        super().__init__(output_take_and_give)
+        self.split = split
+        self._data: pd.DataFrame | None = None
+        self.default_anno_column = "Species"
+        self.unknown_annos: List[str] = []
+        self.annotation_columns = ["Genus", "Family", "Order", "Common", "Species"]
+        assert self.default_anno_column == self.annotation_columns[-1]
```
GaganNarula: no need for this assert

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,298 @@
...
+            Optional mapping of original → new output keys (filters columns as well).
+        sample_rate : int | None
+            If set, audio is resampled to this rate.
+        data_root : str | AnyPathT | None
+            Optional root directory to prepend to each row['audio_fp'].
+        """
+        super().__init__(output_take_and_give)
+        self.split = split
+        self._data: pd.DataFrame | None = None
+        self.default_anno_column = "Species"
```
GaganNarula: would this maybe be better as an argument to `WABAD(... , annotation_column="Species")` ?

FYI: we are moving to an approach where if a dataset that needs more than the standard input arguments to `__init__` it needs a specific config. For e.g., in this module you'd have to add a `WABADConfig(DatasetConfig)` class with these extra parameters. See the `voxaboxen.py` module for an example. Not saying you need it here, but if you want to add a param like `annotation_column` it would be needed.

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,298 @@
...
+    def columns(self) -> list[str]:
+        return list(self._data.columns) if self._data is not None else []
+
+    @property
+    def available_splits(self) -> list[str]:
+        return list(self.info.split_paths.keys())
+
+    # -----------------------------
+    # Internals
+    # -----------------------------
```
GaganNarula: Can we remove these Cursor things ?

esp_data/datasets/wabad.py
GaganNarula: of if you dont want to use pandas you can do:

```
with fs.open("gs://path/to/label_mapping.json") as f:
     label_map = json.load(f)
```

esp_data/datasets/wabad.py
GaganNarula: @mil-ad i am ok with adding `torchaudio` to esp-data. Wdyt ? Then i'll create a PR and move this resampling code to `read_utils.py` (or similar)

tests/test_wabad.py
```diff
@@ -0,0 +1,91 @@
...
+    for idx in indices:
+        item = ds[idx]
+        issues = _basic_item_qc(item)
+        all_issues.extend([(idx, i) for i in issues])
+
+    # If you want this to be strictly clean, assert no issues:
+    # assert not all_issues, f"Issues found: {all_issues}"
+    # Instead: allow visibility without hard failing CI on minor glitches
+    if all_issues:
+        pytest.fail(f"Quality issues in random items: {all_issues}")
```
GaganNarula: this is a bit of strange way to do things.. I would prefer that each test is a small function itself, for e.g. `test_check_audio`, `test_check_selection_table`, `test_ds_not_empty`

esp_data/datasets/wabad.py
```diff
@@ -0,0 +1,298 @@
...
+
+if __name__ == "__main__":
+    ds = WABAD()
+
+    sample = ds[10]
+
+    assert "audio" in sample
+    print(f"Audio shape: {sample['audio'].shape}")
+    st1 = sample["selection_table"]
+    print(st1.head(10))
```
GaganNarula: This bit here is not needed (lines 290 - 298), you're covering it in unit tests

esp_data/datasets/wabad.py
benjaminsshoffman: The reason was I didn't know i could do this! Fixed

esp_data/datasets/wabad.py
benjaminsshoffman: It's important that "Species" is last in this list, and I wanted to ensure that future edits preserved this property. Changed to a comment.

esp_data/datasets/wabad.py
benjaminsshoffman: This should not be user-defined, it defines the column that the label mapping looks at for remapping labels; see usage in `__get_item__`

esp_data/datasets/wabad.py
benjaminsshoffman: Removed

esp_data/datasets/wabad.py
mil-ad: then can we make sure we only add CPU-version of torch to the dependencies?

Please follow uv's guide for adding torch indices: https://docs.astral.sh/uv/guides/integration/pytorch/

GaganNarula (changes requested):

esp_data/datasets/wabad.py
benjaminsshoffman: removed

esp_data/datasets/wabad.py
benjaminsshoffman: fixed in next commit

tests/test_wabad.py
benjaminsshoffman: fixed these tests

GaganNarula (approved):
Lgtm!

mil-ad:
@benjaminsshoffman just one thing to check before merging this: people will use this as a library so can you just make sure that the torch dependency we've added works if people have their own gpu torch installed in their project.

I'd basically create a new project somewhere, add torch GPU with pip or whatever, and then try to `uv add` this branch of esp-data and confirm that dependencies are resolved succesfully

esp_data/datasets/wabad.py
mil-ad: Sounds good, maybe just add a comment here that says these numbers are from librosa. I just left one comment about torch/torchaudio dependency. Assuming that test passes I'm okay with it

benjaminsshoffman: Gah I'd added it and then the AI deleted it. Added again >.<

benjaminsshoffman:
@mil-ad importing did break dependencies (i.e. opted into the cpu-only version for the importing package).

I'm wondering if it's ok to simply have the non-pinned version of torch in the esp-data dependencies? It seemed to run fine on my dev machine and on github CI when that was the case (both of which don't have a GPU).

I don't have a great understanding of how uv works...I tried to figure out a way to have it default to cpu, and then use importing package's version otherwise. It seemed to involve extra steps for the importing package:

This is what the oracle proposed:

(in esp-data)

```
# --- uv (Astral) resolver configuration ---
# These mappings apply only when *this* project is being resolved.
# If another project depends on `mypkg`, *their* resolver & indexes win.
[tool.uv.index]
# Define named indexes for PyTorch variants we might want to target.
"pytorch-cpu"   = "https://download.pytorch.org/whl/cpu"
"pytorch-cu126" = "https://download.pytorch.org/whl/cu126"

[tool.uv.sources]
# Force *default resolution for this project* to CPU wheels for torch/torchvision.
# This guarantees CPU-by-default on Linux, macOS, and Windows when users do:
#   uv pip install .   or   uv add mypkg
torch       = { index = "pytorch-cpu" }
torchvision = { index = "pytorch-cpu" }

# (Optional) expose an extra that *declares intent* for GPU, but still
# relies on the *calling environment* to map torch to a CUDA index.
[project.optional-dependencies]
gpu = [
  "torch",
  "torchvision",
]
```

(in importing package)

```
[tool.uv.index]
"pytorch-cu126" = "https://download.pytorch.org/whl/cu126"

[tool.uv.sources]
torch       = { index = "pytorch-cu126" }
torchvision = { index = "pytorch-cu126" }
```

This seemed undesirable since it requires pinning a specific version in the importing package.

benjaminsshoffman:
@mil-ad In the interest of getting things moved along, and since it sounds like torchaudio resampling as a utility is planned for a future PR, I'm removing the torchaudio resampling from this and using librosa. I think that should resolve the remaining requests that I couldn't figure out.

mil-ad (approved):
LGTM

PR #140, opened 2025-10-16
Enable UP007 and UP045 rules in ruff

mil-ad:
ruff auto-fixed all the issues but I had to add an exception in `esp_data/transforms/registry.py` and revert the change because we explicitly need a `Union`.

### DIFF
--- a/esp_data/concat.py
+++ b/esp_data/concat.py
@@ -1,7 +1,7 @@
 """Dataset concatenation utilities."""
 
 import logging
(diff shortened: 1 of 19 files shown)


GaganNarula (approved):
very nice addition in ruff!

mil-ad:
FYI @GaganNarula I have added this to the template too

https://github.com/earthspecies/python-repo-template/pull/24

PR #142, opened 2025-10-20
add subsegmentation

benjaminsshoffman:
This is the subsegmentation dataset, provided by logan and used for:
- subsegmentation intern project (which we need an intern for)
- as a dataset for the common calls project


GaganNarula:
Hi benj, I see that some of the changes made to the wabad PR would also apply here, (plus the conflicts). Will review when we merge that PR

benjaminsshoffman:
@GaganNarula I've removed torchaudio dependencies in the interest of getting things moved along (since I understand that torchaudio resampling as a utility is planned for a future PR). I removed these dependencies in the wabad PR as well. I've also resolved the merge conflicts. So it should be ok to review now.

esp_data/datasets/subsegmentation.py
```diff
@@ -0,0 +1,211 @@
...
+        output_take_and_give : dict[str, str] | None
+            Optional mapping of original → new output keys (filters columns as well).
+        sample_rate : int | None
+            If set, audio is resampled to this rate.
+        data_root : str | AnyPathT | None
+            Optional root directory to prepend to each row['audio_path'].
+        """
+        super().__init__(output_take_and_give)
+        self.split = split
+        self._data: pd.DataFrame | None = None
```
mil-ad: same comment as here: https://github.com/earthspecies/esp-data/pull/151#discussion_r2473142599

tests/test_subsegmentation.py
mil-ad, line 1: same comment as https://github.com/earthspecies/esp-data/pull/151#discussion_r2473146016

would be great to have one test that kind of detects if the underlying data in the buckets change

mil-ad (commented):
small comments

esp_data/datasets/subsegmentation.py
benjaminsshoffman: Fixed

tests/test_subsegmentation.py
benjaminsshoffman: added these tests

esp_data/datasets/subsegmentation.py
```diff
@@ -0,0 +1,204 @@
...
+        split_paths={
+            "all": "gs://subsegmentation/xeno_canto_annotations/all.csv",
+            "train": "gs://subsegmentation/xeno_canto_annotations/train.csv",
+            "val": "gs://subsegmentation/xeno_canto_annotations/val.csv",
+            "test": "gs://subsegmentation/xeno_canto_annotations/test.csv",
+        },
+        version="0.1.0",
+        description="[MISSING]",
+        sources="Logan James",
+        license="Internal (currently)",
```
GaganNarula: lets call these "private"

GaganNarula (approved):
I think some conflicts need to be resolved but otherwise looks good

PR #144, opened 2025-10-24
add iNaturalist dataset

david-rx:
Adding the iNaturalist dataset as a separate dataset in esp-data.

This points to a new version downloaded in mid-2025. It uses the original sample rates (unlike the version available via the "AnimalSpeak" dataset.) It includes a copy resampled in advance to 32khz, which will be used if the desired sampling rate matches (specified via "available_sample_rates".)

Versioning for the future is still an open question for the moment, as this dataset will be routinely updated. However, the path to the metadata is delimited as gs://esp-ml-datasets/inaturalist/v0.1.0/raw/metadata_{year}_{month}.csv corresponding to when the data was downloaded. In subsequent versions, I expect we could just have a new metadata file.

This adds over 300,000 audio files compared to the version of iNaturalist included in esp-data through the AnimalSpeak dataset.


esp_data/datasets/inaturalist.py
```diff
@@ -0,0 +1,257 @@
...
+        if self.split not in self.info.split_paths:
+            raise LookupError(
+                f"Invalid split: {self.split}."
+                "Expected one of {list(self.info.split_paths.keys())}"
+            )
+
+        location = self.info.split_paths[self.split]
+        # Read CSV content
+        csv_text = read_text(location, encoding="utf-8")
+        self._data = pd.read_csv(StringIO(csv_text), low_memory=False)
```
GaganNarula: We're finding that doing read_text(path) first before using pandas is causing issues with memory OOM kills. Prefer using pandas `pd.read_csv("bucket/path/to/file")` directly #145

GaganNarula:
I think since we don't have a proper versioning library / tool right now my suggestion would be the following: 

If a version (say 0.1.0) has more than one sample rate, it's still the same version but a different sampling rate (we should document how it was resampled, can add some info to the DatasetInfo description field?. So I say that the we add some private attributes to the dataset which are audio paths to different sampling rates. Then, we also add a property to the Dataset called "available_sample_rates" which lists the available SRs. When a user specifies a sample rate, if it's already available, then they audio path is used, and if it is not, the audio is resampled on the fly. Keep one SR as the default !

david-rx:
> I think since we don't have a proper versioning library / tool right now my suggestion would be the following:
> 
> If a version (say 0.1.0) has more than one sample rate, it's still the same version but a different sampling rate (we should document how it was resampled, can add some info to the DatasetInfo description field?. So I say that the we add some private attributes to the dataset which are audio paths to different sampling rates. Then, we also add a property to the Dataset called "available_sample_rates" which lists the available SRs. When a user specifies a sample rate, if it's already available, then they audio path is used, and if it is not, the audio is resampled on the fly. Keep one SR as the default !

Thanks - I've gone with this approach!

esp_data/datasets/inaturalist.py
```diff
@@ -0,0 +1,309 @@
...
+    ...     output_take_and_give={"canonical_name": "species"}
+    ... )
+    >>> print(dataset.info.name)
+    inaturalist
+    >>> print(dataset.available_sample_rates)
+    [32000]
+    >>> # Load with pre-resampled 32kHz audio (no on-the-fly resampling needed)
+    >>> dataset_32k = INaturalist(split="train", sample_rate=32000)
+    >>> # Load with on-the-fly resampling to 16kHz from original (variable rate) files
+    >>> dataset_16k = INaturalist(split="train", sample_rate=16000)
```
GaganNarula, line 82: we should add a short test  suite `test_inaturalist.py`. Especially to make sure that the split paths to csvs and available sample rate paths always exist.

GaganNarula (changes requested):

esp_data/datasets/inaturalist.py
david-rx: Yes!

esp_data/datasets/inaturalist.py
```diff
@@ -0,0 +1,343 @@
...
+            List of sample rates (in Hz) for which pre-resampled audio is available.
+            Audio at these sample rates can be loaded directly without on-the-fly resampling.
+            This checks which path columns actually exist in the loaded data.
+        """
+        available = []
+        for sr, path_column in self._sample_rate_paths.items():
+            # Check if the path column exists in the loaded data
+            if path_column in self._data.columns:
+                available.append(sr)
+        return available
```
GaganNarula, line 171: Do we want to add a column to the csv ? Seems like it would be a column of duplicates ? I was thinking we just have the relative path of the audio in the csv, and then the parent would be the sample rate saved as an attribute. So then you just prepend the parent string value to the audio relative path

tests/test_inaturalist.py
```diff
@@ -0,0 +1,405 @@
...
+
+    Returns
+    -------
+    Dataset
+        An instance of the iNaturalist dataset.
+    """
+    ds = INaturalist(split="train")
+    return ds
+
+
```
GaganNarula, line 35: Benj has added a couple of audio hash checks to his recent dataset PRs. Have a look at xeno-canto JeanTet for e.g .. I think we should continue that approach so we know that the audio loaded from a location is what's expected

GaganNarula (commented):
Some more comments ! Almost there

esp_data/datasets/inaturalist.py
david-rx: So the column does not fully contain duplicates: specifically, there are many .mp3 files in the originals. After resampling, rather than doing a lossy re-encoding back to .mp3, I just saved them as .wav files, because I thought it should better or exactly match the audio you get when you do live resampling in esp-data. So, some files referenced in this column have a different extension as well as parent folder. Open to options here (it's easiest not to re-run resampling, but it can be done if needed)

tests/test_inaturalist.py
david-rx: Added!

esp_data/datasets/inaturalist.py
GaganNarula: understood!

GaganNarula (approved):
Looks good to me, needs a rebase before merge i think

PR #146, opened 2025-10-24
Paul/expanding benchmark to other datasets

Paulchen-git:
- expand loading time.py and  benchmark_latency.py to all datasets (almost)

- add a memory_profiler to track the memory usage of building a dataset.

- add plot script.


Paulchen-git:
What I'll do for the next PR : 
- make the benchmark easy to run for others (one job that runs experiments and visualizes the results) -> especially for the benchmark_latency.py (num_workers, prefetch, job array, batch_size, ...) -> line plots across parameters would be interesting

- loading_time.py is meant to be run regularly -> write a job to run it and visualize it -> boxplots are more adapted to compare datasets or compare different configuration of a specific dataset

- documentation to run easily benchmarks

jobs/loading_time.sh
```diff
@@ -3,19 +3,35 @@
...
 
-echo "=== Finished job at $(date) ==="
+echo "=== Starting job at $(date) ==="
 
-# resubmit this same script for 1 hours later
-sbatch --begin=now+1hours $0
\ No newline at end of file
+cd "$WORKDIR"
+uv sync
+srun uv run --with torch python $SCRIPT \
```
GaganNarula: re what milad said on friday, it might be better to make a `benchmark` group in the pyproject toml with `torch` and `torchaudio`, then also add something like:

```
[tool.uv]
default-groups = ["dev"]
```
so that the benchmark group is not considered default, (default groups are installed with `uv sync`)

jobs/lt.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+#SBATCH --partition=cpu
+#SBATCH --job-name=loading_time_benchmark
+#SBATCH --output=/home/%u/outputs/loading_time/%A.log
+#SBATCH --mail-type=FAIL
+
+cd ~/esp-data
+uv sync
+
+srun uv run --with torch python scripts/benchmarks/generate_list.py
```
GaganNarula: this list creator doesn't have to be run with `srun` right ? and it doesnt need `--with torch`

jobs/multi_dataset_loading_time.sh
```diff
@@ -0,0 +1,35 @@
...
+
+echo "=== Starting job at $(date) ==="
+
+for DATASET in "${DATASETS[@]}"; do
+    # # trim whitespace and ignore empty lines or comments
+    # DATASET="$(echo "$DATASET" | xargs)"
+    # [[ -z "$DATASET" || "${DATASET:0:1}" == "#" ]] && continue
+
+    echo "Running benchmark for dataset: $DATASET"
+    sbatch jobs/loading_time.sh --dataset "$DATASET"
```
GaganNarula, line 31: will we add a similar job for latency ?

jobs/loading_time.sh
mil-ad: `dev` is the only default group by default anyway 

image

scripts/benchmarks/benchmark_latency.py
```diff
@@ -76,24 +76,23 @@ def benchmark_loader(
...
 
-            if n_batches == 1:
-                shard_download_time = t0.elapsed()
-                t0.start = time.perf_counter()  # Reset start time after first batch
-
             if sleep > 0:
                 time.sleep(sleep)
 
+            if n_batches == 1:
+                shard_download_time = t0.elapsed()[0]
```
GaganNarula: the name `shard_download_time` was used in a different context when I was testing a sharded data format like tar and parquet.. we can replace it with `first_batch_time`, which might be a useful metric to track

scripts/benchmarks/benchmark_latency.py
```diff
@@ -292,13 +291,30 @@ def benchmark_raw_dataset(
     show_default=True,
     help="Batch size for DataLoader (default: 128)",
 )
+@click.option(
+    "--from-config",
+    is_flag=True,
+    help="Use configuration from file instead of command line options",
```
GaganNarula: i don't think you need this flag.. instead check if both the `config_path` or `dataset_name` are supplied, if both are given raise error, otherwise use the one that is provided

scripts/benchmarks/benchmark_latency.py
```diff
@@ -347,6 +368,8 @@ def main(
 
     else:
         prefetch_factor = None if num_workers == 0 else prefetch_factor
+        if prefetch_factor == 0:
```
GaganNarula: might as well check if `prefetch_factor <= 0`

scripts/benchmarks/benchmark_utils.py
```diff
@@ -137,7 +160,7 @@ def build_dataloader(
         num_workers=num_workers,
         collate_fn=collate_fn,
         prefetch_factor=prefetch_factor,
-        persistent_workers=(num_workers > 0),
+        persistent_workers=False,
```
GaganNarula, line 179: good catch

scripts/benchmarks/benchmark_utils.py
GaganNarula: the new `io.paths` module has some a `PureGCSPath` that can give you the bucket name if you like

scripts/benchmarks/loading_time.py
```diff
@@ -34,49 +43,120 @@
...
-                "nominal_speed (samples/second)": 10.0 / (times_stored["time_for_10_samples"]),
-                "timestamp": pandas.Timestamp.now(),
-                **raw_config,
-            }
-        ]
-    )
+    # Shape of a sample
+    sample = ds[0]
+    if "audio" in sample:
+        shape_audio = sample["audio"].shape[0]
```
GaganNarula: hmm i guess we need sample_rate info somehow... otherwise simply having the length wont help.. ok for now 👍

jobs/loading_time.sh
GaganNarula: got it thanks

GaganNarula (changes requested):
looking good, just a few comments

jobs/multi_dataset_loading_time.sh
Paulchen-git: If needed we can yes. First I'll make a job that runs a loop to try different configurations (num_workers/batch_size/prefetch_factor and later job array) on one specific dataset and creates line plots to see the evolution of latency with these parameters.
Then I'll write another job to run the previous one on all datasets. 

But do we want to benchmark the default dataset configurations ? or do we want to avoid resampling on fly while doing this benchmark for example ?(I could also measure the impact of resampling if needed)

jobs/lt.sh
Paulchen-git: I thought it was mandatory to use srun, but yes you're right it doesn't need torch

jobs/loading_time.sh
Paulchen-git: Once I created the benchmark group, how should I launch a script with it ?

scripts/benchmarks/loading_time.py
Paulchen-git: Okay then I'll remove it for now

jobs/loading_time.sh
GaganNarula: in this script, you change `uv sync` to `uv sync --group benchmark` then continue as usual

jobs/multi_dataset_loading_time.sh
GaganNarula: ah i think that we want to do it:
A) without resampling
B) with resampling

I would suggest that we change the `generate_list`function to instead create a long yaml with every dataset in our datasets, but with the `sample_rate` argument set to `null` in one yaml (A), and to some value like 16000 in the second (B)

But we can do this in a second PR if its too much wdyt @mil-ad

scripts/benchmarks/benchmark_utils.py
```diff
@@ -53,28 +54,65 @@ def elapsed(self) -> float:
...
+        # sample_rate:
+        # data_root:
+        raw = {}
+        raw["dataset_name"] = dataset_name
+        if hasattr(dataset, "split"):
+            raw["split"] = dataset.split
+        if hasattr(dataset, "sample_rate"):
+            raw["sample_rate"] = dataset.sample_rate
+        if hasattr(dataset, "data_root"):
+            raw["data_root"] = dataset.data_root
```
GaganNarula: I think all datasets should have these attributes (as of now).. if one does not, it should actually throw an error imo

esp_data/__init__.py
```diff
@@ -5,6 +5,7 @@
     DatasetConfig,
     DatasetInfo,
     dataset_from_config,
+    get_class_from_name,
```
mil-ad: this name is too generic. I'd rename to `dataset_class_from_name`

pyproject.toml
```diff
@@ -20,6 +20,11 @@ dependencies = [
 ]
 
 [dependency-groups]
+benchmark = [
+    "matplotlib>=3.10.7",
+    "memory-profiler>=0.61.0",
+    "torch>=2.9.1",
```
mil-ad, line 26: `torch` is a funny package. Let's add uv's specific index to make sure we're consistent

See https://docs.astral.sh/uv/guides/integration/pytorch/

basically we want to add something like this:

```
[tool.uv.sources]
torch = [
  { index = "pytorch-cpu", marker = "sys_platform != 'linux'" },
  { index = "pytorch-cu128", marker = "sys_platform == 'linux'" },
]

[[tool.uv.index]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
explicit = true

[[tool.uv.index]]
name = "pytorch-cu128"
url = "https://download.pytorch.org/whl/cu128"
explicit = true
```

scripts/benchmarks/generate_list.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def main() -> None:
+    """Print available datasets for benchmarking and write configs."""
+    registry = list_registered_datasets()
+    print("Available datasets for benchmarking:")
+    print(registry)
+
+    # Save registry in a dataset_list.txt file
+    with open("scripts/benchmarks/dataset_list.txt", "w", encoding="utf-8") as f:
```
mil-ad, line 16: maybe add this path to `.gitignore` to make sure we don't commit it by mistake

mil-ad (approved):
LGTM

jobs/loading_time.sh
Paulchen-git: Done

scripts/benchmarks/benchmark_latency.py
Paulchen-git: Done

scripts/benchmarks/benchmark_latency.py
Paulchen-git: Done

scripts/benchmarks/benchmark_latency.py
Paulchen-git: Done

scripts/benchmarks/benchmark_utils.py
Paulchen-git: I change it thanks

scripts/benchmarks/generate_list.py
Paulchen-git: Done

pyproject.toml
Paulchen-git: Done

esp_data/__init__.py
Paulchen-git: Done

scripts/benchmarks/benchmark_utils.py
Paulchen-git: You're right -> I removed the if conditions

GaganNarula (approved):
🚀

PR #149, opened 2025-10-28
Powdermill detection dataset

benjaminsshoffman:
This PR adds powdermill detection dataset. It is one of several used for sound event detection evaluation

### DIFF
--- a/esp_data/datasets/powdermill.py
+++ b/esp_data/datasets/powdermill.py
@@ -197,9 +197,11 @@ class Powdermill(Dataset):
             return ds, meta
         return ds, {}
 
-    def get_available_labels(self, anno_column: str) -> List[str]:
+    def get_available_labels(self, anno_column: str = "Species") -> List[str]:
         """
         Return all possible labels for a given annotation column
+        anno_column is included as an optional argument for consistency
+        with other detection datasets.
 
         Returns
         ---------
--- a/esp_data/datasets/powdermill.py
+++ b/esp_data/datasets/powdermill.py
@@ -1,0 +1,223 @@
+"""Powdermill dataset"""
+
+from __future__ import annotations
+
+from io import StringIO
+from typing import Any, Dict, Iterator, List
+
+import librosa
+import numpy as np
+import pandas as pd
+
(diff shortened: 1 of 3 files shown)


esp_data/datasets/powdermill.py
```diff
@@ -0,0 +1,223 @@
...
+            output_take_and_give=cfg["output_take_and_give"],
+            data_root=cfg["data_root"],
+            sample_rate=cfg["sample_rate"],
+        )
+        if dataset_config.transformations:
+            meta = ds.apply_transformations(dataset_config.transformations)
+            return ds, meta
+        return ds, {}
+
+    def get_available_labels(self, anno_column: str) -> List[str]:
```
GaganNarula: i think if the annotation_column is just one possible choice ("Species") then we can call this method without arguments and just use `self.anno_column`

esp_data/datasets/powdermill.py
```diff
@@ -0,0 +1,223 @@
...
+    ----------
+    https://esajournals.onlinelibrary.wiley.com/doi/full/10.1002/ecy.3329
+
+    """
+
+    info = DatasetInfo(
+        name="powdermill",
+        owner="benjamin",
+        split_paths={
+            "all": "gs://fewshot/evaluation/raw/Powdermill/all_gbif.csv",
```
GaganNarula: can we move this dataset to `esp-ml-datasets` ? i'd like to start getting these into one place in a non-multi region bucket (which is slower than a us-central bucket because slurm nodes are in us-central)

benjaminsshoffman: done. I left a copy at the original location since it's small (3GB) and I don't want to break functionality. split paths updated in next commit

esp_data/datasets/powdermill.py
benjaminsshoffman: I added a default value ("Species"). I want to retain the signature for consistency with other detection datasets (e.g. WABAD)

benjaminsshoffman:
Added tests for dataset stability

GaganNarula (approved):

PR #150, opened 2025-10-28
Make cloud paths compatible with Python 3.12+

mil-ad:
Turns out relying on Python's internal APIs and implementation (pathlib in particular) is not a great idea. :(

Python 3.12 did a big rehaul of pathlib implementation which broke the interoperability with out PureCloudPath. I initially [tried](892a7a3ec529e9604f902af911241ded95437068) to make it work and do things differently based on Python version but that seems like something that will need patching again soon. 

Instead I de-coupled PureCloudPath from PurePath and made it its own independant thing. There's still interoperability with Path/PurePath (for example when joining with `/`). 

I also revamped the unit tests (and moved them into `io` dir) with Claude's help to cover more cases.


esp_data/io/paths.py
```diff
@@ -1,213 +1,867 @@
...
+        return f"{self.cloud_prefix}{self.bucket}"
+
+    @property
+    def root(self) -> str:
+        """The root component (/ if path extends beyond bucket).
+
+        Returns:
+            "/" if path has components after bucket, "" otherwise.
+        """
+        if not self._path.startswith(self.cloud_prefix):
```
GaganNarula: if `self` is a `PureCloudPath` then isn't `self._path` always supposed to start with a cloud prefix ? looks like we check for this in `_validate_path`

esp_data/io/paths.py
```diff
@@ -1,213 +1,867 @@
...
+        """Validate that a path is appropriate for this cloud type.
+
+        Args:
+            path: The path string to validate.
+
+        Raises:
+            ValueError: If the path is invalid for this cloud type.
+        """
+        if not path:
+            raise ValueError(f"Path must start with '{self.cloud_prefix}': ")
```
GaganNarula: i think we should raise a different error , that the path is empty

tests/io/test_paths_core.py
```diff
@@ -0,0 +1,576 @@
...
+        assert str(result) == "gs://bucket/folder/subfolder/file.txt"
+
+    def test_cloud_path_with_windows_path(self):
+        """Test joining cloud paths with WindowsPath preserves backslashes."""
+        from pathlib import PureWindowsPath
+
+        gs_path = PureGSPath("gs://bucket/folder")
+        windows_path = PureWindowsPath("subfolder\\file.txt")
+        result = gs_path / windows_path
+        assert str(result) == "gs://bucket/folder/subfolder\\file.txt"
```
GaganNarula: should this really be the case ? I think all cloud paths should have / slashes only

tests/test_cloud_paths.py
```diff
@@ -0,0 +1,576 @@
...
+        joined = bucket_only / "bar/text.txt"
+        assert str(joined) == "gs://my-bucket/bar/text.txt"
+
+    def test_gs_path_joining_with_absolute(self):
+        """Test joining with absolute path replaces path after bucket."""
+        base = PureGSPath("gs://my-bucket/folder")
+        joined = base / "/absolute/path"
+        assert str(joined) == "gs://my-bucket/absolute/path"
+
+
```
GaganNarula, line 374: maybe add a test for the `parents` property because it tests the `_ParentsSequence` implicitly too

GaganNarula (commented):
left some comments

esp_data/io/paths.py
mil-ad: no because `relative_to()` exists and we create them internally (by using `__new__`) and are not validated

esp_data/io/paths.py
mil-ad: fixed

tests/io/test_paths_core.py
mil-ad: right. I don't even care about Windows :D 

removed the test

tests/test_cloud_paths.py
mil-ad: done

esp_data/io/paths.py
GaganNarula: i'm not sure I understand the logic here. Per definition a `PureCloudPath` must have a `cloud_prefix`

esp_data/io/paths.py
```diff
@@ -1,213 +1,867 @@
...
+            other_str = str(other[0])
+        else:
+            other_path = self.__class__(*other)
+            other_str = str(other_path)
+
+        # Normalize both paths for comparison
+        self_str = self._path
+
+        # Ensure other_str ends with / for prefix matching (unless it's bucket-only)
+        if other_str.startswith(self.cloud_prefix):
```
GaganNarula: continuing the same conversation, why is `other` anything but a cloud path ? If `other` is a local path then `self` wont be relative_to it (we could move the conditional on line 704 to 693)

GaganNarula (commented):
couple more comments

esp_data/io/paths.py
mil-ad: ok so my point is that `relative_to()` creates PureCloudPath instances internally (via object.__new__()) that don't have the cloud prefix. We have a test for this:

```py
def test_cloud_path_relative_to_valid_base(self):
    """Test relative_to with valid base path."""
    path = PureGSPath("gs://bucket/folder/file.txt")
    rel = path.relative_to("gs://bucket/folder/")
    assert str(rel) == "file.txt"
```

so this is technically a `PureGSPath` that does not have a cloud prefix. But I agree that this is not nice. Should we just remove relative to?

GaganNarula: Ok now I got it 😅 well we should either remove it or return a pathlib.Path because if it is relative to a cloud path, it will return a non-cloud path (but which actually refers to something in a bucket 😆)

mil-ad: yeah, I think you're right. Let's remove 🚮

GaganNarula (changes requested):
Approved pending removal of `relative_to` :)

esp_data/io/paths.py
mil-ad: done in 4f1b83fe00c9ef598c5ee4d9e6549acc0bf443f6

GaganNarula (approved):
I think we need to add docs to io about removing support for cloudpathlib and what the new `anypath` can / cannot do

PR #151, opened 2025-10-28
xeno canto jeantet annotated 23

benjaminsshoffman:
Another dataset for detection evaluation


esp_data/datasets/xeno_canto_annotated_jeantet_23.py
```diff
@@ -0,0 +1,220 @@
...
+        output_take_and_give : dict[str, str] | None
+            Optional mapping of original → new output keys (filters columns as well).
+        sample_rate : int | None
+            If set, audio is resampled to this rate.
+        data_root : str | AnyPathT | None
+            Optional root directory to prepend to each row['audio_path'].
+        """
+        super().__init__(output_take_and_give)
+        self.split = split
+        self._data: pd.DataFrame | None = None
```
mil-ad: This is probably copied from somewhere else but I'd just remove it. It doesn't make much sense to have 

```py
 self._data: pd.DataFrame | None = None
```

when we're doing `self._load()` a few lines below.

There are if a `if self._data is None:` in other places too. Let's remove those too? `self._data` cannot be None and we should verify that in `__init__`.

tests/test_xeno_canto_annotated_jeantet_23.py
```diff
@@ -0,0 +1,71 @@
...
+def sample_indices(ds: XenoCantoAnnotatedJeantet23) -> List[int]:
+    """Deterministically choose up to 5 random indices for quick spot checks."""
+    n = len(ds)
+    rng = random.Random(23)
+    return [rng.randrange(n) for _ in range(min(5, n))]
+
+
+def test_ds_not_empty(ds: XenoCantoAnnotatedJeantet23):
+    """Dataset should have at least one example."""
+    assert len(ds) > 0, "Dataset appears empty"
```
mil-ad, line 69: maybe add a test that asserts actual expected length too? And even hash of an audio file or something to make sure this dataset is what we think it is.

mil-ad (changes requested):
Looks good. Left small comments

esp_data/datasets/xeno_canto_annotated_jeantet_23.py
benjaminsshoffman: fixed in next commit

tests/test_xeno_canto_annotated_jeantet_23.py
benjaminsshoffman: added these tests

esp_data/datasets/xeno_canto_annotated_jeantet_23.py
```diff
@@ -0,0 +1,213 @@
...
+
+    info = DatasetInfo(
+        name="xeno_canto_annotated_jeantet_23",
+        owner="benjamin",
+        split_paths={
+            "all": "gs://esp-ml-datasets/xeno_canto_annotated_jeantet_2023/all_gbif.csv",
+        },
+        version="0.1.0",
+        description="[MISSING]",
+        sources="Zenodo",
```
GaganNarula: Source should be XenoCanto if all files are XC files

GaganNarula (approved):
Small comment on the dataset info otherwise looks good. We should / will replicate your audio hash test elsewhere !

esp_data/datasets/xeno_canto_annotated_jeantet_23.py
benjaminsshoffman: fixed

tests/test_xeno_canto_annotated_jeantet_23.py
```diff
@@ -0,0 +1,149 @@
...
+
+# h = hashlib.sha256(audio0.tobytes()).hexdigest()
+# print("sha256:", h)
+
+
+EXPECTED_LEN_ALL = 967  #
+EXPECTED_FIRST_ITEM_AUDIO_SHA256 = (
+    "65fc1372fa64983d4998cf1be43d4469c7770e2f3485c860c654688ea3a3c30b"
+)
+# ---------------------------------------------------------------------------
```
GaganNarula, line 50: just for the sake of doing things the same way as for the other PRs, lets do the annotation hash test too

GaganNarula (changes requested):
lets add that annotation hash test

tests/test_xeno_canto_annotated_jeantet_23.py
benjaminsshoffman: done

GaganNarula (approved):

PR #152, opened 2025-10-29
Add python matrix version for CI subset

mil-ad:
esp-data is a library and we don't control what python version our users are going to use (besides forcing 3.10+). It's useful to test core esp-data functionality with multiple python versions. This PR adds the pipeline for doing that.

For now I've only added tests for `io` module into this step but obviously we'll add more as we go.

@GaganNarula  Adding this also exposed a race condition in our tests. We have to be careful with paths in `@pytest.mark.parametrize` because if they're fixed then running multiple tests at once (which this PR does) can cause issues. I have fixed it in the io tests but we use a similar pattern in other places.

### DIFF
--- a/tests/io/test_paths.py
+++ b/tests/io/test_paths.py
@@ -4,6 +4,7 @@ This module tests the integration between path objects (from paths.py) and
 filesystem operations (from filesystem.py).
 """
 
+import uuid
 from pathlib import Path
 
 import pytest
@@ -23,8 +24,8 @@ def test_anypath_local_path_with_file_operations():
 @pytest.mark.parametrize(
     "cloud_path",
     [
-        "gs://esp-ci-cd-tests/esp-data-tests/file1.txt",
-        "r2://esp-ci-cd-tests/esp-data-tests/file1.txt",
+        f"gs://esp-ci-cd-tests/esp-data-tests/test-{uuid.uuid4()}.txt",
+        f"r2://esp-ci-cd-tests/esp-data-tests/test-{uuid.uuid4()}.txt",
     ],
 )
 def test_cloud_filesystem_operations(cloud_path):
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -45,3 +45,38 @@ jobs:
 
       - name: Run tests
         run: uv run pytest tests
+
+  test-multi-version:
+    name: Run tests on multiple Python versions
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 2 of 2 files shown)


tests/io/test_paths.py
```diff
@@ -23,8 +24,8 @@ def test_anypath_local_path_with_file_operations():
 @pytest.mark.parametrize(
     "cloud_path",
     [
-        "gs://esp-ci-cd-tests/esp-data-tests/file1.txt",
-        "r2://esp-ci-cd-tests/esp-data-tests/file1.txt",
+        f"gs://esp-ci-cd-tests/esp-data-tests/test-{uuid.uuid4()}.txt",
+        f"r2://esp-ci-cd-tests/esp-data-tests/test-{uuid.uuid4()}.txt",
```
GaganNarula, line 28: So is the race condition caused because two different unit tests run on the same files at the same time ?

GaganNarula (commented):
I wonder if this pathlib stuff is too much to support long term. We only need a few methods for cloud paths and the rest we have to use fs for anyway. especially since it makes the CI slow (does it?) Maybe we should refactor this io CI action into another workflow yaml and trigger it only only on push to main ?

tests/io/test_paths.py
mil-ad: > I wonder if this pathlib stuff is too much to support long term. We only need a few methods for cloud paths and the rest we have to use fs for anyway.

I kind of agree. What we have right now is exactly to do what you mention it's still a lot of work to keep supporting it. If we run into one more big issue we should maybe drop it? But this conversation is outside the scope of this PR. I think we should have a python version matrix in CI anyway. CI speed is secondary priority IMO.

> especially since it makes the CI slow (does it?) 

It's not. Slowness is coming from all the bucket interactions.

> Maybe we should refactor this io CI action into another workflow yaml and trigger it only only on push to main ?

This is super quick. Doesn't take any time

> so is the race condition caused because two different unit tests run on the same files at the same time ?

Yeah if you run the same test in parallel they'd try to put/delete the same path at the same time

GaganNarula: ` think we should have a python version matrix in CI anyway. CI speed is secondary priority IMO.` i agree it seems important. I think there are lots of unit tests for the datasets that will run faster when we have the `streaming=True` backend PR integrated

mil-ad: yeah we should separate all these tests that actually touch gcs/r2/etc. 

We might even want to run them less frequently (they cost $$) etc.

GaganNarula (approved):
lgtm

PR #153, opened 2025-10-30
add anuraset strong

benjaminsshoffman:
strongly labeled anuraset for detection eval


esp_data/datasets/anuraset.py
```diff
@@ -0,0 +1,214 @@
...
+import librosa
+import numpy as np
+import pandas as pd
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+
+@register_dataset
+class AnuraSetStrong(Dataset):
```
GaganNarula, line 17: Do we plan to add the "weak" dataset too? Would be good to have that as another class in this module

esp_data/datasets/anuraset.py
```diff
@@ -0,0 +1,214 @@
...
+            output_take_and_give=cfg["output_take_and_give"],
+            data_root=cfg["data_root"],
+            sample_rate=cfg["sample_rate"],
+        )
+        if dataset_config.transformations:
+            meta = ds.apply_transformations(dataset_config.transformations)
+            return ds, meta
+        return ds, {}
+
+    def get_available_labels(self, anno_column: str = "Species") -> List[str]:
```
GaganNarula, line 193: Just a comment for future datasets like this which would have this method .. if the split is large (more than 100,000 rows for e.g) we should probably precompute and save the labels either in code or in a file in the same location as the split csv.

esp_data/datasets/anuraset.py
```diff
@@ -0,0 +1,214 @@
...
+    info = DatasetInfo(
+        name="anuraset_strong",
+        owner="benjamin",
+        split_paths={
+            "all": "gs://esp-ml-datasets/anuraset/anuraset_all_gbif.csv",
+        },
+        version="0.1.0",
+        description="[MISSING]",
+        sources="Zenodo",
+        license="Public Domain",
```
GaganNarula: Is this the actual license ? Not sure what public domain would mean..

esp_data/datasets/anuraset.py
benjaminsshoffman: I don't plan to right now, the preprocessing is different, nontrivial, and I haven't done it at all. And, we don't need it at the moment. I think it would make sense to wait until there is a need.

esp_data/datasets/anuraset.py
benjaminsshoffman: Changed

esp_data/datasets/anuraset.py
benjaminsshoffman: Changed: The paper says "The dataset and the raw data are provided under the Public Domain Dedication license (CC0) and are deposited in Zenodo". But actually the license on zenodo is slightly different so I used that (CC-BY-1)

GaganNarula (approved):
lgtm

PR #155, opened 2025-11-01
add birdeep dataset

benjaminsshoffman:
Yet another detection dataset

Note I made a small change to `io` to remove case-sensitivity in file extensions (some people name their files with `.WAV` instead of `.wav` and we should support that without renaming).


esp_data/datasets/birdeep.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        if self._data is None:
+            return []
+        available_labels = set()
+        for _, row in self._data.iterrows():
+            st = pd.read_csv(StringIO(row["selection_table"]), sep="\t")
+            available_labels.update(st[anno_column].astype(str).tolist())
+        if self.unknown_label in available_labels:
+            available_labels.remove(self.unknown_label)
+        return sorted(available_labels)
```
GaganNarula, line 213: can we do the same as anuraset and just add the labels here ? or too many ?

GaganNarula (commented):
just one comment

esp_data/datasets/birdeep.py
benjaminsshoffman: I prefer to keep the label generation programmatic, since it avoids errors and effort introduced by hard-coding what's already in the annotations. I understood the motivation for anuraset, since there were thousands of files, but here there are <300 in a split. So, this takes very little time to run.

GaganNarula (approved):
Lgtm

PR #157, opened 2025-11-03
fix hash for subsegmentation due to processing bug

benjaminsshoffman:
There was a bug in the subsegmentation data processing which we had to fix. THis commit fixes the annotation hash in the test_subsegmentation.py test

### DIFF
--- a/tests/test_subsegmentation.py
+++ b/tests/test_subsegmentation.py
@@ -19,7 +19,7 @@ from esp_data.datasets import Subsegmentation
 
 
 
-# --- Dataset snapshot ---
+# # --- Dataset snapshot ---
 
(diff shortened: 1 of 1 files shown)


mil-ad:
Thanks Benj. We've started to think about dataset versioning. We don't have a process yet but I guess we have be super careful about changing the underlying data in buckets once they've landed in esp-data. Luckily subsegmentation data isn't in a release yet.

mil-ad (approved):

benjaminsshoffman:
Yup, makes sense. For this, Greg found a c/p error where I'd set all the end times = start times, so all vox had duration 0. Adding this as a test for future dataset ingestion.

PR #158, opened 2025-11-04
Validate docs generation before deploy

mil-ad:
- Add a separate workflow that runs docs generation (without pushing) so that we can be sure that we can actually build and deploy the docs
- Fix the docstring issues the were happening here: https://github.com/earthspecies/esp-data/actions/runs/19043480409

### DIFF
--- a/.github/workflows/docs-validate.yml
+++ b/.github/workflows/docs-validate.yml
@@ -1,0 +1,30 @@
+name: Validate docs
+
+on:
+  pull_request:
+
+jobs:
+  validate-docs:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: docs-validate-${{ github.ref }}
+      cancel-in-progress: true
+
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-validate.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: "Set up Python"
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: ".python-version"
+
+      - name: Install the project
+        run: uv sync --all-extras --dev
+
+      - name: Build docs (dry run)
+        run: uv run mkdocs build --strict
```
GaganNarula, line 30: What does the --strict flag do?

mil-ad: it makes it abort on any warnings. But actually it's not really necessary because we have `strict: true` in `mkdocs.yml`. But it's the main point of this workflow so I prefer to keep it here

GaganNarula (approved):

PR #160, opened 2025-11-04
Added hooks to insert datasetinfo into docs

mil-ad:
- Also noticed two warnings when running mkdocs locally so fixed those too:
  ```
  Passing extra options directly under `options` is deprecated. Instead, pass them under `options.extra`, and update your templates.
  ```

### DIFF
--- a/docs/hooks/dataset_info_hook.py
+++ b/docs/hooks/dataset_info_hook.py
@@ -1,0 +1,165 @@
+"""
+MkDocs hook to inject dataset info into documentation.
+
+This hook automatically adds DatasetInfo metadata to each dataset's
+documentation by intercepting the HTML generation process.
+"""
+
+import logging
+import re
+from typing import Any
+
+from mkdocs.config.defaults import MkDocsConfig
+from mkdocs.structure.files import Files
+from mkdocs.structure.pages import Page
+
+from esp_data.dataset import DatasetInfo
+
+log = logging.getLogger("mkdocs.plugins.dataset_info_hook")
+
+
+def format_dataset_info_html(info: "DatasetInfo", class_name: str) -> str:
+    """Format a dataset's DatasetInfo as HTML.
+
+    Parameters
+    ----------
+    info : DatasetInfo
+        The dataset info object
+    class_name : str
+        The name of the dataset class
+
+    Returns
+    -------
+    str
+        Formatted HTML string
+    """
+    # Format sources
+    if isinstance(info.sources, list):
+        sources_str = ", ".join(info.sources)
+    else:
+        sources_str = info.sources
+
+    # Format split paths - limit display if too many
+    splits_list = list(info.split_paths.keys())
+    if len(splits_list) > 10:
+        splits_display = ", ".join(f"<code>{s}</code>" for s in splits_list[:10])
+        splits_display += f", ... ({len(splits_list)} total)"
+    else:
+        splits_display = ", ".join(f"<code>{s}</code>" for s in splits_list)
+
+    # Build HTML
+    html = []
+    html.append('<details class="dataset-info">')
+    html.append("<summary><strong>📊 Dataset Information</strong></summary>")
+    html.append('<div class="dataset-info-content">')
+    html.append("<table>")
+    html.append("<tbody>")
+    html.append(f"<tr><td><strong>Name</strong></td><td><code>{info.name}</code></td></tr>")
+    html.append(f"<tr><td><strong>Version</strong></td><td><code>{info.version}</code></td></tr>")
+    html.append(f"<tr><td><strong>Owner</strong></td><td>{info.owner}</td></tr>")
+    html.append(f"<tr><td><strong>License</strong></td><td>{info.license}</td></tr>")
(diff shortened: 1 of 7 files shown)


docs/hooks/dataset_info_hook.py
```diff
@@ -0,0 +1,165 @@
...
+    html.append(f"<tr><td><strong>Name</strong></td><td><code>{info.name}</code></td></tr>")
+    html.append(f"<tr><td><strong>Version</strong></td><td><code>{info.version}</code></td></tr>")
+    html.append(f"<tr><td><strong>Owner</strong></td><td>{info.owner}</td></tr>")
+    html.append(f"<tr><td><strong>License</strong></td><td>{info.license}</td></tr>")
+    html.append(f"<tr><td><strong>Sources</strong></td><td>{sources_str}</td></tr>")
+    html.append(f"<tr><td><strong>Available Splits</strong></td><td>{splits_display}</td></tr>")
+    html.append("</tbody>")
+    html.append("</table>")
+
+    if info.description and info.description not in ["[MISSING]", ""]:
```
GaganNarula, line 63: referencing an issue i created #165 ... we have a lot more info in the doc string than in the `info.description` field

docs/hooks/dataset_info_hook.py
```diff
@@ -0,0 +1,165 @@
...
+from mkdocs.config.defaults import MkDocsConfig
+from mkdocs.structure.files import Files
+from mkdocs.structure.pages import Page
+
+from esp_data.dataset import DatasetInfo
+
+log = logging.getLogger("mkdocs.plugins.dataset_info_hook")
+
+
+def format_dataset_info_html(info: "DatasetInfo", class_name: str) -> str:
```
GaganNarula: i think `class_name` isnt used in the function

docs/hooks/dataset_info_hook.py
```diff
@@ -0,0 +1,165 @@
...
+        Modified HTML content
+    """
+    # Only process the datasets page
+    if page.file.src_path != "datasets.md":
+        return html
+
+    try:
+        # Import here to avoid issues during mkdocs build
+        import esp_data.datasets  # noqa: F401 - ensure all datasets are registered
+        from esp_data.dataset import _dataset_registry
```
GaganNarula, line 110: Paul also has a function called `get_dataset_by_name` in his PR or something similar .. we do have `list_datasets` to get all the names, but we dont have a `fetch_dataset` or `get_dataset` to retrieve from the registry

docs/hooks/dataset_info_hook.py
```diff
@@ -0,0 +1,165 @@
...
+
+        log.info(f"Injected dataset info for {len(dataset_info_map)} datasets")
+        return modified_html
+
+    except Exception as e:
+        log.error(f"Failed to inject dataset info: {e}", exc_info=True)
+        return html
+
+
+def on_config(config: dict[str, Any]) -> dict[str, Any]:
```
GaganNarula: is this needed ? seems to just print something

mkdocs.yml
```diff
@@ -60,6 +60,9 @@ plugins:
             show_bases: false
             show_source: true
 
+hooks:
+  - docs/hooks/dataset_info_hook.py
```
GaganNarula, line 64: just for me to learn, are all hooks called at the same point before the final html is ready ?

docs/transforms.md
```diff
@@ -154,4 +153,4 @@ The `Deduplicate` transform removes duplicate rows from your dataset based on sp
     handler: python
     options:
         show_root_heading: true
-        show_source: true
\ No newline at end of file
+        show_source: true
```
GaganNarula, line 156: whats the diff here ? confused but its fine

GaganNarula (changes requested):
a few comments to improve!

docs/transforms.md
mil-ad: just whitespace 

Image

mkdocs.yml
mil-ad: Good question! I don't know how multiple hooks would work but my assumption is that they would run sequentially. But they can listen to the same event I guess

docs/hooks/dataset_info_hook.py
mil-ad: fixed

docs/hooks/dataset_info_hook.py
mil-ad: this is an actual event handler (https://www.mkdocs.org/dev-guide/plugins/#events) but I don't think is needed. `on_page_content` is probably enough

mil-ad: removed

docs/hooks/dataset_info_hook.py
mil-ad: sure, this PR just adds the mechanism.

docs/hooks/dataset_info_hook.py
mil-ad: let's revisit later?

GaganNarula (approved):
lgtm

PR #161, opened 2025-11-04
add nbm dataset

benjaminsshoffman:
detection dataset

### DIFF
--- a/esp_data/datasets/nocturnal_bird_migration.py
+++ b/esp_data/datasets/nocturnal_bird_migration.py
@@ -1,0 +1,234 @@
+"""NBM dataset"""
+
+from __future__ import annotations
+
+from io import StringIO
+from typing import Any, Dict, Iterator, List
+
+import librosa
+import numpy as np
+import pandas as pd
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+
+@register_dataset
+class NocturnalBirdMigration(Dataset):
+    """NocturnalBirdMigration Dataset
+
+    Description
+    -----------
+    Dataset of nocturnal vocalizations from migratory birds in Europe.
+    Vocalizations are annotated with start- and end- times, as well as high- and
+    low-frequencies.
+
+    The dataset consists of a train split and a test split. The test split consists
+    entirely of xeno-canto recordings which the dataset authors annotated. The train
+    split consists of recordings submitted by French citizen-scientists, as well as
+    xeno-canto recordings annotated by the dataset authors.
+
+    Description from the paper:
+
+    The persisting threats on migratory bird populations
+    highlight the urgent need for effective monitoring tech-
+    niques that could assist in their conservation. Among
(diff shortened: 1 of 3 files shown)


esp_data/datasets/nocturnal_bird_migration.py
```diff
@@ -0,0 +1,234 @@
...
+        split_paths={
+            "train": "gs://esp-ml-datasets/nocturnal_bird_migration/train.csv",
+            "train_nonxc": "gs://esp-ml-datasets/nocturnal_bird_migration/train_nonxc.csv",
+            "train_xc": "gs://esp-ml-datasets/nocturnal_bird_migration/train_xc.csv",
+            "test": "gs://esp-ml-datasets/nocturnal_bird_migration/test.csv",
+        },
+        version="0.1.0",
+        description="[MISSING]",
+        sources="Zenodo, xeno-canto",
+        license="CC BY-ND 3.0",
```
GaganNarula, line 80: since this is a "ND" non-derivate license dataset, are these splits and data the original splits and data ?

benjaminsshoffman: They're all the same (splits, data, annotations) as the original except I had to modify the labels -- there were a ton of typos ("backgrooud") and the annotations didn't match what they reported in the paper. The license allows us to use but not distribute the modifications. I think possible work-arounds are 1) to distribute the preprocessing code, 2) to ask for permission to redistribute with modifications (really, i was just cleaning up their sloppiness), 3) reserve this dataset for validation and not distribute as an eval set.

benjaminsshoffman: I lean toward 3 at the moment since that fits within the boundaries of what we're planning. Could simply add a comment in the dataset docstring, regarding the status?

GaganNarula: yeah lets go with a comment to make the change clear. @mil-ad when we open-source, we'll have to use the license info to avoid releasing datasets that have "ND" and are changed (and private ones)

mil-ad: Do we have another ND dataset in esp_data? I can't find any

GaganNarula: its only this one for now

mil-ad: I think option 3 is the only safe one for now

mil-ad: but we should probably add a bunch of guardrails in esp_data to shout about ND datasets

mil-ad: > the license allows us to use but not distribute the modifications

We're planning to open-source esp-data and this won't be able to go in after the tweaks. We need a mechanism to mark these datasets

GaganNarula (approved):

esp_data/datasets/nocturnal_bird_migration.py
benjaminsshoffman: We could also reach out -- the authors might simply agree to include it if they know what we did. For now, going with option 3 is fine

PR #162, opened 2025-11-05
add xeno canto

david-rx:
Adding the Xeno-canto dataset.


esp_data/datasets/xeno_canto.py
```diff
@@ -0,0 +1,342 @@
...
+    The dataset contains audio recordings with rich taxonomic information,
+    including species scientific and common names, family, genus, order,
+    and other metadata such as location, date, and recordist information.
+
+    Available Metadata Fields
+    -------------------------
+    **Taxonomic Information:**
+        - ``scientificName``: Scientific species name
+        - ``scientific_name_unified``: Unified scientific name
+        - ``species``: Common/vernacular species name
```
GaganNarula: wouldnt it be better to call it `species_common` and to match with `AnimalSpeak` ?

esp_data/datasets/xeno_canto.py
```diff
@@ -0,0 +1,342 @@
...
+    >>> from esp_data.datasets import XenoCanto
+    >>> dataset = XenoCanto(
+    ...     split="train",
+    ...     output_take_and_give={"scientificName": "species"}
+    ... )
+    >>> print(dataset.info.name)
+    xeno-canto
+    >>> print(dataset.available_sample_rates)
+    []
+    >>> # Load with pre-resampled 32kHz audio (when available)
```
GaganNarula: We have to remove lines like these `>>> # some comment` and just replace with `# some comment`. The >>> messes with mkdocs apparently.

esp_data/datasets/xeno_canto.py
```diff
@@ -0,0 +1,342 @@
...
+        name="xeno-canto",
+        owner="david; gagan",
+        split_paths={
+            "train": "gs://esp-ml-datasets/xeno-canto/v0.1.0/raw/metadata.csv",
+        },
+        version="0.1.0",
+        description="Xeno-canto audio dataset with taxonomic metadata. "
+        "Available at original (variable) sample rates and 32kHz (pre-resampled). "
+        "Pre-resampled audio uses librosa's kaiser_best resampling method.",
+        sources=["Xeno-canto"],
```
GaganNarula, line 100: Lets also add which xeno-canto dump this is, for e.g. Oct 2025 to the info.description. We'll have to find a way to make this a part of versioning

esp_data/datasets/xeno_canto.py
```diff
@@ -0,0 +1,342 @@
...
+        license="CC BY",
```
GaganNarula: Are all the files the same license ?

GaganNarula (commented):
some comments!

esp_data/datasets/xeno_canto.py
david-rx: Added!

esp_data/datasets/xeno_canto.py
david-rx: No, thanks

esp_data/datasets/xeno_canto.py
```diff
@@ -0,0 +1,345 @@
...
+        split_paths={
+            "train": "gs://esp-ml-datasets/xeno-canto/v0.1.0/raw/metadata.csv",
+        },
+        version="0.1.0",
+        description="Xeno-canto audio dataset with taxonomic metadata. "
+        "Available at original (variable) sample rates and 32kHz (pre-resampled). "
+        "Pre-resampled audio uses librosa's kaiser_best resampling method. "
+        "Xeno-canto dump as of Oct 2025.",
+        sources=["Xeno-canto"],
+        license="multiple",
```
GaganNarula: hey i know this is some work, but can we have a quick check of the licenses for all the samples and see how many unique values there are ? if there are only a few, we can make a comma separated string here. If there are a lot, but all have something in common like "CC BY XX" we can use that term. "multiple" doesnt make it clear if all of the samples actually can be used for training

david-rx: Good point - added

GaganNarula (approved):
lgtm!

PR #163, opened 2025-11-05
Fix comments in inaturalist doctest

mil-ad:
No point in putting comments after `>>`. That would break our document generation logic

image


GaganNarula (approved):
good to know! i think i've made this mistake a few times as well

mil-ad:
> good to know! i think i've made this mistake a few times as well

I don't think it's a mistake. It's just that mkdocs is making assumptions so I guess we have to go with it

PR #164, opened 2025-11-05
arctic bird sounds dataset

benjaminsshoffman:
the final detection eval dataset!


esp_data/datasets/arctic_bird_sounds.py
```diff
@@ -0,0 +1,251 @@
+"""NBM dataset"""
```
GaganNarula: Module name is incorrect

GaganNarula (approved):
just a tiny comment

PR #171, opened 2025-11-12
Fixed Filter -> FilterConfig typo

alf223:
I fixed a typo in the docs where Filter was imported instead of FilterConfig.


GaganNarula (approved):
Thanks! let us know if you find anything else in docs

PR #172, opened 2025-11-12
Fix #145 for all datasets

### LINKED ISSUE #145: AnimalSpeak - train split - OOM killed
When loading the AnimalSpeak train split, the process get OOM killed -> it needs to much memory.

`dataset = AnimalSpeak(split="train")`

GaganNarula:
Use pandas directly to load csv files instead of `read_text` because the latter causes OOM kill for large csvs. Somehow pandas handles this better..

+  update doc strings to fix minor formatting issues.


Paulchen-git (approved):
LGTM

mil-ad (commented):
but this won't work with R2 buckets would it?

GaganNarula:
> but this won't work with R2 buckets would it?

You mean reading directly with `pd.read_csv` ? no i guess not.. we'd have to supply the `storage_options` parameter with keys..

What do you suggest ?

mil-ad:
I think for now this is okay but let's raise an issue about losing R2 support. I don't want to provide `storage_options` to every place we use pd.read*. But we don't need R2 support quite yet so we can delay this

PR #173, opened 2025-11-14
Benchmark docs

Paulchen-git:
- Add docs for benchmark
- Fix jobs to be run on SLURM -> One job to rule them all 💍🔥
- update plot script

### DIFF
--- a/jobs/multi_dataset_latency.sh
+++ b/jobs/multi_dataset_latency.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #SBATCH --partition=cpu
-#SBATCH --job-name=multi_loading_time_benchmark
+#SBATCH --job-name=multi_latency_benchmark
 #SBATCH --output=/home/%u/outputs/loading_time/%A.log
 #SBATCH --mail-type=FAIL
 
@@ -24,9 +24,7 @@ echo "=== Starting job at $(date) ==="
 
 for DATASET in "${DATASETS[@]}"; do
     echo "Running benchmark for dataset: $DATASET"
-    sbatch --dependency=singleton jobs/benchmark_loading_time_default.sh --dataset "$DATASET"
+    sbatch --dependency=singleton jobs/benchmark_latency_default.sh --dataset "$DATASET"
 done
-
-uv run python scripts/benchmarks/plot_loading_bench.py
 
 echo "=== Finished job at $(date) ==="
--- a/scripts/benchmarks/README.md
+++ b/scripts/benchmarks/README.md
@@ -62,7 +62,12 @@ Job scripts are located in `jobs/` and are designed to be submitted with `sbatch
 
 Examples:
 
-- With default configuration.
+- Simple way to launch a global default benchmark (might take some time)
+```
+sbatch jobs/run_benchmark.sh
+```
+
+- Specific benchmarks with default configuration.
 ```
 # dataloader latency over different parameters for a single dataset
 sbatch jobs/benchmark_latency_default.sh --dataset beans
@@ -74,7 +79,7 @@ sbatch jobs/benchmark_loading_time_default.sh --dataset beans
 sbatch jobs/multi_dataset_loading_time.sh
 ```
 
-- With a provided config file.
+- Specific benchmark with a provided config file.
 ```
 # dataloader latency over different parameters for a single dataset
 sbatch jobs/benchmark_latency_config.sh --config your/path/to/config --data-location nfs
--- a/scripts/benchmarks/README.md
+++ b/scripts/benchmarks/README.md
@@ -1,0 +1,104 @@
+# Benchmarks
+
+This README documents how to use the benchmarking utilities in scripts/benchmarks/ and the SLURM job wrappers in jobs/.
+
+Note: most benchmark default functionality is currently bucket‑oriented. The engineering team is migrating to a bucket‑first workflow (GCS), and several utilities, default paths, and upload/plot pipelines assume bucket/GCS semantics by default. However `yaml` config files can be used to work with different configurations and data location such as NFS (see **Configurations** section).
+
+Contents
+--------
+- `benchmark_config.yaml` — config file listing the configuration of the experiment settings for `nfs` and `bucket` locations.
+- `benchmark_dataset.py`, `benchmark_latency.py`, `loading_time.py` — main benchmark scripts to measure loading and latency characteristics.
+- `benchmark_utils.py` — shared utilities used by the benchmarks.
+- `dataset_list.txt`, `generate_list.py` — helpers for producing dataset lists used in jobs.
+- `plot_loading_bench.py`, `plot_latency_bench.py` — plotting helpers that create figures from CSV results.
+- `fig/` — default local directory where plots are written.
+
+**Note: create the `fig/` folder to ensure plots are saved.**
+
+Quick overview
+--------------
+- Benchmarks are run on a SLURM cluster.
+- Results are saved as CSV files in a GCS bucket.
+- Plots are saved locally to `scripts/benchmarks/fig/`.
+
+You can specify a dataset configuration via a YAML file or use the default configuration.
+
+Cluster-first workflow (SLURM)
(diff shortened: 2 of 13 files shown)


scripts/benchmarks/plot_latency_bench.py
```diff
@@ -40,10 +40,17 @@ def main(number_of_samples: int, parameter: str) -> None:
...
 
+    # replace Nan with -1 for plotting
+    df.replace({float("nan"): -1}, inplace=True)
+    print(df[parameter])
+    labels = [x if x != -1 else "None" for x in df.sort_values(parameter)[parameter]]
     fig, axs = plt.subplots(1, 2, figsize=(10, 6))
 
     axs[0].set_title("Nominal Speed")
     axs[0].set_xlabel(parameter)
+    # Pour -1 le tick label doit être "None"
```
GaganNarula: I like the comment but maybe English is easier to follow :D

Paulchen-git: Sure !😂

scripts/benchmarks/README.md
```diff
@@ -0,0 +1,104 @@
...
+```
+info = DatasetInfo(
+        name="beans",
+        ...
+)
+```
+
+Collecting results and plots
+---------------------------
+- Plots are written to `scripts/benchmarks/fig/` inside the job’s workspace; results are saved as CSV files to a GCS bucket.
```
mil-ad, line 103: can we include the bucket path here?

mil-ad (approved):
tiny comment but otherwise this is good

scripts/benchmarks/README.md
Paulchen-git: Sure !

scripts/benchmarks/README.md
```diff
@@ -0,0 +1,106 @@
...
+Contents
+--------
+- `benchmark_config.yaml` — config file listing the configuration of the experiment settings for `nfs` and `bucket` locations.
+- `benchmark_dataset.py`, `benchmark_latency.py`, `loading_time.py` — main benchmark scripts to measure loading and latency characteristics.
+- `benchmark_utils.py` — shared utilities used by the benchmarks.
+- `dataset_list.txt`, `generate_list.py` — helpers for producing dataset lists used in jobs.
+- `plot_loading_bench.py`, `plot_latency_bench.py` — plotting helpers that create figures from CSV results.
+- `fig/` — default local directory where plots are written.
+
+**Note: create the `fig/` folder to ensure plots are saved.**
```
mil-ad, line 16: can't we create this in the job/script?

scripts/benchmarks/README.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+- With a provided config file.
+```
+# dataloader latency over different parameters for a single dataset
+sbatch jobs/benchmark_latency_config.sh --config your/path/to/config --data-location nfs
+
+# dataset loading time and sample access for a single dataset
+sbatch jobs/benchmark_loading_time_config.sh --config your/path/to/config --data-location bucket
+```
```
mil-ad, line 89: this is great. 

I think one thing that would be nice is to have a top-level script where one can literally do something like `run_everything.sh` with no args etc and that's kind of our official benchmark

scripts/benchmarks/README.md
Paulchen-git: Yes good idea

jobs/multi_dataset_latency.sh
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+# Log DATASETS for debugging
+echo "Datasets to benchmark: ${DATASETS[*]}"
+
+echo "=== Starting job at $(date) ==="
+
+for DATASET in "${DATASETS[@]}"; do
+    echo "Running benchmark for dataset: $DATASET"
+    sbatch --dependency=singleton jobs/benchmark_latency_default.sh --dataset "$DATASET"
```
GaganNarula, line 27: just for my knowledge .. what is `singleton` here ?

scripts/benchmarks/benchmark_utils.py
```diff
@@ -99,6 +99,8 @@ def build_raw_dataset(config_path: Path, data_location: str, dataset_name: str)
         )
         dataset, _ = dataset_from_config(config_path, key=data_location)
         raw = raw[data_location]["dataset"]
+        raw["split_name"] = raw["split"]
+        raw["split"] = "config"
```
GaganNarula, line 103: i'm not sure i understand these variables "split" and "split_name" or what their purpose is.. could you add a comment ?

scripts/benchmarks/benchmark_utils.py
```diff
@@ -109,8 +111,10 @@ def build_raw_dataset(config_path: Path, data_location: str, dataset_name: str)
         # data_root:
         raw = {}
         raw["dataset_name"] = dataset_name
-        raw["split"] = dataset.split
-        raw["sample_rate"] = dataset.sample_rate
+        raw["split"] = "default"
+        raw["split_name"] = dataset.split
+        # sample_rate column in benchmark results should be default if there is no resampling
+        raw["sample_rate"] = "default"
```
GaganNarula, line 117: this is a comment for us to discuss at the next meeting.. i think resampling may actually be the biggest contributor to iteration time .. in the next version, we might need to add `sample-rate` as a parameter to the the `benchmark_latency` script, and explicitly set it to a value or `None` (for no resampling).

scripts/benchmarks/plot_loading_bench.py
```diff
@@ -71,14 +74,17 @@ def plot_results(results: pd.DataFrame) -> None:
...
             f"-- Bucket: {bucket_location} -- Machine: {machine_location}"
+            f"\n Sample Rate: {sr}, Split: {split}"
         )
         plt.tight_layout()
-        plt.savefig(f"scripts/benchmarks/fig/boxplot_{name}_{location}_{col_name}.png")
-        logger.info(f"Saved plot: scripts/benchmarks/fig/boxplot_{name}_{location}_{col_name}.png")
+        # Ensure the fig directory exists
+        os.makedirs("scripts/benchmarks/fig", exist_ok=True)
+
+        plt.savefig(f"scripts/benchmarks/fig/boxplot_{name}_{location}.png")
```
GaganNarula, line 86: Another thing to discuss: may be we should be commiting these figures to the repo when we make a release ? if there are not too many it would be a good have them in an `assets` folder or something .. we could then automatically include them in docs.

GaganNarula (approved):
lgtm ! (just a few comments to discuss or keep in mind)

jobs/multi_dataset_latency.sh
Paulchen-git: I use this dependency to make sure each job wait for the previous job to be done. singleton ensures serial execution of jobs with the same name and same user

scripts/benchmarks/benchmark_utils.py
Paulchen-git: I just wanted to store the information that says this is not the default split -> the split was given in a config file.
I think I can definitely improve this, I use this to make it easy to create plots.

Paulchen-git: 'split' could be renamed with a boolean "from_config" for example

scripts/benchmarks/benchmark_utils.py
Paulchen-git: Yes it could be a parameter for benchmark latency.
I also think that by default there shouldn't be a specified sample rate (for most datasets it's already None) but for some dataset a default sample rate is given, then there is a resampling in the default configuration.

scripts/benchmarks/plot_loading_bench.py
Paulchen-git: I have some plots for the loading time benchmark with first sample access etc that could be added to the repo. But for the latency, the results depends a lot on the experiment (especially on batch_size and sleep) so they are less interesting as general information and they can only be use as example of what we can get from an experiment.

Issue #122, opened 2025-08-29
Resample with torchaudio rather than librosa for better performance


benjaminsshoffman:
Datasets currently use librosa for resampling, but torchaudio is much much faster.

See https://docs.pytorch.org/audio/stable/tutorials/audio_resampling_tutorial.html for performance comparison and for settings which approximately reproduce librosa resamping

PR #176, opened 2025-11-18
Resamplers

GaganNarula:
Adding torchaudio resampler. #122
It is definitely an order of magnitude faster..


mil-ad:
>  It is definitely an order of magnitude faster..

On GPU or CPU?

Also why are we doing this now? Is it a priority?

nkundiushuti:
sorry for intervening...since we noted difference in the results given the resampling method, this should be done cautiously, benchmarking with torch and the usual librosa method we used in the past to see if we get similar results

mil-ad:
yeah this changes the default to torchaudio so we have to careful

GaganNarula:
yup you guys are right! thats why its only a draft PR. However, we should discuss this soon because for most of the datasets in esp-data on-the-fly resampling is required because a) there aren't already existing sample rate versions in the buckets (like iNaturalist) and b) resampling with librosa is about 300x slower (not a full benchmark, just taking one file 90 sec file and resampling it 100 times).

GaganNarula:
> > It is definitely an order of magnitude faster..
> 
> On GPU or CPU?
> 
> Also why are we doing this now? Is it a priority?

cpu test

PR #178, opened 2025-11-23
Add SuperbStarling dataset

sara-keen:
- Add superb_starling_dataset.py: Dataset class for 2,179 superb starling flight calls
- Includes individual ID, group ID, sex for each bird
- Data stored in gs://superb-starlings-keen/organized_data/
- Tests to check all audio and metadata are correct
- All tests pass locally

### DIFF
--- a/esp_data/datasets/superb_starling_dataset.py
+++ b/esp_data/datasets/superb_starling_dataset.py
@@ -1,0 +1,249 @@
+"""Superb Starling dataset"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, Iterator, List
+
+import librosa
+import numpy as np
+import pandas as pd
+import os
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+
+@register_dataset
+class SuperbStarling(Dataset):
+    """Superb Starling Dataset
+
+    Description
+    -----------
+    Dataset of superb starling (Lamprotornis superbus) flight calls with precise time bounds, 
+    individual ID, and social group ID 
+
+    Each entry includes:
+    - An audio clip containing one flight call
+    - Annotations for exact start/stop of the call in audio clip 
+    - Metadata (bird ID, group, sex, ring, timestamp)
+
+    The metadata file is a tab-separated text file that is formatted as a Raven selection table. 
+    This lets you open all sound files in Raven and see the annotations aligned for every selection, which each correspond to a single flight call
+
+    
+
+    References
+    ----------
+    Keen, S. C., Meliza, C. D., & Rubenstein, D. R. (2013). Flight calls signal group and individual identity but not kinship in a cooperatively breeding bird. Behavioral Ecology, 24(6), 1279-1285.
+
+    """
+
+    info = DatasetInfo(
+        name="superb_starling",
+        owner="Sara",  
+       # split_paths={
+       #     "all": "gs://superb-starlings-keen/organized_data/superb_starlings_flightcalls.txt",   # FIX
+       # },
+        split_paths={
+            "all": os.path.expanduser("/home/sara_earthspecies_org/individual_id/data/superb_starlings_flightcalls.txt"),
+            },
+        version="0.1.0",
+        description="superb starling flight calls with individual ID and group ID annotations",
+        sources="Kenya field recordings",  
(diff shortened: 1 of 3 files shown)


esp_data/datasets/superb_starling_dataset.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+
+    info = DatasetInfo(
+        name="superb_starling",
+        owner="Sara",  
+       # split_paths={
+       #     "all": "gs://superb-starlings-keen/organized_data/superb_starlings_flightcalls.txt",   # FIX
+       # },
+        split_paths={
+            "all": os.path.expanduser("/home/sara_earthspecies_org/individual_id/data/superb_starlings_flightcalls.txt"),
```
GaganNarula: split paths have to be bucket paths :) . I would also recommend placing this dataset within the `esp-ml-datasets` bucket (for e.g. `gs://esp-ml-datasets/superb-startlings-keen-2022/v0.1.0/...`) . The reason is that a) this makes it faster to load the data (because this bucket is in the same zone as the cluster, and it has some other new features) and b) it helps us maintain and monitor the datasets. We need to start having stricter control over the datasets because we have no way of knowing if the someone changed the underlying data. We want to be able to version and track dataset changes.. @mil-ad wdyt ?

esp_data/datasets/superb_starling_dataset.py
```diff
@@ -0,0 +1,249 @@
...
+                res_type="kaiser_best",
+            )
+            sr = target_sr
+
+        # Add audio and sample rate to output
+        row["audio"] = audio
+        row["sample_rate"] = sr
+
+        # Calculate duration from audio
+        row["duration_s"] = len(audio) / float(sr)
```
GaganNarula: is it possible to rename this to `duration_secs` ? just for more clarity

esp_data/datasets/superb_starling_dataset.py
```diff
@@ -0,0 +1,249 @@
...
+            return pd.DataFrame()
+        
+        stats = (
+            self._data.groupby(["bird", "group", "sex", "ring"])
+            .size()
+            .reset_index(name="num_vocalizations")
+        )
+        return stats.sort_values("num_vocalizations", ascending=False)
+
+    def get_group_stats(self) -> pd.DataFrame:
```
GaganNarula: `get_individual_stats` and `get_group_stats` are very specific helper functions which would more suitable in a script / notebook that uses this dataset imo.. is it necessary to keep them as methods in the dataset ?

tests/test_superb_starling.py
```diff
@@ -0,0 +1,299 @@
...
+# quit()
+# # #
+
+# TODO: Run the code above once you have your dataset working to generate these values
+EXPECTED_LEN_ALL = 2179  # Update this after running snapshot generation
+EXPECTED_FIRST_ITEM_AUDIO_SHA256 = (
+    "REPLACE_WITH_ACTUAL_HASH"  # Update this after running snapshot generation
+)
+ANNOTATIONS_SHA256 = (
+    "REPLACE_WITH_ACTUAL_HASH"  # Update this after running snapshot generation
```
GaganNarula: i guess you will update these soon.. which snapshot are you referring to ?

GaganNarula (changes requested):
Great start! a few changes / comments

esp_data/datasets/superb_starling_dataset.py
sara-keen: not necessary, I'll remove

esp_data/datasets/superb_starling_dataset.py
sara-keen: thanks, I had forgotten to update this!

esp_data/datasets/superb_starling_dataset.py
sara-keen: done

esp_data/datasets/superb_starling_dataset.py
GaganNarula: these functions are still here.. :D

GaganNarula:
Sorry for being nitpicky, but can we rename the file to `superb_starling.py` and remove the `dataset` part. Just so that it conforms to the way the other files are named.

GaganNarula (changes requested):
a few more changes

GaganNarula (approved):
Lgtm!

PR #181, opened 2025-12-05
Benchmark minor improvements

Paulchen-git:
- rename some functions and variables to make it clearer
- fix --plot and --save flag for loading_time.py

### DIFF
--- a/scripts/benchmarks/plot_loading_bench.py
+++ b/scripts/benchmarks/plot_loading_bench.py
@@ -8,8 +8,8 @@ from benchmark_utils import set_logging_config
 set_logging_config()
 
 
-def get_results() -> pd.DataFrame:
-    """Retrieve benchmark loading time results from a CSV file in a GCS bucket.
+def get_saved_results_from_cloud() -> pd.DataFrame:
+    """Retrieve all benchmark loading time results from a CSV file saved in a GCS bucket.
     Returns
     -------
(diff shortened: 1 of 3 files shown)


scripts/benchmarks/plot_loading_bench.py
```diff
@@ -20,16 +20,19 @@ def get_results() -> pd.DataFrame:
...
     ----------
     results : pandas.DataFrame
         DataFrame containing benchmark results for a specific dataset.
     """
-    logger = logging.getLogger("results_plotter")
+    logger = logging.getLogger("last_results_plotter")
 
-    df = get_results()
+    df = get_saved_results_from_cloud()
     name = results["dataset_name"].iloc[0]
```
GaganNarula, line 36: is the first index in the dataframe always the `last_experiment` ? it seems from the name of the function that the 0th is the last one but i'm not sure. If it is, just mention that in the doc string

GaganNarula (commented):
one comment!

scripts/benchmarks/plot_loading_bench.py
Paulchen-git: the results dataframe contains the results you just got in your experiment.
the df dataframe contains all stored results.
So the 0th is the first row of the latest results.

GaganNarula (approved):

Issue #183, opened 2025-12-05
sr is None -> issue raised when computing duration


Paulchen-git:
https://github.com/earthspecies/esp-data/blob/44b2fe01a1c42a03db073331eb368061688e1484/esp_data/datasets/anuraset.py#L175

sr = target_sr should be indented.

https://github.com/earthspecies/esp-data/blob/44b2fe01a1c42a03db073331eb368061688e1484/esp_data/datasets/anuraset.py#L181

Issue #182, opened 2025-12-05
default split "all" isn't an available splits


Paulchen-git:
https://github.com/earthspecies/esp-data/blob/44b2fe01a1c42a03db073331eb368061688e1484/esp_data/datasets/nocturnal_bird_migration.py#L86

PR #184, opened 2025-12-05
fix bugs in anuraset and nocturnal birds datasets #182 #183

Paulchen-git:
- fix anuraset _process function #183 
- set Nocturnal birds default split to an existing split #182


GaganNarula (approved):
nice!

PR #185, opened 2025-12-05
Docs for backends

GaganNarula:


docs/backends.md
Paulchen-git, line 1: Lgtm

Paulchen-git (approved):
Lgtm !

mil-ad (commented):
This is great documentation. My only comment is that we shouldn't advertise it in the docs. I don't see it as something our users have to know about or change. Am I wrong?

I'd move the link from the top level and put it somewhere in API Reference section

GaganNarula:
> This is great documentation. My only comment is that we shouldn't advertise it in the docs. I don't see it as something our users have to know about or change. Am I wrong?
> 
> I'd move the link from the top level and put it somewhere in API Reference section

Well mostly I agree but I see situations where users have to access the actual data to do some custom thing, for which we have a `.unwrap` property. I thought with this documentation we get to explain what the backends are, how the interface works and what they need to do for custom operations that esp-data cant do yet or is out of scope..

For e.g. Benj is redo-ing WABAD now in a new PR where he changed the default backend back to `pandas` from `polars`. I think its because he is comfortable with pandas and wouldn't like to learn another library for doing custom tasks.. 

Anyhow, I can move it to API Reference if you like although I prefer it in the main section so people understand what we are aiming at

mil-ad:
That's fair enough. Let's keep it at high-level for now. We could do a big revamp later. I want to move that Concat too :D

mil-ad (approved):

PR #186, opened 2025-12-08
update wabad dataset

benjaminsshoffman:
WABAD was the first dataset I implemented, and as a result it was the lowest quality implementation. I wanted to updated it since WABAD is probably the most important dataset we have for detection evals.

Here, I've updated the dataset so that it is more self-contained (i.e. no weird label mapping logic), and also had the full suite of tests we want to run (e.g. hashing for the underlying data).

### DIFF
--- a/esp_data/datasets/wabad.py
+++ b/esp_data/datasets/wabad.py
@@ -13,6 +13,8 @@ from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
 from esp_data.backends import BackendType
 from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
 
+SPECIES_INFO_PATH = "gs://esp-ml-datasets/wabad/v0.1.0/raw/gbif_labels.csv"
+
 
 @register_dataset
 class WABAD(Dataset):
@@ -69,8 +71,7 @@ class WABAD(Dataset):
         name="wabad",
         owner="benjamin",
         split_paths={"all": "gs://esp-ml-datasets/wabad/v0.1.0/raw/all_info_gbif.csv"},
-        species_info_path="gs://esp-ml-datasets/wabad/v0.1.0/raw/gbif_labels.csv",
-        version="0.2.0",
+        version="0.1.0",
         description="[MISSING]",
         sources="zenodo.org",
         license="CC-BY-4.0",
@@ -103,12 +104,12 @@ class WABAD(Dataset):
         """
         super().__init__(output_take_and_give, backend=backend, streaming=streaming)
         self.split = split
-        self._data: pd.DataFrame | None = None
+        self._data = None
         self.annotation_columns = ["Species"]
         self.unknown_label = "Unknown"
         self.sample_rate = sample_rate
 
-        self.available_labels = pd.read_csv(self.info.species_info_path)["Species"].to_list()
+        self.available_labels = pd.read_csv(SPECIES_INFO_PATH)["Species"].to_list()
 
         # Load split CSV
         self._load()
@@ -178,7 +179,7 @@ class WABAD(Dataset):
                 scale=True,
                 res_type="kaiser_best",
             )
-        sr = target_sr
+            sr = target_sr
 
         # Selection table
         st = pd.read_csv(StringIO(row["selection_table"]), sep="\t")
@@ -219,7 +220,7 @@ class WABAD(Dataset):
 
         Yields
         -------
-        Dict[str, Any]
+        dict[str, Any]
             Each sample in the dataset.
         """
         for row in self._data:
@@ -247,6 +248,8 @@ class WABAD(Dataset):
             output_take_and_give=cfg["output_take_and_give"],
             data_root=cfg["data_root"],
             sample_rate=cfg["sample_rate"],
+            backend=cfg["backend"],
+            streaming=cfg["streaming"],
         )
 
         if dataset_config.transformations:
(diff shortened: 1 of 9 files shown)


esp_data/datasets/wabad.py
```diff
@@ -71,8 +68,9 @@ class WABAD(Dataset):
     info = DatasetInfo(
         name="wabad",
         owner="benjamin",
-        split_paths={"all": "gs://esp-ml-datasets/wabad/v0.1.0/raw/all_info.csv"},
-        version="0.1.0",
+        split_paths={"all": "gs://esp-ml-datasets/wabad/v0.1.0/raw/all_info_gbif.csv"},
+        species_info_path="gs://esp-ml-datasets/wabad/v0.1.0/raw/gbif_labels.csv",
```
GaganNarula: we dont have a `species_info_path`  so I wouldn't add it here, just keep the constant on top.

esp_data/datasets/wabad.py
```diff
@@ -71,8 +68,9 @@ class WABAD(Dataset):
     info = DatasetInfo(
         name="wabad",
         owner="benjamin",
-        split_paths={"all": "gs://esp-ml-datasets/wabad/v0.1.0/raw/all_info.csv"},
-        version="0.1.0",
+        split_paths={"all": "gs://esp-ml-datasets/wabad/v0.1.0/raw/all_info_gbif.csv"},
```
GaganNarula, line 73: has the old data been overwritten ?

esp_data/datasets/wabad.py
```diff
@@ -71,8 +68,9 @@ class WABAD(Dataset):
...
+        species_info_path="gs://esp-ml-datasets/wabad/v0.1.0/raw/gbif_labels.csv",
+        version="0.2.0",
```
GaganNarula: if its really version 0.2.0 I would have kept the older data in the v0.1.0/ path and made a new path with this version number in it. But if the data is over-written, then just keep the version at v0.1.0.

esp_data/datasets/wabad.py
```diff
@@ -105,25 +103,20 @@ def __init__(
...
         self.split = split
-        self._data = None
-        self.default_anno_column = "Species"  # This is the annotation column used for label mapping
-        self.unknown_annos: List[str] = []
-        # in the following, default anno column must appear last for label mapping to work properly:
-        self.annotation_columns = ["Genus", "Family", "Order", "Common"] + [
-            self.default_anno_column
-        ]
-
+        self._data: pd.DataFrame | None = None
```
GaganNarula: we're moving to the `Backends` framework so its no longer a Dataframe

esp_data/datasets/wabad.py
```diff
@@ -205,17 +181,12 @@ def _process(self, row: dict[str, Any]) -> dict[str, Any]:
         sr = target_sr
```
GaganNarula: this would set sr to None if target_sr is None.. i think it should be indented

esp_data/datasets/wabad.py
```diff
@@ -249,45 +219,54 @@ def __iter__(self) -> Iterator[dict[str, Any]]:
 
         Yields
         -------
-        dict[str, Any]
+        Dict[str, Any]
```
GaganNarula: why upper case ?

esp_data/datasets/wabad.py
```diff
@@ -249,45 +219,54 @@ def __iter__(self) -> Iterator[dict[str, Any]]:
...
+            and the metadata will be returned as dict, otherwise an empty dict.
+        """
         cfg = dataset_config.model_dump(exclude={"dataset_name", "transformations"})
         ds = cls(
             split=cfg["split"],
             output_take_and_give=cfg["output_take_and_give"],
             data_root=cfg["data_root"],
             sample_rate=cfg["sample_rate"],
-            backend=cfg["backend"],
-            streaming=cfg["streaming"],
```
GaganNarula, line 267: please keep those fields because we may want to change the backend using a config file

GaganNarula (changes requested):
looking good but some changes needed

esp_data/datasets/wabad.py
benjaminsshoffman: fixed

esp_data/datasets/wabad.py
benjaminsshoffman: no (the old data is also the starting point of the reformatting code, so it should not be deleted).

However, the old data is no longer compatible with this dataset class.

esp_data/datasets/wabad.py
benjaminsshoffman: The older data is not compatible with this new dataset class, so I have changed it to v0.1.0 now. We'll consider the first iteration of WABAD as a practice run.

esp_data/datasets/wabad.py
benjaminsshoffman: Removed the type annotation altogether

esp_data/datasets/wabad.py
benjaminsshoffman: Whoa good catch. I fixed this, and also fixed it on like 8 other datasets where the error was copied.

esp_data/datasets/wabad.py
benjaminsshoffman: changed. probably some ai thing.

esp_data/datasets/wabad.py
benjaminsshoffman: Fixed. Not sure why they got deleted. Fixed in one other dataset, too

benjaminsshoffman:
I think I fixed all the changes requested

GaganNarula (approved):
🦅

Issue #179, opened 2025-12-03
ConcatenatedDataset mistakenly returns strings of TSVs instead of selection tables


alf223:
esp-data question: I'm using ConcatenatedDataset to merge two datasets that each have the inline-TSV format for storing selection tables within a row.
When I concatenate the two datasets using soft merge, and both have a selection_table column in _data, I'm getting surprising behavior. The datasets' __getitem__ parses the TSV string from _data into a DataFrame before returning it. But when I do ConcatenatedDataset.__getitem__ , it returns the actual TSV string, and not the parsed DataFrame. My understanding of why this is is that after calling the source dataset's __getitem__, there's this line:

`source_item.update(row)
`
that overwrites the properly parsed DataFrame with the raw string from the merged _data, so downstream code receives a string instead of the DataFrame it expects.
I see the comment in the code that says this is so that "transforms applied at the level of the concatenated dataset take precedence"  but in my case there's no transform, it's just the raw string overwriting the processed value from __getitem__.
Is this intended behavior? I think I could make a workaround where all of these strongly labeled datasets with TSVs have the column name in _data to be different from what __getitem__ returns, but that's also a little strange. Or is this something that should be changed in esp-data?

alf223:
here is a little diagram from Claude that helped me to understand what's going on:

Simple/Scalar Case (what they designed for)
DatasetA._data:
┌─────────────┬───────┐
│ audio_path  │ label │
├─────────────┼───────┤
│ "a1.wav"    │ "dog" │
│ "a2.wav"    │ "cat" │
└─────────────┴───────┘

DatasetB._data:
┌─────────────┬───────┐
│ audio_path  │ label │
├─────────────┼───────┤
│ "b1.wav"    │ "bird"│
└─────────────┴───────┘

ConcatenatedDataset._data (after soft merge):
┌─────────────┬───────┬─────────────────┬───────────────┐
│ audio_path  │ label │ _source_dataset │ _source_index │
├─────────────┼───────┼─────────────────┼───────────────┤
│ "a1.wav"    │ "dog" │ 0               │ 0             │
│ "a2.wav"    │ "cat" │ 0               │ 1             │
│ "b1.wav"    │ "bird"│ 1               │ 0             │
└─────────────┴───────┴─────────────────┴───────────────┘

When you call combined[0]:
1. row = {"audio_path": "a1.wav", "label": "dog", ...}
[...]

PR #187, opened 2025-12-09
fix for issue 179: a simple "chain" of datasets

GaganNarula:
#179 Deals with failure case of dataset concatenation. Basically the problem is:

1. User creates a concatenated dataset ```ds = ConcatenatedDataset([ds1, ds2], merge_level="soft")```
2. When iterating, the current version of ConcatenatedDataset assumes that a key in the source dataset is not transformed into something else when the source dataset's `__getitem__` is called. The problem occurs when updating the returned `item` dict from a source dataset with the same item from the concatenated data backend. If any key is _colliding_ , the item from the concatenated backend overwrites the source.
3. This fails in the case of datasets like `HawaiianBirds` and `NocturnalBirdMigration` etc, where a field like `selection_table` is a string in the data backend (i.e. in the csv file), but that string is converted to a pandas dataframe in the `__getitem__` call. So, in the ConcatenatedDataset, the backend has concatenated dataframe with the column `selection_table` being a string, while the returned item from the source dataset has `type(item["selection_table"]) = pd.DataFrame` .. the user actually wants the dataframe not the string :(

The reason we were overwriting source with concatenated (point 1) was because the whole point of dataset concatenation was to allow transforms like subsampling etc to be applied at a combined dataset level.

The proposed solution (which is not very nice and up for discussion) in this PR is to add `collision_policy` which allows the user to select the behavior .

EDIT: `collision_policy` is removed in favor of a second concat type, called `ChainedDataset` which, as the name suggests, just chains datasets together while maintaining the Dataset api, which requires `__getitem__`. This requirement can be dropped and made simpler in another PR


esp_data/io/read_utils.py
```diff
@@ -44,6 +45,23 @@ def read_text(
         return f.read()
 
 
+def read_yaml(path: str | AnyPathT) -> dict:
```
mil-ad: can we add a few tests for this?

mil-ad: could be in a separate PR

esp_data/__init__.py
```diff
@@ -39,29 +49,40 @@
...
+    "Subsegmentation",
+    "WABAD",
     "XenoCanto",
+    "XenoCantoAnnotatedJeantet23",
+    "HawaiianBirds",
+    "AnuraSetStrong",
+    "Powdermill",
+    "ArcticBirdSounds",
+    "NocturnalBirdMigration",
+    "Birdeep",
```
mil-ad, line 88: random idea for future: maybe we could write our own little pre-commit hook for this so we don't forget adding them. We could automatically find dataset files in datasets dir and try to import them. It's hacky but might be useful

tests/test_concat.py
mil-ad, line 1: can we put the new ones in `test_chaining.py` or something? Our tests are getting long

esp_data/concat.py
```diff
@@ -532,3 +546,166 @@ def __str__(self) -> str:
             f"Columns: {', '.join(self.columns)}\n"
             f"Source datasets: {len(self._source_datasets)}"
         )
+
+
+class ChainedDataset(Dataset):
```
mil-ad: I'd move this to `chain.py` but up to you

mil-ad (approved):
LGTM. Left a few comments but don't want to block it. If they make sense do them in a separate PR please

esp_data/__init__.py
GaganNarula: good idea! one problem though: adding more and more datasets to init, I _think_ slows down import speed of any one dataset

mil-ad: good point, there are ways to do lazy importing. We could look into it

PR #188, opened 2025-12-11
Add job array experiment to benchmark

Paulchen-git:
Job array experiment to explore the impact of parallel jobs on latency.

Most changes are new jobs scripts for different array sizes, it may not be the best way to do.

-> The job array experiment is quite independent from previous benchmark experiments.

### DIFF
--- a/scripts/benchmarks/job_array_exp_config.yaml
+++ b/scripts/benchmarks/job_array_exp_config.yaml
@@ -1,0 +1,15 @@
+# Benchmark configuration for job array experiments
+
+nfs:
+  dataset:
+    dataset_name: beans
+    split: cbi_train
+    sample_rate: 44100
+    data_root: /home/milad_earthspecies_org/data-migration/marius-highmem/mnt/foundation-model-data/
+
+bucket:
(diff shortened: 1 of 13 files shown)


scripts/benchmarks/job_array_exp_config.yaml
```diff
@@ -0,0 +1,15 @@
+# Benchmark configuration for job array experiments
+
+nfs:
+  dataset:
+    dataset_name: beans
+    split: cbi_train
+    sample_rate: 44100
```
GaganNarula, line 7: did we want this as a fixed value ? i thought we wanted to test with no resampling. Anyway i think it might be better to have one `job_array_exp_config_no_resample` and one `job_array_exp_config_with_resample`

GaganNarula (approved):
lgtm!

scripts/benchmarks/job_array_exp_config.yaml
Paulchen-git: Yes agreed, I'll fix it in the next PR

PR #189, opened 2025-12-12
Data discovery 177

### LINKED ISSUE #177: Data discovery with esp-data
There seems to be a need to discover which of the esp-data datasets contains what "kind" of data. Currently, the only descriptive metadata we hold for each dataset is in the class doc string or the `DatasetInfo` attribute. These descriptions are plain text, and could be more structured.

I propose extending `DatasetInfo` with `tags` or additional json structured fields that can be queried to answer questions like:

1. Which datasets contain X species ?
2. Which datasets contain X,Y,Z family / genus / order ?
3. Which datasets have selection tables or "strong" labels, i.e. start and end times
4. Which datasets have call types annotated ?
5. What task does the dataset solve for ? E.g. species_classification, individual_id, number_of_vocalizing_individuals

inesnolas:
This change is linked to this [issue](https://github.com/earthspecies/esp-data/issues/177)
goal is to generate an updated database of the existent datasets in esp-data, where searching contents and attributes is easy and quick without having to open each dataset and explore in depth. 

At this stage, I've defined new attributes in DatasetInfo and there is a script for generating the database. 
so when we run /scripts/data_discovery.py the table is generated in ./dataset_attributes_table_test.csv"

Questions: 
1) not sure about location and timing to run the script. possibly we want to automatically generate the table whenever there are changes to datasets
2) location and how to use this table? 
3) the field: taxonomic_info is intended to be semi auto-filled by looking up a database, I've already included the download of that, but the logic behind it is still flagged as a #TODO. I need to test it alongside a adding a new dataset perhaps?


esp_data/dataset.py
```diff
@@ -1,17 +1,25 @@
...
 from esp_data.transforms.registry import RegisteredTransformConfigs
 
+try:
+    # Attempt to load the NCBITaxa object once
+    TAXON_DB = ete3.NCBITaxa()
+except Exception:
+    # TODO if taxonomic database cannot be loaded, set to None and skip validation :/?
+    TAXON_DB = None
+    print("Warning: Could not initialize ete3.NCBITaxa. All taxonomic validation will be skipped.")
+
```
GaganNarula, line 22: hmm this is very specific library code and this module is supposed to be more abstract.. needs a refactor, lets discuss

esp_data/dataset.py
```diff
@@ -153,6 +159,57 @@ def delay_importing_reg(cls, v: Any) -> Any:  # noqa: ANN401
...
+        default_factory=lambda: "", description="list of taxonomic classes included in the dataset"
+    )
+
+    phylum: list[str] | str = Field(
+        default_factory=lambda: "", description="list of taxonomic phyla included in the dataset"
+    )
+
+    # --- VALIDATOR ---
+    # TODO
+    #  validate that the name provided exists in the NCBI taxonomy database for the given rank
```
GaganNarula, line 204: people in esp use `gbif` more than NCBI.. we even have a taxonomy app that is built on the gbif api

esp_data/dataset.py
```diff
@@ -236,6 +293,59 @@ class DatasetInfo(BaseModel):
...
+            "temporal annotations",
+            "single label for whole audio clip",
+            "multiple labels for whole audio clip",
+        ]
+    ] = Field(default_factory=list, description="Type of labels provided in the dataset.")
+
+    label_content: List[
+        Literal[
+            "species",
+            "individual identification",
```
GaganNarula, line 312: for string Literals that are categories its preferable to not have spaces  (replace with underscore), and not to have two possible names (like `position/distance` is not a good idea). We would also need some place for a description of what each term means, for e.g. "health state" or "group belonging" .. are these terms canonical ?

scripts/data_discovery.py
```diff
@@ -0,0 +1,32 @@
...
+        attributes_list.append(attributes)
+    table = pd.DataFrame(attributes_list, index=registry)
+    table.to_csv("./dataset_attributes_table_test.csv")
+
+
+if __name__ == "__main__":
+    generate_dataset_table_with_attributes()
+
+    #  TODO
+    # make it a command line call and create a url to access the generated file and search on it.
```
GaganNarula, line 32: we can add this script to the CI

GaganNarula (commented):
some comments, lets discuss further :) I feel like this taxonomic info thing is bloating the DatasetInfo a bit much

esp_data/dataset.py
GaganNarula: Add description of each type to the `description` field in `Field`.

esp_data/dataset.py
```diff
@@ -236,6 +293,59 @@ class DatasetInfo(BaseModel):
...
+            "call type",
+            "song content",
+            "group belonging",
+            "behavioural observations",
+            "position/distance",
+            "age",
+            "health state",
+            "emotional state",
+            "distractor classes / background animals",
+            "recording conditions",  # weather, habitat, time of day or year, season
```
GaganNarula, line 324: The comment is very descriptive! so lets move descriptions of these categories to the `description` field

PR #191, opened 2025-12-14
New datasets

GaganNarula:
+ Dolphin whistles
+ Gibbons
+ Geladas monkeys
+ (WIP) Beans zero


esp_data/datasets/beans_zero.py
```diff
@@ -0,0 +1,361 @@
...
+        # if self.sample_rate is not None and sr != self.sample_rate:
+        #     audio = librosa.resample(
+        #         y=audio,
+        #         orig_sr=sr,
+        #         target_sr=self.sample_rate,
+        #         scale=True,
+        #         res_type="kaiser_best",
+        #     )
+
+        # row["audio"] = audio
```
nkundiushuti: maybe remove this

esp_data/datasets/beans_zero.py
```diff
@@ -0,0 +1,361 @@
...
+        use_presampled = False
+        if self.sample_rate is not None and self.sample_rate in self._sample_rate_paths:
+            path_column = self._sample_rate_paths[self.sample_rate]
+            # Check if the pre-resampled path column exists in the data
+            if path_column in row and row[path_column] is not None and row[path_column] != "":
+                # Use pre-resampled audio
+                audio_path = anypath(self.data_root) / row[path_column]
+                use_presampled = True
+
+        if use_presampled:
```
nkundiushuti, line 248: I like this, it should be the default way of dealing with pre-resampled versions

nkundiushuti (approved):

esp_data/datasets/beans_zero.py
GaganNarula: yes but i'm still not sure about storing paths for each sampling rate in the csv... we should just have a programmatic way of doing this because column names are not standardized across datasets

PR #192, opened 2025-12-15
add audioset strong

david-rx:
Adds the strongly-labeled subset of AudioSet. The dataset is separate from the main AudioSet to match the format of our other strongly-labeled datasets.


esp_data/datasets/audioset_strong.py
```diff
@@ -0,0 +1,316 @@
...
+    info = DatasetInfo(
+        name="audioset_strong",
+        owner="david; marius; masato",
+        split_paths={
+            "train": "gs://esp-ml-datasets/audioset/v0.2.0/raw/csv-data/audioset_train_strong_selection_tables_filtered_with_32khz.csv",
+        },
+        version="0.1.0",
+        description="AudioSet Strong: Strongly-labeled subset with temporal annotations",
+        sources=["YouTube"],
+        license="Mixed",
```
GaganNarula: I think Audioset is CC BY 4.0? https://research.google.com/audioset/download.html . Can you make this change to both audioset_strong and audioset ?

esp_data/datasets/audioset_strong.py
```diff
@@ -0,0 +1,316 @@
...
+    )
+
+    # Mapping of sample rates to their corresponding path columns
+    # Pre-resampled 32kHz audio is available for ~92% of entries
```
GaganNarula: isnt this going to be a problem if not all are available?

esp_data/datasets/audioset_strong.py
```diff
@@ -0,0 +1,316 @@
...
+        if self.output_take_and_give:
+            item: dict[str, Any] = {}
+            for old_key, new_key in self.output_take_and_give.items():
+                item[new_key] = row[old_key]
+            return item
+
+        return row
+
+    def __getitem__(self, idx: int) -> dict[str, Any]:
+        if self._data is None:
```
GaganNarula: I dont think we need these RuntimeError and NotImplementedError checks here or in `__iter__` because the `self._data` should never be None anyway and the backend throws the NotImplementedError exception

esp_data/datasets/audioset_strong.py
```diff
@@ -0,0 +1,316 @@
...
+    def get_available_labels(self) -> List[str]:
+        """
+        Return all possible labels found in the dataset.
+
+        Returns
+        -------
+        List[str]
+            A sorted list of all unique labels in the dataset.
+        """
+        if self._data is None:
```
GaganNarula: same here, ref earlier comment

tests/test_audioset_strong.py
```diff
@@ -0,0 +1,315 @@
...
+    We use pandas backend here since it's the same data.
+    """
+    # Use a limited subset for speed - check first few items manually
+    labels_sample = set()
+    for i in range(min(100, len(ds_pandas))):
+        try:
+            item = ds_pandas[i]
+            st = item["selection_table"]
+            if "Label" in st.columns:
+                labels_sample.update(st["Label"].astype(str).tolist())
```
GaganNarula: you can get a sample from `ds_pandas` by doing `ds_pandas.sample(n=100)`, and then get `labels_sample`

tests/test_audioset_strong.py
```diff
@@ -0,0 +1,315 @@
...
+
+def test_pre_resampled_audio_32khz():
+    """Test loading pre-resampled 32kHz audio with fallback for corrupt files.
+
+    This test verifies that:
+    1. When sample_rate=32000, the dataset attempts to use pre-resampled files
+    2. Corrupt pre-resampled files (very short) fall back to on-the-fly resampling
+    3. Audio is returned correctly in both cases
+    """
+    ds_32k = AudioSetStrong(split="train", sample_rate=32000)
```
GaganNarula: I made some changes to the tests in iNaturalist so that we dont load multiple datasets per test, because each load (when default `streaming=False`) consumes time and memory.. maybe you could follow that template here

GaganNarula (changes requested):
some changes :)

esp_data/datasets/audioset_strong.py
```diff
@@ -0,0 +1,302 @@
...
+        -------
+        List[str]
+            A sorted list of all unique labels in the dataset.
+        """
+        labels: set[str] = set()
+        for row in self._data:
+            selection_table_blob = row.get("selection_table", "")
+            if selection_table_blob is None or selection_table_blob == "":
+                continue
+            st = pd.read_csv(StringIO(selection_table_blob), sep="\t")
```
GaganNarula: Anthony says using pandas to convert the selection table from string to dataframe is slow, and is a bottleneck in training. Could you try with `polars` ? if its much faster he can make that change too

GaganNarula (commented):
one more comment, almost there

esp_data/datasets/audioset_strong.py
david-rx: A quick benchmarking by Claude claimed it was indeed 5x faster

GaganNarula (approved):
lets go!

PR #193, opened 2025-12-16
David audioset v2

david-rx:
Adds a new version of AudioSet.

There is a question of how to add new versions -- I've implemented one way here but happy to have feedback. This could also be an entirely separate dataset from the current AudioSet if that's preferrable.

### DIFF
--- a/esp_data/datasets/audioset.py
+++ b/esp_data/datasets/audioset.py
@@ -234,10 +234,7 @@ class AudioSet(Dataset):
             )
 
         location = self.info.split_paths[self.split]
-        # Read CSV directly (supports local and cloud paths) via backend to avoid
-        # materializing large text blobs in memory.
-        #
-        # Note: We parse JSON-encoded "labels" lazily per-row in _process().
+
         self._data = self._backend_class.from_csv(location, streaming=self._streaming)
 
     @classmethod
--- a/esp_data/datasets/audioset.py
+++ b/esp_data/datasets/audioset.py
@@ -36,7 +36,13 @@ class AudioSet(Dataset):
     The "caption" column contains the caption from AudioSetCaps when available.
     AudioSetCaps Paper: https://arxiv.org/abs/2411.18953
     AudioSetCaps Dataset: https://huggingface.co/datasets/baijs/AudioSetCaps
-    Note these are empty with the exception of the unbalanced_train split.
+    Note these are empty with the exception of the unbalanced_train split of the V1 dataset.
+
+    Note that AudioSet contains different files depending on YouTube video availability at
+    time of download. Version 0.1.0 contains a dump of AudioSet pulled in 2021 and resampled
+    to 16khz. Version 0.2.0 contains a larger set of audios pulled from this HuggingFace
+    release https://huggingface.co/datasets/agkphysics/AudioSet and maintaining the sample
+    rates of the original files.
 
     Pre-resampled Audio
     -------------------
--- a/esp_data/datasets/audioset.py
+++ b/esp_data/datasets/audioset.py
@@ -291,7 +291,7 @@ class AudioSet(Dataset):
 
         ds = cls(
             split=cfg["split"],
-            version=cfg["version"],
+            version=cfg.get("version"),
             output_take_and_give=cfg["output_take_and_give"],
             data_root=cfg["data_root"],
             sample_rate=cfg["sample_rate"],
--- a/esp_data/datasets/audioset.py
+++ b/esp_data/datasets/audioset.py
@@ -8,7 +8,46 @@ import numpy as np
 
 from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
 from esp_data.backends import BackendType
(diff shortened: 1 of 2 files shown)


esp_data/datasets/audioset.py
```diff
@@ -37,8 +36,26 @@ class AudioSet(Dataset):
...
+    Version 0.2.0 includes pre-resampled 32kHz audio that can be loaded directly
+    without on-the-fly resampling for faster data loading:
+
+    >>> # Load with pre-resampled 32kHz audio (v0.2.0, no resampling needed)
+    >>> dataset_32k = AudioSet(split="validation", version="0.2.0", sample_rate=32000)
+    >>> print(dataset_32k.available_sample_rates)
+    [32000]
+
+    >>> # Load with on-the-fly resampling to 16kHz
+    >>> dataset_16k = AudioSet(split="validation", version="0.2.0", sample_rate=16000)
```
GaganNarula, line 97: So this is interesting! You're introducing versions in this dataset which we haven't yet solved for all datasets. However, i'm not against this implementation(wdyt @mil-ad ?) . But I would add an additional `AudiosetConfig(DatasetConfig)` (see voxaboxen.py for e.g.) because you have a new dataset specific parameter called `version`. Then use this config in the `from_config` method. Otherwise, in its current state, we can not declare a version in a yaml file (dataset_from_config wont work).

esp_data/datasets/audioset.py
```diff
@@ -49,33 +66,65 @@ class AudioSet(Dataset):
...
+    VERSIONS = {
+        "0.1.0": {
+            "split_paths": {
+                "train": "gs://esp-ml-datasets/audioset/v0.1.0/raw/csv-data/unbalanced_train_segments_processed.csv",
+                "train-balanced": "gs://esp-ml-datasets/audioset/v0.1.0/raw/csv-data/balanced_train_segments_processed.csv",
+                "validation": "gs://esp-ml-datasets/audioset/v0.1.0/raw/csv-data/eval_segments_processed.csv",
+                "train-animal": "gs://esp-ml-datasets/audioset/v0.1.0/raw/csv-data/unbalanced_train_segments_processed_animal.csv",
+                "validation-animal": "gs://esp-ml-datasets/audioset/v0.1.0/raw/csv-data/eval_segments_processed_animal.csv",
+                "train-noise": "gs://esp-ml-datasets/audioset/v0.1.0/raw/csv-data/unbalanced_train_segments_processed_noise.csv",
+                "validation-noise": "gs://esp-ml-datasets/audioset/v0.1.0/raw/csv-data/eval_segments_processed_noise.csv",
```
GaganNarula: I've seen some issues with missing audio files in some of these splits ? can you iterate through them and check ? If some splits are broken we should remove them

tests/test_audioset.py
```diff
@@ -233,6 +233,41 @@ def test_str_representation(dataset: Dataset) -> None:
...
+    assert "train-balanced" not in ds_v2.info.split_paths
+
+
+def test_available_sample_rates_v020() -> None:
+    """Check pre-resampled sample-rate reporting for v0.2.0.
+
+    We condition on the metadata containing the expected path column to avoid
+    hard-failing if the underlying CSV schema changes.
+    """
+    ds = AudioSet(split="validation", version="0.2.0", streaming=True)
```
GaganNarula, line 254: nice, use streaming=True for reducing memory and time in tests

esp_data/datasets/audioset.py
```diff
@@ -248,7 +362,20 @@ def __getitem__(self, idx: int) -> dict[str, Any]:
         -------
         dict[str, Any]
             A dictionary containing the audio data, text label, label, and path.
+
+        Raises
+        ------
+        IndexError
+            If the index is out of bounds.
         """
+        if self._streaming:
```
GaganNarula: we dont need this because `backend` will throw the same error

esp_data/datasets/audioset.py
```diff
@@ -248,7 +362,20 @@ def __getitem__(self, idx: int) -> dict[str, Any]:
...
+            raise NotImplementedError(
+                "Random access (__getitem__) is not available in streaming mode. "
+                "Iterate over the dataset instead."
+            )
+        if idx >= len(self._data):
+            raise IndexError(f"Index {idx} out of bounds for dataset of length {len(self._data)}.")
```
GaganNarula: same here

GaganNarula (changes requested):
looking good, but some changes needed. I'd wait for milad's input

esp_data/datasets/audioset.py
david-rx: Good catch

esp_data/datasets/audioset.py
david-rx: True - removed the broken old splits

mil-ad (commented):
Looks good to me. Letting Gagan decide

GaganNarula (approved):

PR #195, opened 2025-12-17
Audio-language datasets

david-rx:
A PR to support Audio-Language datasets in esp-data.

The revised scope supports adding prompts, using jinja2 templates. Prompts are multiturn as a series of messages. Both prompts and responses use jinja2 templates.

Transform adding prompts and responses for an audio-language dataset in a follow-up PR.


esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+            If None, uses a default "Describe this audio." prompt.
+        response_prefix : str, optional
+            Prefix to add before the response.
+        response_suffix : str, optional
+            Suffix to add after the response.
+        seed : int, optional
+            Random seed for deterministic variant selection.
+        """
+        if variants is None:
+            variants = [PromptVariant("Describe this audio.", "{text}")]
```
GaganNarula: I think its better not to make `variants` optional and have to choose an arbitrary default. Also, prompts maybe used for text only models so `Describe this audio.` wouldnt be very useful

esp_data/prompts/registry.py
```diff
@@ -0,0 +1,58 @@
...
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .base import PromptTemplate
+
+_PROMPT_REGISTRY: dict[str, "PromptTemplate"] = {}
+
+
+def register_prompt(template: "PromptTemplate") -> "PromptTemplate":
```
GaganNarula: i'm not sure we need prompt template registries.. i think the coding agents (claude / cursor) etc are picking up on our registry pattern and using it everywhere.. 

A fixed set of default PromptTemplates that are useful across bioacoustics can be added to the code base so that users can model their prompts after these, but i dont think we need to expose `register_prompt` to users.

GaganNarula: oh wait, just thought about this. Do you have in mind to declare which prompt template to use from a config file ? in that case I can see the use

esp_data/prompts/registry.py
```diff
@@ -0,0 +1,58 @@
+"""Prompt template registry."""
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .base import PromptTemplate
+
+_PROMPT_REGISTRY: dict[str, "PromptTemplate"] = {}
```
GaganNarula: why not just use `PromptTemplate` as a type ? (without "") since its runtime_checkable anyway

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+        -------
+        str
+            The formatted prompt string.
+
+        Raises
+        ------
+        KeyError
+            If the prompt variant contains placeholders not found in item.
+        """
+        self._selected_variant = self._rng.choice(self._variants)
```
GaganNarula: i think a class like BasePromptTemplate should not have several variants that are chosen from based on a random number generator. A base class should be simple and deterministic, while a StochasticPromptTemplate can have this advanced capability

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+        try:
+            return self._selected_variant.prompt.format(**item)
```
GaganNarula: i think we might want this template application to be a bit more advanced. There's the issue of escape characters in the item dict's , or being able to validate the type in a prompt template. We might want to look at t strings https://docs.python.org/3/library/string.templatelib.html or jinja2 templates

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+            The source item dictionary.
+
+        Returns
+        -------
+        dict[str, Any]
+            A copy of the item with 'prompt' and 'text' keys added.
+        """
+        result = item.copy()
+        result["prompt"] = self.format_prompt(item)
+        result["text"] = self.format_response(item)
```
GaganNarula: Is "text" a good field to have hard coded ? text can mean anything (prompt, response, metadata, output)

esp_data/audio_language.py
```diff
@@ -0,0 +1,298 @@
...
+
+from esp_data.dataset import Dataset, DatasetConfig, DatasetInfo
+from esp_data.prompts import PromptTemplate, get_prompt
+
+
+def is_audio_language_dataset(dataset: Dataset) -> bool:
+    """Check if a dataset natively provides audio-language format.
+
+    A dataset is considered audio-language native if its items contain
+    'audio', 'prompt', and 'text' fields.
```
GaganNarula: Not sure how this would be used .. if a prompt template is used to compute `result["prompt"]` and `result["text"]` in prompts.base.py, then why should an esp-dataset already have these fields ?

esp_data/prompts/base.py
david-rx: Thanks, will look into the prebuilt ones including jinja2

esp_data/prompts/registry.py
david-rx: Yes I was imagining declaring the prompts for each task in a config file (e.g. https://github.com/earthspecies/esp-research/blob/ecf0df63b30b6ba0cb2ba473e485ace8f1125cc2/projects/NatureLM-audio-v1.5/data/configs/xeno_canto_life_stage.yaml)

esp_data/prompts/base.py
david-rx: I'm not sure. For some reason "text" is commonly used as the label or response (e.g. in huggingface training scripts in NLP, by SALMONN, etc.) But I agree it's misleading, I'm fine to go with "response" I think

esp_data/audio_language.py
david-rx: There will be some datasets which natively have questions and answers built in, we'll eventually want to add these from e.g. AudioFlamingo

david-rx: It's also easy enough to specify a "pass through" prompt in those cases though, maybe it doesn't need special handling

esp_data/prompts/base.py
david-rx: What did you use in the NatureLM-audio training data release in the end? Maybe we can stick with it

esp_data/audio_language.py
```diff
@@ -0,0 +1,298 @@
...
+        """Create a dataset from configuration (not implemented).
+
+        For AudioLanguageDataset, use direct instantiation with a source dataset.
+
+        Raises
+        ------
+        NotImplementedError
+            Always raised. Use direct instantiation instead.
+        """
+        raise NotImplementedError(
```
GaganNarula: i think it would be nice to have this if we're going with a wrapper, see for e.g. `ChainedDataset`

GaganNarula (commented):
lets get back to this in Jan :)

esp_data/prompts/base.py
GaganNarula: i used `instruction` for input and `output` for what is expected, but even that is kind of meh. Lets discuss.. its not a huge deal but if its an important keyword then we should choose appropriately. I like `response` actually

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+
+        Raises
+        ------
+        KeyError
+            If the prompt variant contains placeholders not found in item.
+        """
+        self._selected_variant = self._rng.choice(self._variants)
+        try:
+            return self._selected_variant.prompt.format(**item)
+        except KeyError as e:
```
nkundiushuti: I would assert keys prior to formatting and have a more generic error here that can be triggered even by formatting errors.

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+        KeyError
+            If response template contains placeholders not found in item.
+        """
+        if self._selected_variant is None:
+            # format_prompt wasn't called first; select a variant now
+            self._selected_variant = self._rng.choice(self._variants)
+
+        try:
+            response = self._selected_variant.response.format(**item)
+        except KeyError as e:
```
nkundiushuti: same here

esp_data/prompts/base.py
nkundiushuti: thinking forward to multi-turn chat prompt styles, maybe we can have something like  List[{"role": "...", "content": "..."}]. 
```
prompt = "What species is this?"
text   = "American Robin"
```
would be:
```
messages = [
  {"role": "user", "content": "What species is this?"},
  {"role": "assistant", "content": "American Robin"}
]
```
as an alternative we could keep whatever you decide to use and then have"
```
{
  "audio": ...,
  "prompt": [
    "What species is this?",
    "Are you confident?"
  ],
  "text": [
    "American Robin",
    "Yes."
  ]
}
```
in this case in the next PR I would transform prompts and texts into lists, but would implicitly code roles, and thinking about naturelm2 would break some things (system messages, tool calls etc). 
maybe as a better name we can have them the roles mentioned: user_prompts instead of prompts and assistant_responses instead of response

esp_data/audio_language.py
```diff
@@ -0,0 +1,298 @@
...
+    >>> xc = XenoCanto(split="train")  # doctest: +SKIP
+    >>> ds1 = AudioLanguageDataset(xc, prompt="species_common")  # doctest: +SKIP
+    >>> ds2 = AudioLanguageDataset(AudioSet(split="train"), prompt="captioning")  # doctest: +SKIP
+    >>> combined = ConcatenatedDataset([ds1, ds2])  # doctest: +SKIP
+
+    Auto-detect native audio-language datasets:
+
+    >>> # If source already has prompt/text, passthrough is used automatically
+    >>> native_al_dataset = SomeNativeALDataset()  # doctest: +SKIP
+    >>> dataset = AudioLanguageDataset(native_al_dataset, prompt="passthrough")  # doctest: +SKIP
```
nkundiushuti: could we have passthrough to be the default behaviour without even declaring the template in the registry e.g. with prompt=None?

david-rx: Yep agree

esp_data/prompts/base.py
david-rx: Good point, we should think that through now. I like your first option of a list of messages - should we switch prompts to already be a list of messages now?

esp_data/audio_language.py
david-rx: Makes sense. Where can I find ChainedDataset?

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+"""
+
+import random
+from abc import abstractmethod
+from dataclasses import dataclass
+from typing import Any, Protocol, runtime_checkable
+
+
+@dataclass
+class PromptVariant:
```
mil-ad: I'm not sure the name `PromptVariant` matches what this is. Why not call it `PromptResponsePair`?

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+    ensuring consistency when multiple prompt variants are used.
+
+    Parameters
+    ----------
+    prompt : str
+        The prompt/instruction template. Can include {field_name} placeholders.
+    response : str
+        The response template. Can include {field_name} placeholders.
+    task : str, optional
+        Task descriptor (e.g., "species_id", "captioning"). Default: "default".
```
mil-ad: does `task` belong here?

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+    - format_prompt(): generates the instruction/question
+    - format_response(): generates the expected answer/completion
+
+    The __call__ method applies the template to an item dict,
+    adding 'prompt' and 'text' keys.
+    """
+
+    name: str
+
+    @abstractmethod
```
mil-ad: When you add `@abstractmethod` to a Protocol becomes a runtime-checkable abstract base class that requires explicit inheritance. Is that what we want? The main point of Protocols is structural typing without inheritance

esp_data/prompts/base.py
```diff
@@ -0,0 +1,312 @@
...
+            A dictionary containing dataset fields (e.g., species, caption).
+
+        Returns
+        -------
+        str
+            The formatted prompt/instruction string.
+        """
+        ...
+
+    @abstractmethod
```
mil-ad: same as above

esp_data/audio_language.py
```diff
@@ -0,0 +1,298 @@
...
+        True if the dataset provides native audio-language format.
+
+    Notes
+    -----
+    This function accesses dataset[0] to check fields, which may trigger
+    audio loading for that sample.
+    """
+    try:
+        sample = dataset[0]
+        return all(k in sample for k in ("audio", "prompt", "text"))
```
mil-ad: this triggers loading of the dataset. We should eventually make sure this operation is cheap but I think currently this is quite slow. We can update Dataset to include this in the metadata?

mil-ad (changes requested):
I think it makes sense to have two stages for preparing the data. esp-data should be model/template/tokenizer agnostic. We should probably then have another stage where we take the data and

- apply chat templates (model specific)
- add special tokens
- do actual tokenization

but I think in this first PR we're missing two main things:

1. we should have:

- conversations by default: even if it's just one turn
- roles: user, assistant, system, etc?

2. Use jinja2 for templating. This could be a string in python _and_ via config files

esp_data/prompts/base.py
GaganNarula: i am for messages instead of prompts!

esp_data/audio_language.py
```diff
@@ -0,0 +1,298 @@
...
+    except (IndexError, KeyError, RuntimeError):
+        return False
+
+
+class AudioLanguageDataset(Dataset):
```
mil-ad: I think this could just be a transform. Each row in the output dataset holds one conversation. 

I can totally imagine that we'll have a class for LLM-specific data in the future, especially if we want to generate data online but right now AudioLanguageDataset doesn't add much.

esp_data/prompts/base.py
david-rx: I tried implementing it this way, but it's a decent amount of duplicated code with a slightly more complex interface, and the benefit of having two classes was not clear to me. In the single-class version, you get deterministic behavior by passing only one prompt-response pair, if you pass multiple you choose between them. I don't mind changing if you have a strong opinion, but I think it's not so much of a base class anyway, I've renamed it without the "Base" as its not obvious we need multiple kinds now

esp_data/prompts/base.py
david-rx: I'm not sure. It's basically a piece of metadata that helps us keep track of what prompts go for what, we may handle them differently in augmentations and evals. I left it on for now but if we want to move it somewhere (e.g. a PromptTemplateWithMetadata wrapper or something) I'm fine with that

esp_data/prompts/base.py
david-rx: Good to know thanks!

esp_data/audio_language.py
david-rx: Good catch

esp_data/prompts/template.py
```diff
@@ -0,0 +1,95 @@
...
+    def render(self, **kwargs: Any) -> Message:
+        """Render message content with Jinja2 template substitution.
+
+        Uses StrictUndefined to raise clear errors for missing variables.
+
+        Returns
+        -------
+        Message
+            New Message with rendered content.
+        """
```
GaganNarula, line 50: maybe a class docstring test or method docstring test would be nice to explain to the user how it works

esp_data/prompts/template.py
```diff
@@ -0,0 +1,95 @@
...
+
+class PromptResponsePair(BaseModel):
+    """A prompt-response pair with explicit separation.
+
+    The messages list contains the prompt (input to the model).
+    The response field is a template string for the target output.
+    """
+
+    messages: list[Message]
+    response: str
```
GaganNarula: shouldnt `response` actually be like `Message(role="assistant", content="{{ something }}")` ?

esp_data/prompts/registry.py
```diff
@@ -0,0 +1,89 @@
...
+    ValueError
+        If a template with the same name is already registered.
+    """
+    if template.name in _REGISTRY:
+        raise ValueError(f"Prompt '{template.name}' already registered.")
+    _REGISTRY[template.name] = template
+    return template
+
+
+def register_prompt_from_config(
```
GaganNarula: i think we dont really need to `register` prompts or get them from a registry or list them.. essentially we just want to construct them from config files right ? If its just that then your config objects (the Message, PromptResponsePair and PromptResponseTemplateConfig) cover the functionality thats needed.

esp_data/prompts/template.py
```diff
@@ -0,0 +1,95 @@
...
+    With a single variant, behavior is deterministic.
+    With multiple variants, randomly selects one per call.
+    Use seed parameter for reproducible results.
+
+    Returns a dict with 'prompt' (JSON of messages) and 'response' (string).
+    """
+
+    def __init__(
+        self,
+        name: str,
```
GaganNarula, line 98: Do we really need `name` ?

esp_data/prompts/template.py
```diff
@@ -0,0 +1,95 @@
...
+        variants: list[PromptResponsePair],
+        seed: int | None = None,
+    ) -> None:
+        self.name = name
+        self.variants = variants
+        self._rng = random.Random(seed)
+
```
GaganNarula, line 105: we should have a `from_config(cfg: PromptResponseTemplateConfig | AnyPathT | dict) -> Self`  method that loads this template from yaml or dict or a `PromptResponseTemplateConfig` object. esp-data.io has a `read_yaml` utility

GaganNarula (changes requested):

esp_data/prompts/registry.py
david-rx: We want to write the prompts once and reuse them across datasets. So I was imagining registering them, so we can reference them by name in config files, rather than re-defining them in every dataset-specific config file. This is also why they have names (re. your below comment)

esp_data/prompts/template.py
david-rx: See thread above

esp_data/prompts/registry.py
david-rx: I guess we can have all dataset configs point to a secondary prompt config file, one for each prompt, so they are shared across datasets. Tbh I prefer the registry for this project, it's just easier to manage a large number of changing prompts which don't need one config file each, but I see why it could be bloat for esp-data. Maybe it is time to move to the project folder in esp-research like you suggested

esp_data/prompts/template.py
david-rx: Maybe, we're getting rid of response anyway

esp_data/prompts/registry.py
GaganNarula: we can add all prompts to one file and just have one function `get_prompt(name=..., version: Optional[str] =...., source_file: str = "default.yml")` that loads the prompt yaml and finds the right one ? we it would be the same as tracking it in a .py file but without the code overhead ?

esp_data/prompts/template.py
```diff
@@ -0,0 +1,158 @@
...
+class Message(BaseModel):
+    """A single message in a conversation."""
+
+    role: Literal["system", "user", "assistant"]
+    content: str
+    _compiled_template: Template | None = None
+
+    def model_post_init(self, __context: Any) -> None:  # noqa: ANN401
+        """Pre-compile the Jinja2 template after model initialization."""
+        object.__setattr__(
```
mil-ad: I don't think we should be using pydantic's private methods. 

Maybe you want 

```
model_config = ConfigDict(arbitrary_types_allowed=True)
```

or making it a proper `Field`?

mil-ad (commented):
This is in a good shape. Just left one comment

esp_data/prompts/registry.py
david-rx: Okay yes that works I am convinced, we only need get_prompt. I'll remove the registry and add this separately when we actually add the prompts

GaganNarula:
Closing this PR because we moved it to esp-research

PR #197, opened 2025-12-18
New plot + small fixes

Paulchen-git:
- plot of the evolution of nominal speed (later other parameters) over time.
- fix default resampling value for loading_time benchmark.
- add config with no_resampling for job array experiment

### DIFF
--- a/jobs/array_jobs/benchmark_latency_array_1.sh
+++ b/jobs/array_jobs/benchmark_latency_array_1.sh
@@ -17,7 +17,7 @@ srun uv run python scripts/benchmarks/benchmark_latency.py \
    --batch-size 64 \
    --prefetch-factor 0 \
    --sleep 0.5 \
-   --config-path scripts/benchmarks/job_array_exp_config.yaml \
+   --config-path scripts/benchmarks/job_array_exp_no_resampling_config.yaml \
    --data-location bucket \
    --save "local" \
    --nb_array 1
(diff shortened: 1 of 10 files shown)


mil-ad (approved):
Looks good to me!

jobs/array_jobs/benchmark_latency_array_1.sh
```diff
@@ -17,7 +17,7 @@ srun uv run python scripts/benchmarks/benchmark_latency.py \
    --batch-size 64 \
    --prefetch-factor 0 \
    --sleep 0.5 \
```
GaganNarula, line 19: is there are reason we decided to set sleep to 0.5 ?

GaganNarula (commented):
one comment

jobs/array_jobs/benchmark_latency_array_1.sh
Paulchen-git: I set sleep to 0.5 for 2 reasons :
- to get quick results
- to highlight prefetching impact

But that's another thing about I'm not sure how to present the results. I don't know how long it takes to loop on  batch in a real training like for natureLM.
If the time to go through a batch (backpropagation, optimizer steps,etc...) is higher than the time to pre-load the next batch then it will never be useful to use a higher prefetch_factor ?
My results are highly dependent on `sleep`and `batch_size`.

Issue #202, opened 2026-01-06
Mp3 read bug when using byte array directly with soundfile


GaganNarula:
@benjaminsshoffman  from slack:

I am having an issue with read_audio in the esp-data_ingestion branch of esp_data. For the file gs://esp-ml-datasets/wascher_corvid_comparison/audio/Corvus_corax/XC174222-Corcorax-C_Poland_PSZ_2014_04_10_2828.mp3, read_audio fails from GCP and also fails when reading locally. The traceback seems to blame libsndfile, but soundfile.read can read the file fine locally, and I can also open it in audacity.

>>> read_audio("gs://esp-ml-datasets/wascher_corvid_comparison/audio/Corvus_corax/XC174222-Corcorax-C_Poland_PSZ_2014_04_10_2828.mp3")
Error reading audio file Error opening <_io.BytesIO object at 0x7f30ac8d8540>: Format not recognised.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/home/esp-data/esp_data/io/read_utils.py", line 147, in read_audio
    raise e
  File "/mnt/home/esp-data/esp_data/io/read_utils.py", line 144, in read_audio
    return _read_audio_from_bytes(f.read(), frames, start)
  File "/mnt/home/esp-data/esp_data/io/read_utils.py", line 74, in _read_audio_from_bytes
    data, samplerate = sf.read(audio_buffer, frames=frames, start=start)
  File "/mnt/home/esp-data/.venv/lib/python3.10/site-packages/soundfile.py", line 305, in read
    with SoundFile(file, 'r', samplerate, channels,
  File "/mnt/home/esp-data/.venv/lib/python3.10/site-packages/soundfile.py", line 690, in __init__
    self._file = self._open(file, mode_int, closefd)
  File "/mnt/home/esp-data/.venv/lib/python3.10/site-packages/soundfile.py", line 1265, in _open
    raise LibsndfileError(err, prefix="Error opening {0!r}: ".format(self.name))
soundfile.LibsndfileError: Error opening <_io.BytesIO object at 0x7f30ac8d8540>: Format not recognised.

>>> read_audio("XC174222-Corcorax-C_Poland_PSZ_2014_04_10_2828.mp3")
Error reading audio file Error opening <_io.BytesIO object at 0x7f30acec3290>: Format not recognised.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #203, opened 2026-01-06
Fix for mp3 / ogg read from bytes

GaganNarula:
#202


esp_data/io/read_utils.py
```diff
@@ -174,7 +220,9 @@ def read_audio(
     try:
         fs = filesystem_from_path(file_path)
         with fs.open(str(file_path), "rb") as f:
-            return _read_audio_from_bytes(f.read(), frames, start)
+            # Extract format from extension (remove leading dot and uppercase)
+            audio_format = extension.lstrip(".").upper()
+            return _read_audio_from_bytes(f.read(), frames, start, format=audio_format)
```
mil-ad: remind me - why are we even doing `f.read()` and then use bytes? `soundfile.read()` support file-like objects which is what `fs.open()` returns. 

Wouldn't that work?

GaganNarula: I also dont know 😆 I will change it to `sf.read`.. looks like the code is taking bytes and re-converting to a file like object with BytesIO.

mil-ad: I'm hoping that if we let sf do the .read with its own logic it can figure out that it's an mp3 etc etc

mil-ad: worth trying before doing a refactor

esp_data/io/read_utils.py
```diff
@@ -102,10 +140,20 @@ def _read_audio_from_bytes(
...
-        data, samplerate = sf.read(audio_buffer, frames=frames, start=start)
-    return data, samplerate
+    try:
+        data, samplerate = sf.read(audio_file, frames=frames, start=start)
+        return data, samplerate
+    except Exception:
+        # logger.warning(
+        #     "Failed to read audio from file-like object directly, "
+        #     f"falling back to temporary file method: {e}"
+        # )
```
mil-ad: actually having a warning is not a bad idea?

esp_data/io/read_utils.py
```diff
@@ -174,7 +222,9 @@ def read_audio(
     try:
         fs = filesystem_from_path(file_path)
         with fs.open(str(file_path), "rb") as f:
-            return _read_audio_from_bytes(f.read(), frames, start)
+            # Extract format from extension (remove leading dot and uppercase)
```
mil-ad: this comment doesn't add anything. code is obvious

esp_data/io/read_utils.py
```diff
@@ -77,22 +112,25 @@ def read_yaml(path: str | AnyPathT) -> object:
     return result
 
 
-def _read_audio_from_bytes(
-    audio_bytes: bytes, frames: int = -1, start: int = 0
+def _read_audio_from_file(
+    audio_file: object, frames: int = -1, start: int = 0, format: str | None = None
```
mil-ad: is there anything better than `object`? maybe `IO` from `typing`?

esp_data/io/read_utils.py
```diff
@@ -102,10 +140,20 @@ def _read_audio_from_bytes(
...
         - samplerate (int): The sample rate of the audio in Hz.
     """
-
-    with io.BytesIO(audio_bytes) as audio_buffer:
-        data, samplerate = sf.read(audio_buffer, frames=frames, start=start)
-    return data, samplerate
+    try:
+        data, samplerate = sf.read(audio_file, frames=frames, start=start)
+        return data, samplerate
+    except Exception:
```
mil-ad: isn't there a more specific exception?

mil-ad (changes requested):
small changes

esp_data/io/read_utils.py
GaganNarula: yeah i thought about it but then if this happens for a lot of files the logs in the training runs might be bloated .. i can put it back ?

mil-ad (approved):

PR #206, opened 2026-01-09
fix uniform_sample

nkundiushuti:
- fix the transform to make it compliant with the current transforms 
- add some tests


esp_data/transforms/uniform_sample.py
```diff
@@ -45,46 +46,74 @@ class UniformSample:
...
+            If the specified property is not found in the DataFrame columns.
         """
+        if self.property not in backend.columns:
+            raise KeyError(f"Property '{self.property}' not found in the DataFrame columns.")
+
+        # Get all unique values for the property
+        unique_values = backend.get_unique(self.property)
+
+        # Create a ratios dict with the same ratio for all unique values
+        ratios = {value: self.ratio for value in unique_values}
```
GaganNarula: I thought we wanted to compute inverse probabilities and use them as ratios ? this logic maintains the original distribution but reduces the total size of the dataset by `self.ratio`

nkundiushuti: you are right, I didn't use it initially with this purpose but changed in the following commit

esp_data/transforms/uniform_sample.py
```diff
@@ -45,46 +46,132 @@ class UniformSample:
...
-        return pd.concat(groups, ignore_index=True), {}
+        if self.property not in backend.columns:
+            raise KeyError(f"Property '{self.property}' not found in the DataFrame columns.")
+
+        # Get all unique values for the property
+        unique_values = backend.get_unique(self.property)
+
+        # Get counts for each category to compute inverse probabilities
+        # We need to use unwrap temporarily to get category counts
+        df = backend.unwrap
```
GaganNarula: these two cases here, one for pandas and one for polars need to be added as methods to the backends in pandas_backend.py and polars_backend.py. 

So you should be calling something like `backend.histogram(column / feature: str = ...)` We dont want to use unwrap here

nkundiushuti: I added the histogram protocol and the backends

esp_data/transforms/uniform_sample.py
```diff
@@ -45,46 +46,111 @@ class UniformSample:
...
+                inverse_prob_ratios[value] = 0.0
+            else:
+                inverse_prob_ratios[value] = total_count / (num_categories * count)
+
+        # Normalize ratios so max is 1.0 (since we can't oversample)
+        # Then scale by self.ratio to control overall sample size
+        max_ratio = max(inverse_prob_ratios.values()) if inverse_prob_ratios else 1.0
+        if max_ratio > 0:
+            ratios = {
+                value: min(1.0, (inv_ratio / max_ratio) * self.ratio)
```
GaganNarula: do we really need `self.ratio` at all ? why would we want to reduce the size of our dataset ?

nkundiushuti: I am a bit biased because I used this transform to downsample but in theory we don't need any downsampling

nkundiushuti: should I just remove it then?

GaganNarula: imo it doesnt work in _this_ transform, but we could add a transform that does that ? We could rename this transform ^^ `BalancedSample` and the other with a fixed ratio `UniformSubsample`.. wdyt ?

GaganNarula:
i think david has the new wabad hash in his audio language PR ..you can already update here

GaganNarula (approved):
🚀

PR #207, opened 2026-01-12
audio-language transform

david-rx:
Adds a transform to convert any dataset into an audio-language one, adding prompt and response templates.

[TO BE CLOSED AFTER DISCUSSION, MOVED TO ESP-RESEARCH]

### DIFF
--- a/esp_data/transforms/audio_language.py
+++ b/esp_data/transforms/audio_language.py
@@ -1,0 +1,82 @@
+"""Audio-language dataset transformation."""
+
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel
+
+from esp_data.backends.protocol import DataBackend
+from esp_data.prompts import (
+    PromptResponseTemplate,
+    PromptResponseTemplateConfig,
+    get_prompt,
+)
+
+from . import register_transform
+
+
+class AudioLanguageConfig(BaseModel):
+    """Configuration for the AudioLanguage transform.
+
+    Parameters
+    ----------
+    type : Literal["audio_language"]
+        Must be "audio_language" for transform registry.
+    prompt : str | PromptResponseTemplateConfig | None
+        Prompt template name, inline config, or None for passthrough.
+    seed : int | None
+        Random seed for stochastic variant selection.
+    """
+
+    type: Literal["audio_language"]
+    prompt: str | PromptResponseTemplateConfig | None = None
+    seed: int | None = None
+
+
+class AudioLanguage:
+    """Transform to add 'prompt' and 'response' columns using a prompt template.
+
+    Parameters
+    ----------
+    prompt : str | PromptResponseTemplateConfig | None
+        - str: Name of a registered prompt template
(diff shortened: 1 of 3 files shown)


esp_data/transforms/audio_language.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @classmethod
+    def from_config(cls, cfg: AudioLanguageConfig) -> AudioLanguage:
+        return cls(prompt=cfg.prompt, seed=cfg.seed)
+
+    def __call__(self, backend: DataBackend) -> tuple[DataBackend, dict]:
+        prompts, responses = [], []
+
+        for row in backend:
+            result = self.template(row)
```
GaganNarula, line 72: there is an `apply_fn` method available in the `backend` which will be more appropriate than iterating over the rows, because iterating has to load audio etc.

esp_data/transforms/audio_language.py
```diff
@@ -0,0 +1,82 @@
...
+            self.template = PromptResponseTemplate(
+                name=prompt.name, variants=prompt.variants, seed=seed
+            )
+
+    @classmethod
+    def from_config(cls, cfg: AudioLanguageConfig) -> AudioLanguage:
+        return cls(prompt=cfg.prompt, seed=cfg.seed)
+
+    def __call__(self, backend: DataBackend) -> tuple[DataBackend, dict]:
+        prompts, responses = [], []
```
GaganNarula, line 69: will the `self.template` assume that the backend has "prompt" and "response" columns ?

esp_data/transforms/audio_language.py
david-rx: I didn't think iterating over the backend rows is loading audio like calling getitem?
My understanding is that apply_fn tends to be a little slower than iteration sadly

esp_data/transforms/audio_language.py
david-rx: I think yes

esp_data/transforms/audio_language.py
GaganNarula: oh yes sorry my mistake, no audio would be loaded, thats just for `Dataset` objects ! ok then lets stick to iterating.. simpler to follow anyway :D

esp_data/transforms/audio_language.py
GaganNarula: we should probably raise an error when this transform is applied on a dataset that has not been configured to return `prompt` and `response` in item dicts. For e.g., we could check whether these two columns exits in the data backend _or_ that the `output_take_and_give` dict is mapping some keys to `prompt` and `response`

david-rx: Oh sorry, I was confused. The template assumes that the fields expected by the template are present in the dataset's backend. E.g. if your template has a species_common key, that needs to be in the backend, or there will be an error

david-rx: That will be an error directly from jinja, but it could be worth a custom one to make it clear

PR #210, opened 2026-01-13
Update detection datasets

benjaminsshoffman:
I updated the underlying data with the GBIF corrections. WABAD was the only one affected and that hash was already updated. So, no major changes except adjusting things for the polar backend.

In the first commit, I removed the hard-coded species lists because they are brittle and can fail silently; generating the get_available_labels() is still very fast and typically should only be called once anyways.


GaganNarula (approved):
lgtm

PR #211, opened 2026-01-13
fixed no anypath in wabad dataroot bug

alf223:
gave wabad the same logic as powderdermill, nocturnal bird migration, etc.


GaganNarula (approved):
lgtm

PR #212, opened 2026-01-14
Webdataset backend

GaganNarula:
- Adds tarfile dataset read via webdataset library as a backend
- Adds `StreamingBackend` protocol which differs from `DataBackend` (should it be StreamingDataBackend?)


esp_data/backends/webdataset_backend.py
```diff
@@ -0,0 +1,557 @@
...
+        """
+
+        def dropna_fn(sample: dict[str, Any]) -> bool:
+            if subset is None:
+                # Check all columns
+                return all(sample.get(col) is not None for col in sample)
+            else:
+                return all(sample.get(col) is not None for col in subset)
+
+        self._dropna_fn = dropna_fn
```
mil-ad: I haven't looked at other backends but doesn't this pattern of storing in `self._dropna_fn()` means that we won't be able to do any chaining?

I'd expect to be able to do:

```
foo.backend.dropna_fn(...).dropna_fn()
```

and right now we only apply the last one.

mil-ad: same thing for filter_isin()

esp_data/backends/webdataset_backend.py
```diff
@@ -0,0 +1,557 @@
...
+    webds_kwargs = {"shardshuffle": shard_shuffle_size if shard_shuffle else False}
+    if shard_shuffle and seed is not None:
+        webds_kwargs["seed"] = seed
+    if split_by_worker:
+        webds_kwargs["workersplitter"] = wds.split_by_worker
+
+    webds = wds.WebDataset(shard_files, **webds_kwargs)
+
+    if shuffle_size:
+        webds = webds.shuffle(shuffle_size, seed=seed if seed is not None else 42)
```
mil-ad: doesn't `None` mean no shuffling?

mil-ad (commented):
few comments while looking at it more carefully

esp_data/backends/webdataset_backend.py
GaganNarula: yeah the implementation is a bit incomplete right now (these methods have to applied to the WebDataset obj) because I wanted you to have a quick look.  But you're right we should instead return a new WebDatasetBackend(). This is the pattern in the other backends

mil-ad: can we raise an issue for this?

esp_data/backends/__init__.py
```diff
@@ -2,17 +2,43 @@
...
-from .protocol import DataBackend
+from .protocol import DataBackend, StreamingBackend
+from .webdataset_backend import (
+    WebDatasetBackend,
+    audio_decoder,
+    audio_encoder,
+    json_decoder,
+    json_encoder,
+    load_webdataset,
+    make_file_opener_for_wds,
```
mil-ad: do we actually need to expose these? some of these are very generic names for very specific function

esp_data/backends/webdataset_backend.py
```diff
@@ -0,0 +1,586 @@
...
+        If the sample does not contain a "sample.json" key.
+    """
+    if "sample.json" not in data:
+        raise ValueError("Sample must contain 'sample.json' key with JSON data")
+
+    json_data = json.loads(data["sample.json"].decode("utf-8"))
+    return json_data
+
+
+def make_file_opener_for_wds(
```
mil-ad: is this used anywhere?

GaganNarula: it'll be used in a stacked PR for exporting.. but i guess i should move it to exporters/webdataset_utils.py ?

esp_data/backends/__init__.py
GaganNarula: we probably do need decoder one's but the encoder one's i could move to exporters/ see comment below

esp_data/backends/webdataset_backend.py
GaganNarula: i've changed the implementation in the recent commits, that allows for chaining . wdyt ?

esp_data/backends/webdataset_backend.py
mil-ad: I mean, at least remove from this PR and add it to the next one?

esp_data/backends/__init__.py
mil-ad: are they specific to WebDataset? it's weird to have

`from esp_data.backends import audio_encoder`

and then know that `audio_encoder` is part of webdataset

GaganNarula: i'll refactor, you're right it doesnt make sense

esp_data/dataset.py
```diff
@@ -454,6 +456,90 @@ def __str__(self) -> str:
...
+        **kwargs : Any
+            Additional arguments forwarded to the underlying backend loader.
+
+        Returns
+        -------
+        GenericDataset
+            Dataset instance with `_data` populated from the given path.
+        """
+        # Imported here to avoid circular import: generic_dataset imports Dataset
+        from esp_data.generic_dataset import GenericDataset
```
Paulchen-git, line 605: I'm not used to keep circular import, I feel like it isn't clean and safe generally, but here I'm still thinking how I could avoid that, wdyt ?

esp_data/dataset.py
```diff
@@ -615,6 +701,17 @@ def _make_dataset_from_config(dataset_config: DatasetConfig | ConcatConfig) -> D
     if _dataset_class is None:
         raise KeyError(f"Dataset '{dataset_config.dataset_name}' is not registered.")
 
+    if isinstance(dataset_config, DatasetConfig) and dataset_config.data_location:
+        ds = Dataset.from_path(
+            dataset_config.data_location,
+            backend=dataset_config.backend,
+            streaming=dataset_config.streaming,
+        )
+        if dataset_config.transformations:
```
GaganNarula: transformations shouldn't be applied here.. we have to think about whether transformations can be applied at all, but if they can they have to be done within `from_path` using `dataset_config.transformations`

esp_data/dataset.py
```diff
@@ -454,6 +456,90 @@ def __str__(self) -> str:
...
+        """
+        # Imported here to avoid circular import: generic_dataset imports Dataset
+        from esp_data.generic_dataset import GenericDataset
+
+        resolved = Path(str(path))
+
+        if resolved.is_dir():
+            data = WebDatasetBackend.from_path(path, **kwargs)
+        else:
+            data = get_backend(backend).from_path(path, streaming=streaming, **kwargs)
```
GaganNarula: you can just use `get_backend` instead of explicitly importing and instantiating `WebDatasetBackend` (no if - else needed),. We shouldnt check for `is_dir()` .. path is expected to be a dir anyway

esp_data/dataset.py
```diff
@@ -454,6 +456,90 @@ def __str__(self) -> str:
...
+
+        info = None
+        config_dir = resolved if resolved.is_dir() else resolved.parent
+        for config_name in ("config.json", "config.yaml"):
```
GaganNarula: lets assume its yaml

esp_data/dataset.py
GaganNarula: good question, this shouldn't really be done.. one solution would be to not add `@register_dataset` decorator to GenericDataset (and also not subclass it to Dataset). that breaks the pattern..lets keep it like this for now

esp_data/dataset.py
```diff
@@ -454,6 +456,90 @@ def __str__(self) -> str:
...
+
+        Returns
+        -------
+        GenericDataset
+            Dataset instance with `_data` populated from the given path.
+        """
+        # Imported here to avoid circular import: generic_dataset imports Dataset
+        from esp_data.generic_dataset import GenericDataset
+
+        resolved = Path(str(path))
```
GaganNarula: should be `anypath(path)` imo

esp_data/dataset.py
GaganNarula: actually this whole bit from line 525 to 539 should be moved to __init__ of GenericDataset. then __init__ only takes (path, **kwargs), not (data, info)

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Parameters
+        ----------
+        data : DataBackend | WebDatasetBackend
+            Already-loaded backend instance.
+        info : DatasetInfo | None, optional
+            If provided, overrides the class-level generic `info`.
+        """
+        self._is_webdataset = isinstance(data, WebDatasetBackend)
+        streaming = self._is_webdataset
```
GaganNarula: hmm, i think `streaming` should be an init param, not dependent on a BackendType. We should not mention any strict types in GenericDataset except literals like `backend="polars"`

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @property
+    def columns(self) -> Sequence[str]:
+        return self._data.columns
+
+    def _load(self) -> None:
+        pass
+
+    def __len__(self) -> int:
+        if self._is_webdataset:
```
GaganNarula: same issue here

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+            raise NotImplementedError(
+                "Length is not available for WebDataset backends. Iterate instead."
+            )
+        return len(self._data)
+
+    def __getitem__(self, idx: int) -> Dict[str, Any]:
+        if self._is_webdataset:
```
GaganNarula: and here

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+        yield from self._data
+
+    def __str__(self) -> str:
+        return (
+            f"GenericDataset(source={self.info.split_paths}, "
+            f"backend={'webdataset' if self._is_webdataset else 'tabular'})"
+        )
+
+    @classmethod
+    def from_config(cls, dataset_config: DatasetConfig) -> tuple["GenericDataset", dict[str, Any]]:
```
GaganNarula: hmm this is getting quite mixed up.. GenericDataset.from_config is calling Dataset.from_path within it.. we need a rethink

GaganNarula (commented):
some comments

esp_data/dataset.py
GaganNarula: we would probably want a GenericDatasetConfig ?

GaganNarula, line 805: maybe we can directly use a 'generic' key and do `GenericDataset.from_config` in dataset_from_config

esp_data/generic_dataset.py
GaganNarula: re my comment on dataset_from_config: if we introduce GenericDatasetConfig, then we dont need to use Dataset.from_path here. Rather just create `cls(path)`

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+        return self._data[idx]
+
+    def __iter__(self) -> Iterator[Dict[str, Any]]:
+        yield from self._data
+
+    def __str__(self) -> str:
+        return (
+            f"GenericDataset(source={self.info.split_paths}, "
+            f"backend={'webdataset' if self._is_webdataset else 'tabular'})"
```
GaganNarula: again _is_webdataset check.. we should maybe just print the self.info information in __str__.. reason is that if we overwrite `self.info` when we load from path, then print(ds) should reflect the actual dataset.. so it could be `anuraset` or something specific

esp_data/dataset.py
Paulchen-git: Ok for now, I've removed it

Paulchen-git:
After playing around with esp_data and webdataset backend, I'm currently confused. The save_to export I've made is working but is useless, you can't do much from a GenericDataset after having it loaded from tar files. maybe I've misunderstood the purpose of the export and load methods  ?                                                                                                                                                                                                                                                             
  
GenericDataset yields raw dicts with no per-sample processing. Round-trip AnuraSetStrong → save_to → from_path → GenericDataset loses all the domain logic (audio decoding, label mapping, typed fields).

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+            (e.g. ``data_processor`` for the webdataset backend).
+
+        Raises
+        ------
+        ValueError
+            If no ``info.yaml`` is found.
+        """
+        self.path = anypath(path)
+        fs = filesystem_from_path(self.path)
+        info_path = self.path / "info.yaml"
```
GaganNarula: hmm, the `info` object itself doesnt contain the params used to construct the dataset object.. for e.g. we need:

1. `backend_type`
2. `split`
3. `streaming`
4. `sample_rate`
5. .. other dataset specific params like `version` in audioset

So i think we should expect one file called `config.yaml`, which contains these params along with an `info` key which is the dict representing `DatasetInfo`

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+        if fs.exists(info_path):
+            info_dict = read_yaml(info_path)
+            self.info = DatasetInfo(**info_dict)
+        else:
+            raise ValueError(
+                f"No info.yaml found at {info_path}. Cannot infer backend or streaming mode."
+            )
+
+        super().__init__(backend=self.info.backend, streaming=self.info.streaming)
```
GaganNarula: same issue, `DatasetInfo` doesnt contain these so `self.info` wont have them

esp_data/generic_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+            raise RuntimeError(
+                "Random access is not supported for streaming backends. Iterate instead."
+            )
+        return self._data[idx]
+
+    def __iter__(self) -> Iterator[Dict[str, Any]]:
+        yield from self._data
+
+    def __str__(self) -> str:
+        return f"GenericDataset(name={self.info.name}, version={self.info.version})"
```
GaganNarula, line 108: we can add `split` and `sample_rate` etc to it

esp_data/dataset.py
```diff
@@ -454,6 +477,113 @@ def __str__(self) -> str:
         """
         pass
 
+    def save_to(self, path: str, format: str = "webdataset", **kwargs: Any) -> int:
+        """Save the dataset to a file.
+
+        Writes sharded tar files and an ``info.yaml`` containing `DatasetInfo`
```
GaganNarula: should not be specific to `tar` because that's just one format. As mentioned, we probably need `config.yaml` not `info.yaml`

esp_data/dataset.py
```diff
@@ -454,6 +477,113 @@ def __str__(self) -> str:
...
+        split_name = getattr(self, "split", "default")
+        info_dict["split_paths"] = {split_name: str(resolved)}
+
+        # Preserve sample_rate so reloaders know what rate was used.
+        sample_rate = getattr(self, "sample_rate", None)
+        if sample_rate is not None:
+            info_dict["sample_rate"] = sample_rate
+
+        # Record backend so from_path can reload without the user specifying it.
+        info_dict["backend"] = backend
```
GaganNarula: oh! i see you're overloading the `info` object here..

esp_data/backends/pandas_backend.py
```diff
@@ -1029,6 +1073,42 @@ def _row_to_ids(row: pd.Series) -> list | None:
...
+        int
+            Number of samples written.
+
+        Raises
+        ------
+        ValueError
+            If `format` is not supported
+        """
+        self._ensure_not_streaming("save_to")
+        if format == "webdataset":
```
GaganNarula: if our exporters are functions like `write_to_webdataset` (which i prefer) , then we can make an abstraction like `export_to(path, format)` which is a public function that does this selection .. and then we dont have to import a webdataset utility in a pandas backend! which is kind of a confusing pattern

Paulchen-git: I took a look to your tar_exporter PR, maybe we could merge it into the webdataset_backend PR, then I would add the export abstraction and I could use the code you wrote for tar files export

Paulchen-git: wdyt ?

Or I can add the export abstraction directly here

GaganNarula: and we can put this `export_to` into an export.py and make it public api from there.. the stuff like webdataset utils or format specific stuff shouldnt be public

GaganNarula: oh i just oh i just saw your message :D .. yes i think that works to merge this and the other branch, but maybe we should stack the export on this one, exposing a dummy export function for now

GaganNarula (commented):
some more comments

esp_data/backends/pandas_backend.py
Paulchen-git: #291 I open this stacked PR for the export abstraction

PR #214, opened 2026-01-14
Add infantmarmosetsvox dataset

EklavyaFCB:
Adds the [InfantMarmosetsVox](https://zenodo.org/records/10130104) dataset.

Implementation has two pre-sampled rates:
- 44.1 kHz (native)
- 16 kHz (downsampled via librosa)

Files uploaded to `gs://esp-ml-datasets/infant_marmosets_vox`.

### DIFF
--- a/esp_data/datasets/infant_marmosets_vox.py
+++ b/esp_data/datasets/infant_marmosets_vox.py
@@ -11,7 +11,7 @@ import polars as pl
 
 from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
 from esp_data.backends import BackendType, PolarsBackend
-from esp_data.io import AnyPathT, anypath
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
 
 
 # Call type mapping (0-10, excluding 11=silence and 12=noise)
@@ -266,19 +266,23 @@ class InfantMarmosetsVox(Dataset):
         """
         # Determine which audio directory to use based on requested sample rate
         # CSV path is like "data/audio_44k/twin_X/file.wav"
-        rel_path = row["path"]
+        audio_rel_path = row["path"]
 
         if self.sample_rate is not None and self.sample_rate in self._sample_rate_paths:
             # Use pre-resampled audio - replace "audio_44k" with appropriate subdirectory
             audio_subdir = self._sample_rate_paths[self.sample_rate]
-            rel_path = rel_path.replace("audio_44k", audio_subdir, 1)
+            audio_rel_path = audio_rel_path.replace("audio_44k", audio_subdir, 1)
 
-        audio_path = anypath(self.data_root) / rel_path
+        audio_path = anypath(self.data_root) / audio_rel_path
         start = float(row["start"])
         end = float(row["end"])
 
-        audio, sr = librosa.load(str(audio_path), sr=None, mono=True, offset=start, duration=end - start)
-
+        # Read the audio clip
+        audio, sr = read_audio(audio_path, start_time=start, end_time=end)
+        audio = audio.astype(np.float32)
+        # Stereo to mono if necessary.
+        audio = audio_stereo_to_mono(audio, mono_method="average")
+        # Normalize
         max_abs = np.max(np.abs(audio))
         if max_abs > 0:
             audio = audio / max_abs
@@ -288,6 +292,7 @@ class InfantMarmosetsVox(Dataset):
             audio = librosa.resample(y=audio, orig_sr=sr, target_sr=self.sample_rate, res_type="kaiser_best")
 
         row["audio"] = audio
+        row["path"] = audio_rel_path  # Update path to reflect actual file loaded
 
         if self.output_take_and_give:
             return {value: row[key] for key, value in self.output_take_and_give.items()}
--- a/esp_data/datasets/infant_marmosets_vox.py
+++ b/esp_data/datasets/infant_marmosets_vox.py
@@ -77,17 +77,26 @@ class InfantMarmosetsVox(Dataset):
         owner="eklavya",
         split_paths={
             # Single split - downstream users handle train/val/test splitting
-            "all": "gs://esp-ml-datasets/InfantMarmosetsVox/labels.csv",
+            "all": "gs://esp-ml-datasets/infant_marmosets_vox/labels.csv",
         },
         version="0.1.0",
         description=(
             "Infant marmoset vocalizations dataset. Contains 11 call types from "
             "10 individuals (5 twin pairs) recorded longitudinally over 8 months. "
-            "Approx. 73k vocalization segments after filtering silence/noise."
+            "Approx. 73k vocalization segments after filtering silence/noise. "
+            "Available at original 44.1kHz and pre-resampled 16kHz."
         ),
         sources=["Zenodo"],
         license="CC-BY-4.0",
     )
+
+    # Mapping of sample rates to audio subdirectory names
+    # CSV paths are like "data/audio_44k/twin_X/file.wav"
+    # We replace "audio_44k" with the appropriate subdirectory
+    _sample_rate_paths = {
+        44100: "audio_44k",  # Original 44.1kHz
+        16000: "audio_16k",  # Pre-resampled to 16kHz
+    }
 
     def __init__(
         self,
@@ -139,6 +148,19 @@ class InfantMarmosetsVox(Dataset):
         """Return the available splits of the dataset."""
         return list(self.info.split_paths.keys())
 
+    @property
+    def available_sample_rates(self) -> list[int]:
+        """Return the available pre-resampled sample rates.
+
+        Returns
+        -------
+        list[int]
+            List of sample rates (in Hz) for which pre-resampled audio is available.
+            Audio at these sample rates can be loaded directly without on-the-fly
+            resampling. Original audio is at 44100 Hz.
+        """
+        return list(self._sample_rate_paths.keys())
+
     def _load(self) -> None:
         """Load and preprocess the dataset.
 
@@ -159,7 +181,7 @@ class InfantMarmosetsVox(Dataset):
         df = pl.read_csv(location)
 
         # Extract twinID and per-file caller index from path
-        # path: "data/twin_1/20160907_Twin1_marmoset1.wav"
+        # path: "data/audio_44k/twin_1/20160907_Twin1_marmoset1.wav"
         # The "1" in "marmoset1" is the per-file caller (1 or 2)
         df = df.with_columns([
             pl.col("path").str.split("/").list.get(-1).str.replace(".wav", "").alias("filename"),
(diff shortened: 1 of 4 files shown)


esp_data/datasets/infant_marmosets_vox.py
```diff
@@ -0,0 +1,350 @@
...
+    >>> sample = dataset[0]
+    >>> sample["audio"].shape, sample["label"]
+    """
+
+    info = DatasetInfo(
+        name="InfantMarmosetsVox",
+        owner="eklavya",
+        split_paths={
+            # Single split - downstream users handle train/val/test splitting
+            "all": "gs://esp-ml-datasets/infant_marmosets_vox/labels.csv",
```
GaganNarula, line 92: might make sense to already do a train-val-test split, because if multiple users run their models on this dataset, they will all have incomparable splits

esp_data/datasets/infant_marmosets_vox.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # Read the audio clip
+        audio, sr = read_audio(audio_path, start_time=start, end_time=end)
+        audio = audio.astype(np.float32)
+        # Stereo to mono if necessary.
+        audio = audio_stereo_to_mono(audio, mono_method="average")
+        # Normalize
+        max_abs = np.max(np.abs(audio))
+        if max_abs > 0:
+            audio = audio / max_abs
```
GaganNarula: generally we don't normalize the audio in any of the datasets in esp-data. Just so that this fits with the other datasets i would not normalize in here, but rather in your collator / augmenter.

esp_data/datasets/infant_marmosets_vox.py
```diff
@@ -0,0 +1,350 @@
...
+            f"Available splits: {', '.join(self.info.split_paths.keys())}"
+        )
+
+    @property
+    def calltype_names(self) -> dict[int, str]:
+        """Return mapping from call type ID to name."""
+        return CALLTYPE_NAMES
+
+    @property
+    def num_calltypes(self) -> int:
```
GaganNarula: i think lets just keep `calltype_names` only as a property. The other two are derived from it. We're trying to keep the esp-data api's as uniform as possible.

esp_data/datasets/infant_marmosets_vox.py
```diff
@@ -0,0 +1,350 @@
...
+        split_paths={
+            # Single split - downstream users handle train/val/test splitting
+            "all": "gs://esp-ml-datasets/infant_marmosets_vox/labels.csv",
+        },
+        version="0.1.0",
+        description=(
+            "Infant marmoset vocalizations dataset. Contains 11 call types from "
+            "10 individuals (5 twin pairs) recorded longitudinally over 8 months. "
+            "Approx. 73k vocalization segments after filtering silence/noise. "
+            "Available at original 44.1kHz and pre-resampled 16kHz."
```
GaganNarula: Can you add a little sentence on what the "labels" of this dataset are ? if there are multiple, which columns are interesting to consider as labels, and which tasks do those serve.

GaganNarula (changes requested):
Needs some changes and also a test suite! look at some of the more recent one's like `test_geladas.py` in tests/unittests/dataset_tests/ and add one `test_infant_marmoset_vox.py` there

esp_data/datasets/infant_marmosets_vox.py
EklavyaFCB: Yes I saw there was no normalization in other datasets. I had it in mine because it's good practice I guess, but will move it.

esp_data/datasets/infant_marmosets_vox.py
EklavyaFCB: I kind of explain it earlier, above the references:

```
  Labels
  ------
  - `calltypeID`: Call type (0-10, 11 classes)
  - `callerID`: Caller identity (0-9, 10 individuals from 5 twin pairs)
```

Will update the `DatasetInfo`'s `description` to:

```
description=(
    "InfantMarmosetsVox is a dataset for multi-class call-type and caller identity classification. "
    "Available at original 44.1 kHz and pre-resampled 16kHz."
    "Pre-resampled audio uses librosa's kaiser_best resampling method. "
    "Contains approx. 73k vocalization segments of infant marmoset vocalizations. "
    "Each vocalization segment sample has a `calltypeID` and `callerID` label. "
    "There are 11 total calltype classes and 10 caller identity classes.
    "A calltypeID index can be associated with its calltype name with `CALLTYPE_NAMES`. "
)
```

That should be fine ?

esp_data/datasets/infant_marmosets_vox.py
EklavyaFCB: Removing `num_calltypes()` makes sense since it's just computed as `len(CALLTYPE_NAMES)`.

`num_callers()` currently is written badly, and should've instead been:

```python
@property
def num_callers(self) -> int:
    """Return number of unique callers."""
    return self._data._df["callerID"].n_unique()
```

Should I remove nonetheless ? It's computable from the internal dataframe (just unique number of callers in that column).

esp_data/datasets/infant_marmosets_vox.py
EklavyaFCB: Can make them, no problem. Previously I was doing a 70:20:10 train:val:test split through my datamodule using `random_split` with a fixed seed. Is that split ratio fine or is there a 'standard' ESP split ratio that you want ? 

Maintaining it as 70:20:10 will make the results comparable with all the previous results on this dataset from my PhD and other papers.

esp_data/datasets/infant_marmosets_vox.py
GaganNarula: ok lets keep `num_callers`

esp_data/datasets/infant_marmosets_vox.py
GaganNarula: you're right we should normalize, but max(abs()) normalization may not be the best way to do it. If we add it here its hard coded for this dataset

esp_data/datasets/infant_marmosets_vox.py
GaganNarula: 70:20:10 is fine! in fact if you have the previous split's indices (from your project) just you those!

EklavyaFCB: I don't have them directly, but I just recreated them by running the old code and saving them as .csv files. 
There were only 2 "issues" with the way I did created my existing splits:

- Since the samples are split across train, val, and test *randomly*, the calltype and caller distribution is probably uneven across splits. But this dataset's class distribution is very imbalanced by default in the combined 'all' split, so not sure there is avoiding this without data augmentation or something, which I'd rather avoid. 

- The splits are not separated by call-type or caller identity, so if e.g. one wanted to do a true open-set caller classification task, the trained model would have already seen all/most caller's vocalizations during training. For a closed-set identification task, this is not really an issue.

So I think I will keep the splits as I had done them, but add a line in the dataset info description to emphasize that the splits were computed randomly.

esp_data/datasets/infant_marmosets_vox.py
```diff
@@ -0,0 +1,410 @@
...
+        df = df.drop("_perFileCaller")
+
+        # Add row index
+        df = df.with_row_index("vocID")
+
+        # Reorder columns (drop original "caller" column from CSV, we computed callerID)
+        cols = ["path", "start", "end", "duration", "calltypeID", "callerID", "twinID", "vocID"]
+        df = df.select(cols)
+
+        self._data = PolarsBackend(df, streaming=self._streaming)
```
GaganNarula: oh this is a lot of computation, and specific only to Polars... is this a dataset preprocessing that must be done once ? if it is, I suggest adding a preprocessing script for this to the scripts/ folder in root, regenerate the dataset, and keep this clean (just for loading the backebd)

tests/unittests/datasets_tests/test_infant_marmosets_vox.py
```diff
@@ -0,0 +1,423 @@
...
+    """Test that duration values are positive."""
+    sample = ds_train[0]
+    duration = sample["duration"]
+    assert duration > 0, f"Duration {duration} should be positive"
+    assert sample["end"] > sample["start"], "End should be greater than start"
+
+
+if __name__ == "__main__":
+    # Run tests manually if executed directly
+    pytest.main([__file__, "-v"])
```
GaganNarula, line 383: we generally dont have this bit here, but lets keep it. I'll keep this pattern going forward

GaganNarula (changes requested):
sorry some more work! 😅

esp_data/datasets/infant_marmosets_vox.py
GaganNarula: hmm if you did it uniformly randomly it should probably maintain the overall shape (skewness etc) of the true distribution ? of course, some classes might be missing in test etc. Thanks for adding a comment, might even be better to use `logger.warning` to explicitly tell the user that :)

esp_data/datasets/infant_marmosets_vox.py
EklavyaFCB: This all the code that makes the internal dataframe once when instantiating the dataset, yes. It was pandas before I switched to polars to make it consistent with the everything. 

That being said, you're right that most of this preprocessing can already be done and saved within the .csv file that's loaded so that it doesn't have to be done here. 

One thing I do here is dropping all the rows that belong to the silence and noise categories. I can also simply remove those rows from the .csv file itself if we want to keep things fast ... but if a user wants to access those classes for whatever reason, they will never have access to them. 

Should I keep them ? 56% of the rows in the .csv are silence or noise classes.

esp_data/datasets/infant_marmosets_vox.py
EklavyaFCB: Sadly, I didn't do the splits with stratification either 🫤. I just did  `torch.utils.data.random_split()` (fully random) to stick with torch, when I should've done it as `sklearn.model_selection.train_test_split(stratify=something)`.

(Or perhaps I did try and it had some issue - it was ages ago in 2022 and I was working on a deadline).

So I guess the question here is if we want to spend some time and make new proper splits (for closed-set classification), or do we just put a warning and let people make their own using the  `all` set ?

esp_data/datasets/infant_marmosets_vox.py
GaganNarula: hmm well if you drop the rows anyway in the code then you might as well do that in the csv ? you could keep a copy of the csv with the silence and add a comment near the `split_paths` definition that a second csv exists that has those ?

GaganNarula: in general i am for just having all processing already done

esp_data/datasets/infant_marmosets_vox.py
GaganNarula: Ok then let's not make splits. Benji also said it's safer to make splits related to the tasks .. keep "all" :)

EklavyaFCB: Haha ok 🤣

GaganNarula (approved):
lgtm !

PR #216, opened 2026-01-14
add corvid dataset

benjaminsshoffman:
this is a corvid sounds dataset from claudia wascher, which we are using for the common calls project

### DIFF
--- a/esp_data/datasets/corvid_wascher.py
+++ b/esp_data/datasets/corvid_wascher.py
@@ -1,0 +1,240 @@
+"""Corvid dataset from Claudia Wascher"""
+
+from __future__ import annotations
+
+from io import StringIO
+from typing import Any, Dict, Iterator, List
+
+import librosa
+import numpy as np
+import pandas as pd
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.backends import BackendType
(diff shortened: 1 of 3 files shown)


esp_data/datasets/corvid_wascher.py
```diff
@@ -0,0 +1,240 @@
...
+    https://www.biorxiv.org/content/biorxiv/early/2024/07/17/2024.07.15.603339.full.pdf
+
+
+    """
+
+    info = DatasetInfo(
+        name="corvid_wascher",
+        owner="benjamin",
+        split_paths={
+            "all": "gs://esp-ml-datasets/wascher_corvid_comparison/all.csv",
```
GaganNarula, line 47: there are a lot of datasets now with just one split.. we should think about splitting them at some point to make official splits. It would make evals more automatic

GaganNarula (approved):
lgtm

esp_data/datasets/corvid_wascher.py
benjaminsshoffman: splitting implicitly makes decisions about task and type of generalization. IMO we shouldn't make these decisions unless we need to (otherwise it's potentially misleading or simply wasted effort). in all of the datasets with one split we've really only needed the one split, which is why i only submitted one.

Issue #215, opened 2026-01-14
Escape glob patterns for read_audio


benjaminsshoffman:
<img width="942" height="716" alt="Image" src="https://github.com/user-attachments/assets/0f91caa4-6021-4ab3-bc29-78a1968fd3f3" />

<img width="950" height="716" alt="Image" src="https://github.com/user-attachments/assets/36ef3ceb-b345-4119-abe8-4b7b544ea459" />

<img width="936" height="292" alt="Image" src="https://github.com/user-attachments/assets/04246924-6a32-4653-81da-69df852b998e" />

GaganNarula:
resolved (for now) in #218

PR #218, opened 2026-01-15
Allow glob patterns

GaganNarula:
#215

### DIFF
--- a/tests/test_fileio_functional.py
+++ b/tests/test_fileio_functional.py
@@ -27,7 +27,7 @@ def test_upload_download_cloud(local_test_dir, cloud_path):
     assert exists(local_file)
 
     # Upload to remote
-
+    rm(cloud_path)
     assert not exists(cloud_path)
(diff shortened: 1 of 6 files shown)


tests/test_fileio_functional.py
```diff
@@ -168,23 +150,24 @@ def test_list_files_in_cloud(cloud_dir, local_test_dir):
...
-    filesystem_from_path(remote_path).put(str(test_file), str(remote_path))
+#     assert not exists(remote_path)
+#     filesystem_from_path(remote_path).put(str(test_file), str(remote_path))
 
-    assert exists(remote_path)
-    rm(remote_path)
-    assert not exists(remote_path)
+#     assert exists(remote_path)
+#     rm(remote_path)
+#     assert not exists(remote_path)
```
mil-ad: can't we enable these with UUID?

GaganNarula: its just that we are testing `put` and `rm` already in the first test (upload, download).. this does almost exactly the same

mil-ad (approved):

Issue #217, opened 2026-01-15
BalancedSample undersamples too much


GaganNarula:
The BalancedSample transform sets a target_count for the data after resampling, but this target_count is set to the minimum count over values. This is too harsh! We need a strategy based method like 'min', 'max', 'median', 'median_with_range' etc

GaganNarula:
Solved in #219

PR #219, opened 2026-01-16
Allow for different resampling strategies other than 'min' count

GaganNarula:
#217 We can now do a few more strategies for sampling than 'min'

### DIFF
--- a/esp_data/transforms/balanced_sample.py
+++ b/esp_data/transforms/balanced_sample.py
@@ -1,7 +1,8 @@
 import logging
+import statistics
 from typing import Literal
 
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
 
 from esp_data.backends.protocol import DataBackend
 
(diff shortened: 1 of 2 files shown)


esp_data/transforms/balanced_sample.py
```diff
@@ -13,21 +14,40 @@
 class BalancedSampleConfig(BaseModel):
     type: Literal["balanced_sample"]
     property: str
+    strategy: Literal["min", "max", "median", "mean", "median_with_range"] = "median"
```
nkundiushuti, line 17: maybe a better name for median_with_range is percentile, but I have no strong opinion

esp_data/transforms/balanced_sample.py
```diff
@@ -89,28 +115,43 @@ def __call__(self, backend: DataBackend) -> tuple[DataBackend, dict]:
...
+        elif self.strategy == "max":
+            target_count = max(counts)
+            target_counts = {v: target_count for v in unique_values if category_counts[v] > 0}
+        elif self.strategy == "median":
+            target_count = int(statistics.median(counts))
+            target_counts = {v: target_count for v in unique_values if category_counts[v] > 0}
+        elif self.strategy == "mean":
+            target_count = round(statistics.mean(counts))
+            target_counts = {v: target_count for v in unique_values if category_counts[v] > 0}
+        elif self.strategy == "median_with_range":
```
nkundiushuti, line 140: this looks like percentile: https://en.wikipedia.org/wiki/Percentile

nkundiushuti (approved):

esp_data/transforms/balanced_sample.py
GaganNarula: i agree that its percentile, but its actually two percentiles, one for the lower and one for the upper so i just chose `median_with_range`

GaganNarula: but `percentiles` might be a better name actually, but lets do it in another PR. In this one i'm asking the user for just one param, the `range_fraction` otherwise we'll need lower_percentile and upper_percentile

PR #220, opened 2026-01-16
Fixed bug for audio files with extensions in caps

bobduru:
Fixed get_audio_info and read_audio_by_time to work with files that have extensions in uppercase.


GaganNarula (approved):
looks good!

PR #222, opened 2026-01-19
return sample rate in all datasets process

nkundiushuti:
- add sample_rate to row whenever we process a dataset 
- sample_rate is initialized to the sr of the original audio then set to the resample sr if resampling happens


nkundiushuti:
superbstarling test is failing, it should be fixed outside this PR

GaganNarula (commented):
looks fine, but maybe we should add a check in some of the dataset tests that sample_rate is returned ? just so that any future changes will need to maintain that field

GaganNarula (approved):
looks ok now, but the multiple sample rate datasets need a better, more rigorous treatment

PR #226, opened 2026-01-21
Moving GBIFConverter to esp-data

GaganNarula:
- Port GBIFConverter from taxonomy repo to esp-data as part of the new data discovery module (calling it `discover` for now, the idea would be to have a cli tool first like `uv run discover .. some queries ..`. Having taxonomy in this module will help to build a more comprehensive DatasetInfo c.f. #189 and do taxonomy validation.
- Added a `AddTaxonomy` transform that can help users add GBIF taxonomy to their datasets.
- Tests

### DIFF
--- a/esp_data/discover/gbif_taxonomy.py
+++ b/esp_data/discover/gbif_taxonomy.py
@@ -6,6 +6,7 @@ downloadable from Google Cloud Storage, but can be cached
 """
 
 import logging
+from pathlib import Path
 from typing import Any, Literal
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


esp_data/discover/gbif_taxonomy.py
```diff
@@ -0,0 +1,332 @@
...
+from esp_data.backends import DataBackend
+from esp_data.io import AnyPathT, exists, filesystem_from_path
+from esp_data.transforms import register_transform
+
+logger = logging.getLogger("esp_data")
+
+
+TAXONOMY_RANKS = ["kingdom", "phylum", "class", "order", "family", "genus"]
+# TODO: need a more managed location for this file
+DEFAULT_LOCATION = "gs://sound-event-detection/taxonomy/gbif_animals.tsv"
```
benjaminsshoffman: Do we want to have the preprocessing script as well? it's probably good in case we need to make changes in the future. It's located here, with explanation in the top docstring (currently, it requires to download the darwin core archive to local)

https://github.com/earthspecies/taxonomy/blob/main/scripts/v2_source_to_tsv.py

benjaminsshoffman (changes requested):
App looks good, but should have the preprocessing steps included for the future.

Also, adding some versioning would be really helpful. Could have a new folder in esp-ml-datasets for taxonomy, with gbif_animals_0_0_1.tsv as the first version? I'm guessing there's a more comprehensive way of doing versioning in the future, but this would be good until then

benjaminsshoffman (approved):
lgtm

PR #229, opened 2026-01-29
Add read_json()

mil-ad:

### DIFF
--- a/esp_data/io/read_utils.py
+++ b/esp_data/io/read_utils.py
@@ -1,5 +1,6 @@
 """This file offers functionalities necessary to read input streams, like audio."""
 
+import json
 import logging
 import tempfile
 from typing import Any, BinaryIO, Literal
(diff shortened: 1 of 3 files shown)


esp_data/io/read_utils.py
```diff
@@ -113,6 +114,40 @@ def read_yaml(path: str | AnyPathT) -> object:
...
+    try:
+        with filesystem_from_path(path).open(str(path), "r") as fp:
+            result = json.load(fp)
+    except json.JSONDecodeError as e:
+        raise json.JSONDecodeError(
+            f"Error parsing JSON file '{path}': {e.msg}", e.doc, e.pos
+        ) from e
+
+    if result is None:
+        raise ValueError(f"JSON file '{path}' is empty")
```
GaganNarula, line 146: when does this case happen ? if a user put a single `null` in the file why not just return it to them ?

GaganNarula (commented):
one comment, otherwise looks fine

esp_data/io/read_utils.py
GaganNarula: ah hold on, this matches what we do in `read_yaml`. We dont return None type.

GaganNarula (approved):
lgtm

PR #235, opened 2026-02-04
remove iterrows from arctic bird sounds

benjaminsshoffman:
there was an iterrows that didn't get properly removed when switching to backends. So I removed it


benjaminsshoffman:
I added tests that would have caught this bug

esp_data/datasets/arctic_bird_sounds.py
```diff
@@ -266,9 +266,11 @@ def get_available_labels(self, anno_column: str = "Species") -> List[str]:
         A list of all the available labels for anno_column
         """
         available_labels = set()
-        for _, row in self._data.iterrows():
+        for row in self._data:
             st = pd.read_csv(StringIO(row["selection_table"]), sep="\t")
             available_labels.update(st[anno_column].astype(str).tolist())
+        if self.unknown_label in available_labels:
+            available_labels.remove(self.unknown_label)
```
GaganNarula, line 274: is this kind of filtering-out of an unknown label specific to this dataset ? I feel that implementing a hard coded for one dataset is not a good pattern, when similar datasets to arctic_bird_sounds dont have the same operation... Also, wouldn't this operation show that the `available_labels` does not have `self.unknown_label` but the rows of the actual data are still not filtered for removing `unknown_label` ?

tests/test_birdeep.py
```diff
@@ -78,6 +78,11 @@ def test_ds_not_empty(ds: Birdeep):
     assert len(ds) > 0, "Dataset appears empty"
 
 
+def test_get_available_labels(ds: Birdeep):
+    """Dataset should provide available labels"""
+    labels = ds.get_available_labels()
+    assert len(labels) > 0, "Dataset has no labels"
```
GaganNarula: looks like `test_get_available_labels` is repeated in the same test suite below (lines 109 - 116)

tests/test_hawaiian_birds.py
```diff
@@ -79,6 +79,11 @@ def test_ds_not_empty(ds: HawaiianBirds):
     assert len(ds) > 0, "Dataset appears empty"
 
 
+def test_get_available_labels(ds: HawaiianBirds):
```
GaganNarula: same here, repeated test on line 121

GaganNarula (commented):
a couple of comments , but looks good

esp_data/datasets/arctic_bird_sounds.py
benjaminsshoffman: It's specific to datasets where there is an unknown label; it is already part of the behavior of NBM and Birdeep and the lack of filtering here was a mistake. The other detection datasets we have don't have unknown labels so it's not relevant.

The reason we don't want it in the available_labels is that Unknown is not a label, it's really a placeholder that indicates that we don't know the label. We typically use get_available_labels to build a model ontology or to know which classes to include in evaluation; we don't want either of these to include Unknown. 

I agree it's a bit fiddly, but I'd argue that the functionality is consistent across the datasets which have Unknowns, and including it in get_available_labels causes other problems that have to be handled externally.

tests/test_birdeep.py
benjaminsshoffman: Fixed; fixed throughout as well

tests/test_hawaiian_birds.py
benjaminsshoffman: Fixed; fixed throughout as well

benjaminsshoffman:
I accidentally pushed a commit which removed a doctest, and now there is one test that's failing. It still passes locally so I think the issue might be with some github cache, since the taxonomy app has been touched by another PR I made.

I think what makes sense is to get the taxonomy PR integrated, merge main -> this branch, and then hopefully all tests will be passing

esp_data/datasets/arctic_bird_sounds.py
GaganNarula: I think we need to think a bit deeper about this. I agree that you are handling it consistently across these datasets but we also need to consider users outside esp because this library will be open sourced soon. If they use `get_available_labels` and dont observe "Unknown" in the ontology, wouldn't they expect that any samples returned from `__getitem__` never contain any unexpected label ?

benjaminsshoffman: Happy to dig deeper, but maybe we could do it in a separate PR? I created this one to fix an urgent bug which simply crashes things. This feels like a broader discussion about assumptions about data formats which is independent of the bug fix. 

Just to make sure it is clear: this method of handling unknowns is already in esp-data via previous PRs (eg. https://github.com/earthspecies/esp-data/pull/161/changes#diff-7ed9d0488aac3deb4ce347cb124a71a4b591916dc8c18d30e12b4e0768505f9b); I am not creating/changing anything about the way unknowns are handled with this PR.

I will create an issue so we can discuss it in a less urgent way

GaganNarula: Yeah maybe I missed it when we first merged these in. Can you add a logger warning so the user knows that calling this method ignores the unknown label? We can discuss the issue separately, thanks for making one

GaganNarula (changes requested):
if we add a logger.warning for that unknown label and a TODO in the code referencing the issue you created then this is ready to go

benjaminsshoffman:
I added the warnings as requested

GaganNarula (approved):
lgtm!

PR #236, opened 2026-02-09
Speedup taxonomy

benjaminsshoffman:
I realized the taxonomy app is quite slow if you need to do many lookups. I wrote a script to cache every possible lookup. It took 3-4 days to run. Now the cached outputs live as their own json, which can be directly loaded by GBIFConverter. Instead of 3-4 days, it takes a few seconds to look up every possible conversion.

I wrote consistency tests and made the fast version the default behavior within GBIFConverter (as well as esp-data transforms)


scripts/data_preprocessing_scripts/cache_gbif_taxonomy_conversion.py
```diff
@@ -0,0 +1,33 @@
...
+
+import pandas as pd
+
+from esp_data.discover import GBIFConverter
+
+CACHE_PATH = "gs://esp-ml-datasets/gbif_taxonomy/v0.1.0/gbif_animals_converter_cache.json"
+
+cached = {}
+converter = GBIFConverter()
+to_lookup = list(converter.df_by_canonical_name.index)
```
GaganNarula: if we are looking up all the canonical names in this script to create the json file, why not just use this "cache" as the actual database ?

scripts/data_preprocessing_scripts/cache_gbif_taxonomy_conversion.py
```diff
@@ -0,0 +1,33 @@
...
+issues = []
+
+for i, lookup in enumerate(to_lookup):
+    if i % 100 == 0:
+        print(f"{i} / {len(to_lookup)} completed")
+    result, ok = converter(lookup)
+    if not ok:
+        issues.append(lookup)
+        continue
+    cached[lookup] = result
```
GaganNarula: any chance that duplicate canonical names with differing results would collide here and the last duplicate's result is stored ?

esp_data/discover/gbif_taxonomy.py
```diff
@@ -209,8 +280,12 @@ class AddTaxonomy:
     ----------
     feature : str
         Column name containing scientific names to look up.
-    gbif_taxonomy_db : str | AnyPathT
+    gbif_taxonomy_path : str | AnyPathT
         Path to GBIF taxonomy TSV file.
+    precomputed_fp : str | AnyPathT
```
GaganNarula, line 164: see comment below, we should just use the cache file , not the original

GaganNarula (changes requested):
great idea for the speedup, but lets just use it instead of the original ?

scripts/data_preprocessing_scripts/cache_gbif_taxonomy_conversion.py
benjaminsshoffman: Essentially that is what's happening, since the json reading is default. But we need the code to create the json. Is what you're suggesting to move the original logic to a script? (We still want tests run on it... is it ok to define classes in scripts and run tests on those classes?)

scripts/data_preprocessing_scripts/cache_gbif_taxonomy_conversion.py
benjaminsshoffman: I don't think so, GBIFConverter should be deterministic in its outputs so putting in the same canonical name twice would yield the same thing.

esp_data/discover/gbif_taxonomy.py
benjaminsshoffman: fixed

benjaminsshoffman:
@GaganNarula I moved all of the slow converter logic to a preprocessing script, which also runs tests. All the tests pass in the script (though I didn't re-run the 3 days of preprocessing). The GBIFConverter in esp_data.discover now only uses the pre-computed dict

GaganNarula (approved):
Looks good to me, but why is the test ci cancelled ? lets run it and see if there's any issue now

benjaminsshoffman:
> Looks good to me, but why is the test ci cancelled ? lets run it and see if there's any issue now

It's cancelled because the new .json is too big for CI to download. c/p from slack:

there is a failing integration test because github doesn’t want to download a 1.6 GB file (the precomputed ontology mappings). I think the options are to 1) remove the integration test or 2) allow the download and make sure github retains the file in its cache so it doesn’t require repeated downloads. I can do 1) but I don’t know how to do 2). Milad told me to ask you

GaganNarula:
Which integration test is it ? The BeansZero one ? Lets use pytest skip with the reason being the large file size

PR #237, opened 2026-02-10
Dclde

david-rx:
Adds the animal sound archive dataset and the DCLDE dataset

### DIFF
--- a/esp_data/datasets/animal_sound_archive.py
+++ b/esp_data/datasets/animal_sound_archive.py
@@ -72,9 +72,9 @@ class AnimalSoundArchive(Dataset):
     - ``train``: Training set (all minus 3000 held-out samples, random split)
     - ``validation``: Validation set (3000 samples, random split)
     - ``all``: Complete dataset (train + validation)
-    - ``train_unseen``: Training set excluding unseen taxa evaluated in BEANS-Zero benchmark
-    - ``validation_unseen``: Validation set excluding unseen taxa evaluated in BEANS-Zero benchmark
-    - ``all_unseen``: Complete dataset excluding BEANS-Zero unseen taxa
+    - ``train_excl_beanszero``: Training set excluding taxa evaluated in BEANS-Zero benchmark
+    - ``validation_excl_beanszero``: Validation set excluding taxa evaluated in BEANS-Zero benchmark
+    - ``all_excl_beanszero``: Complete dataset excluding BEANS-Zero taxa
 
     References
     ----------
@@ -106,9 +106,9 @@ class AnimalSoundArchive(Dataset):
             "train": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/train.csv",
             "validation": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/val.csv",
             "all": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/all.csv",
-            "train_unseen": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/train_unseen.csv",
-            "validation_unseen": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/val_unseen.csv",
-            "all_unseen": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/all_unseen.csv",
+            "train_excl_beanszero": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/train_unseen.csv",
+            "validation_excl_beanszero": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/val_unseen.csv",
+            "all_excl_beanszero": "gs://esp-ml-datasets/tierstimmenarchiv/v0.1.0/raw/all_unseen.csv",
         },
         version="0.1.0",
         description="Animal Sound Archive (Tierstimmenarchiv) audio dataset with "
--- a/esp_data/datasets/animal_sound_archive.py
+++ b/esp_data/datasets/animal_sound_archive.py
@@ -33,10 +33,10 @@ class AnimalSoundArchive(Dataset):
         - ``gbifID``: GBIF (Global Biodiversity Information Facility) identifier
 
     **Audio File Paths:**
-        - ``relative_path``: Path to original MP3 audio relative to audio root
+        - ``originals_path``: Path to original MP3 audio relative to data root
         - ``gcs_path``: Full GCS path to original audio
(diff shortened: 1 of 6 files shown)


esp_data/datasets/dclde2026.py
```diff
@@ -0,0 +1,545 @@
...
+        16000: "audio_16k",
+        32000: "audio_32k",
+    }
+
+    def __init__(
+        self,
+        split: str = "all",
+        output_take_and_give: dict[str, str] | None = None,
+        sample_rate: int | None = 16000,
+        providers: list[str] | None = None,
```
GaganNarula: we want to keep the Dataset api stable so these params shouldn't be included here. This requires a separate DCLDE2026Config

esp_data/datasets/dclde2026.py
```diff
@@ -0,0 +1,545 @@
...
+    @property
+    def columns(self) -> list[str]:
+        return list(self._data.columns) if self._data is not None else []
+
+    @property
+    def available_splits(self) -> list[str]:
+        return list(self.info.split_paths.keys())
+
+    @property
+    def available_providers(self) -> list[str]:
```
GaganNarula: lets not add more public properties .. this  convenience thing can be done by the user

esp_data/datasets/dclde2026.py
```diff
@@ -0,0 +1,545 @@
...
+        ``annotation_level``, ``confidence``.
+    provider : str
+        Data provider name (see :data:`PROVIDERS`).
+    16khz_path, 32khz_path : str | None
+        Paths to pre-resampled audio (when available).
+
+    Splits
+    ---------
+    - "all": All data (default)
+    - "unseen": All data, with unseen taxa held out for BEANS-Zero evaluation.
```
GaganNarula: i dont follow. So its all the data with some taxa that are part of BEANS-Zero removed ? I think the split name needs to be different. "unseen" sounds like this is a test set. Maybe something like "all-x-beanszero" or "all-w/o-beanszero"

esp_data/datasets/dclde2026.py
```diff
@@ -0,0 +1,545 @@
...
+        self._load()
+
+        # Filter to requested providers
+        if providers is not None:
+            unknown = set(providers) - set(PROVIDERS)
+            if unknown:
+                raise ValueError(
+                    f"Unknown providers: {sorted(unknown)}. Valid providers: {PROVIDERS}"
+                )
+            self._data = self._data.filter_isin("provider", providers)
```
GaganNarula: the whole provider-positive only thing seem like Transforms to me and they create different preprocessed versions of a dataset before training/eval. I dont think we should have this logic here.

esp_data/datasets/animal_sound_archive.py
```diff
@@ -0,0 +1,393 @@
...
+        - ``fieldNotes``: Observer's field notes (~70%)
+        - ``weather``: Weather conditions during recording (~35%)
+        - ``recordingEquipment``: Equipment used (~81%)
+
+    Available Splits
+    ----------------
+    - ``train``: Training set (all minus 3000 held-out samples, random split)
+    - ``validation``: Validation set (3000 samples, random split)
+    - ``all``: Complete dataset (train + validation)
+    - ``train_unseen``: Training set excluding unseen taxa evaluated in BEANS-Zero benchmark
```
GaganNarula: same issue here as "unseen" in dclde

esp_data/datasets/animal_sound_archive.py
```diff
@@ -0,0 +1,393 @@
...
+        dict[str, Any]
+            The processed row.
+        """
+        # Determine which path column to use based on requested sample rate
+        # If a pre-resampled version is available, use it; otherwise resample on-the-fly
+        use_presampled = False
+        if self.sample_rate is not None and self.sample_rate in self._sample_rate_paths:
+            path_column = self._sample_rate_paths[self.sample_rate]
+            if path_column in row and row[path_column] is not None and row[path_column] != "":
+                if self.sample_rate == 16000:
```
GaganNarula: saw this in another PR. why do we need different data roots ?

GaganNarula (changes requested):
We need to slim down the public api as well as internal logic in these datasets. I'm unfortunately going to have to be stricter because we want to open source in a few months.

esp_data/datasets/dclde2026.py
david-rx: Okay, I hear that. I added it because I shared with Benj and Emmanuel F who are more familiar with the data, and they said because the providers' data are quite different, we should make it clear combining them should be handled carefully and they may want to filter

david-rx: But I'll just leave a docstring note and do it the normal esp-data way, makes sense

esp_data/datasets/animal_sound_archive.py
david-rx: We don't, moved data around to simplify

david-rx:
There is still one test failing, which is because the audio for one provider failed to resample. My plan is to remove that provider: the audio is 1TB with extremely long individual files and not that much signal. I will leave a note and we can expand if needed.

esp_data/datasets/dclde2026.py
```diff
@@ -0,0 +1,453 @@
...
+            )
+            sr = self.sample_rate
+
+        # Parse selection table from serialized TSV
+        st = pd.read_csv(StringIO(row["selection_table"]), sep="\t", keep_default_na=False)
+
+        # Fill NaN in annotation string columns
+        for col in self.annotation_columns:
+            if col in st.columns:
+                st[col] = st[col].fillna("")
```
GaganNarula: this is a bit strange.. does this mean unknown label ? i know benj has a variable called `self.unknown_label` in the selection table datasets. maybe we should follow that pattern instead of hard coding "" ?

esp_data/datasets/dclde2026.py
```diff
@@ -0,0 +1,453 @@
...
+
+    def _process(self, row: dict[str, Any]) -> dict[str, Any]:
+        audio_fp, is_presampled = self._resolve_audio_path(row)
+
+        window_start = row.get("window_start_sec")
+        window_end = row.get("window_end_sec")
+
+        if window_start is not None and window_end is not None:
+            audio, sr = read_audio(
+                audio_fp, start_time=float(window_start), end_time=float(window_end)
```
GaganNarula, line 316: looks fine to me ! could check that `window_end > window_start` but not necessary, it will fail if that happens

GaganNarula (approved):
one comment, otherwise looks fine

esp_data/datasets/dclde2026.py
david-rx: You're right this is unnecessary, removing

PR #238, opened 2026-02-12
add long-tail and select columns

david-rx:
Adds two transforms:
- LongTailUpsample: A balancing transform slightly more tailored to bioacoustics, being used in both NatureLM-audio and SED so seems general enough to add
- select columns: filters to a subset of columns, used in NatureLM and flagged for addition to esp-data

### DIFF
--- a/esp_data/transforms/long_tail_upsample.py
+++ b/esp_data/transforms/long_tail_upsample.py
@@ -1,0 +1,162 @@
+import logging
+from typing import Literal
+
+from pydantic import BaseModel, field_validator
+
+from esp_data.backends.protocol import DataBackend
+
+from . import register_transform
+
+logger = logging.Logger("esp_data")
+
+
+class LongTailUpsampleConfig(BaseModel):
+    type: Literal["long_tail_upsample"]
+    property: str
+    sufficient_threshold: int
+    max_repeats: int
+    seed: int = 42
+
+    @field_validator("sufficient_threshold")
+    @classmethod
+    def validate_sufficient_threshold(cls, v: int) -> int:
+        if v < 1:
+            raise ValueError("`sufficient_threshold` must be >= 1")
+        return v
+
+    @field_validator("max_repeats")
+    @classmethod
+    def validate_max_repeats(cls, v: int) -> int:
+        if v < 1:
+            raise ValueError("`max_repeats` must be >= 1")
+        return v
+
+
+class LongTailUpsample:
+    """Upsample under-represented categories without excessive repetition.
+
+    Designed for long-tail distributions (e.g. bioacoustic species counts) where
+    a few categories dominate and many categories have very few examples. This
+    transform lifts the tail towards a `sufficient_threshold` while capping how
+    many times any single example can be repeated via `max_repeats`.
+
+    For each category with count *c*:
+
+    - If ``c >= sufficient_threshold``: the category is left untouched.
+    - If ``c < sufficient_threshold``: the target becomes
+      ``min(sufficient_threshold, c * max_repeats)``.
+
+    This produces a gradual compression of the distribution: well-represented
+    categories keep all their data, moderately-represented categories are boosted
+    to the threshold, and very rare categories are boosted as much as possible
+    without repeating any single example more than `max_repeats` times.
+
+    Parameters
+    ----------
+    property : str
+        The name of the property (column) to balance on.
+    sufficient_threshold : int
+        Categories with at least this many examples are left as-is. Categories
+        below this count are upsampled towards it, subject to `max_repeats`.
+    max_repeats : int
+        Maximum number of times any individual example may appear in the output.
+        Prevents over-fitting on very rare categories.
+    seed : int
+        Random seed for reproducibility. Defaults to 42.
+
+    Examples
+    -------
+    >>> from esp_data.transforms import LongTailUpsample, LongTailUpsampleConfig
+    >>> from esp_data.backends import PandasBackend
+    >>> import pandas as pd
+    >>> config = LongTailUpsampleConfig(
+    ...     type="long_tail_upsample",
+    ...     property="species",
+    ...     sufficient_threshold=6,
+    ...     max_repeats=3,
+    ...     seed=42,
+    ... )
+    >>> transform = LongTailUpsample.from_config(config)
+    >>> df = pd.DataFrame({
+    ...     "species": (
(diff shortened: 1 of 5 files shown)


esp_data/transforms/long_tail_upsample.py
```diff
@@ -0,0 +1,162 @@
...
+
+from . import register_transform
+
+logger = logging.Logger("esp_data")
+
+
+class LongTailUpsampleConfig(BaseModel):
+    type: Literal["long_tail_upsample"]
+    property: str
+    sufficient_threshold: int
```
GaganNarula, line 16: is this a common way to do this ? seems like the user would need to have some idea of the distribution already to compute `sufficient_threshold` already

esp_data/transforms/select_columns.py
```diff
@@ -0,0 +1,92 @@
...
+            A tuple containing the transformed backend with only the selected
+            columns and an empty metadata dictionary.
+
+        Raises
+        ------
+        KeyError
+            If any of the specified columns are not found in the backend.
+        """
+        missing = [c for c in self.columns if c not in backend.columns]
+        if missing:
```
GaganNarula, line 83: maybe nicer to just have `if any(c not in backend.columns for c in self.columns)`

esp_data/transforms/long_tail_upsample.py
david-rx: Yeah, I think this is tuned with some knowledge of what works well on your dataset. I don't mind having this be project-specific instead of merging it here if you prefer

david-rx: We'll need something to this effect in most of our foundations projects, but we could wait for the specific logic to stabilize

tests/test_long_tail_upsample.py
```diff
@@ -0,0 +1,256 @@
+import pandas as pd
+import polars as pl
+import pytest
+
+from esp_data.backends import PandasBackend, PolarsBackend
+from esp_data.transforms import LongTailUpsample, LongTailUpsampleConfig
+
+
+def _get_class_counts(
```
GaganNarula: couldnt this be replaced with the `backend.histogram` method ?

tests/test_long_tail_upsample.py
```diff
@@ -0,0 +1,256 @@
...
+    counts = _get_class_counts(result, backend_type)
+    assert counts["rare"] == 15
+
+
+@pytest.mark.parametrize("backend_type", ["pandas", "polars"])
+def test_long_tail_distribution(backend_type: str) -> None:
+    """End-to-end test with a realistic long-tail distribution."""
+    # common=500, moderate=100, uncommon=40, rare=3
+    # threshold=300, max_repeats=5
+    # Expected: common=500, moderate=300, uncommon=200, rare=15
```
GaganNarula, line 88: i think this is one of the issues i have with this transform (though I am not sure what another solution is). Once the transform is applied, it changes the relative ratios of the classes in the tail (for e.g. moderate / uncommon goes from 2.5 to 1.5 but uncommon / rare stays at about 13).. but ok, if its useful lets go ahead with it!

esp_data/transforms/long_tail_upsample.py
```diff
@@ -0,0 +1,162 @@
...
+    max_repeats: int
```
GaganNarula, line 17: maybe add constraint max_repeats >= 1

esp_data/transforms/select_columns.py
```diff
@@ -0,0 +1,92 @@
...
+            raise KeyError(
+                f"Columns {missing} not found in the DataFrame. "
+                f"Available columns: {backend.columns}"
+            )
+
+        return backend.select_columns(self.columns), {}
```
GaganNarula, line 89: i think we have to return the self.columns in the metadata no ? its empty right now

GaganNarula: actually forget that, not useful

esp_data/transforms/long_tail_upsample.py
```diff
@@ -0,0 +1,162 @@
...
+                    count * self.max_repeats,
+                )
+
+        sampled_backend = backend.upsample_by_column(
+            column=self.property,
+            target_counts=target_counts,
+            seed=self.seed,
+        )
+
+        return sampled_backend, {}
```
GaganNarula: i wonder if returning the histogram before and after applying the transform would be useful metadata 🤔 .. might help the user tune their thresholds ?

GaganNarula (changes requested):
a couple of changes but mostly looks good

tests/test_long_tail_upsample.py
david-rx: You could definitely design a continuous version that maybe fixes this? I think I liked this version in part because we are controlling based on knobs we somewhat understand (i.e.: I'm not sure what the relative ratios should be. But from a few papers we've seen what is considered "enough" representation of a species. And it feels intuitive how many copies of an audio becomes repetitive.) By the way the original intent was to also consider the audio duration as a parameter, upsampling long audios more aggressively since we take random clips each epoch. This would be a future version since audio dur is not pre-computed for most datasets.

I'm not convinced this is the final version either, so happy to have input!

esp_data/transforms/long_tail_upsample.py
david-rx: Good idea yeah

tests/test_long_tail_upsample.py
GaganNarula: lets keep this for now, i dont see a better solution as of now. Lets add the histograms to metadata, that should help the user tune it

esp_data/transforms/select_columns.py
david-rx: We have f"Columns {missing} not found.." which may be worth keeping?

esp_data/transforms/long_tail_upsample.py
david-rx: Yep - added as field validator

tests/test_long_tail_upsample.py
david-rx: sounds good

esp_data/transforms/select_columns.py
GaganNarula: oh yes you're right

GaganNarula (approved):
lgtm!

PR #239, opened 2026-02-18
watkins

david-rx:
Adds the watkins dataset. This was originally part of AnimalSpeak, but this is a fresh download to separate the dataset and avoid resampler issues. It additionally extracts structured call types. We don't include AnimalSpeak captions for now (similar to the new Xeno-canto and iNaturalist releases.)

### DIFF
--- a/esp_data/datasets/watkins.py
+++ b/esp_data/datasets/watkins.py
@@ -102,6 +102,13 @@ class Watkins(Dataset):
         32000: "32khz_path",
     }
 
+    _sample_rate_subdirs = {
+        16000: "audio_16k",
+        32000: "audio_32k",
+    }
+
+    _originals_path_column = "audio_path"
+
     def __init__(
         self,
         split: str = "train",
@@ -124,7 +131,7 @@ class Watkins(Dataset):
             32kHz), it will be loaded directly; otherwise audio is resampled
             on-the-fly.  ``None`` returns original sample rate.
         data_root : str | AnyPathT | None
-            Root directory prepended to ``audio_path``.  Defaults to the GCS
+            Root directory prepended to audio paths.  Defaults to the GCS
             bucket holding the original FLAC files.
         backend : BackendType
             DataFrame backend (``"polars"`` or ``"pandas"``).
@@ -138,12 +145,8 @@ class Watkins(Dataset):
 
         if data_root is None:
             self.data_root = anypath(_GCS_RAW_ROOT)
-            self._data_root_16k = anypath(f"{_GCS_RAW_ROOT}/audio_16k")
-            self._data_root_32k = anypath(f"{_GCS_RAW_ROOT}/audio_32k")
         else:
             self.data_root = anypath(data_root)
-            self._data_root_16k = anypath(data_root)
-            self._data_root_32k = anypath(data_root)
 
         self._load()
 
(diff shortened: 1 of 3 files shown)


esp_data/datasets/watkins.py
```diff
@@ -0,0 +1,281 @@
...
+from typing import Any, Iterator
+
+import librosa
+import numpy as np
+
+from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from esp_data.backends import BackendType
+from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+_GCS_RAW_ROOT = "gs://esp-data-ingestion/superwhale/v0.1.0/raw"
```
GaganNarula: why is it called `superwhale` ? wouldn't `watkins` be better ?

esp_data/datasets/watkins.py
```diff
@@ -0,0 +1,281 @@
...
+        -------
+        dict[str, Any]
+            The processed row with ``audio`` and ``sample_rate`` keys added.
+        """
+        use_presampled = False
+        if self.sample_rate is not None and self.sample_rate in self._sample_rate_paths:
+            col = self._sample_rate_paths[self.sample_rate]
+            if col in row and row[col] is not None and str(row[col]).strip():
+                if self.sample_rate == 16000:
+                    audio_path = self._data_root_16k / row[col]
```
GaganNarula: why do we need separate `self._data_root_16k`  and `..32k` ? i dont think we need these

GaganNarula (changes requested):
Some changes needed.. also what is the difference between this and watkins that is part of beans and beans-zero ?

david-rx:
> Some changes needed.. also what is the difference between this and watkins that is part of beans and beans-zero ?

In BEANS-Zero and BEANS, we have only the "best-of" cuts of Watkins, which is I think around 1200-2000 clips. This is the Watkins "All Cuts", which is around ~15,000

esp_data/datasets/watkins.py
david-rx: I was working on a combined dataset with a bunch of whale datasets in a common schema (name pending.) Watkins is important and should be on its own so I'll move it to "watkins"

esp_data/datasets/watkins.py
david-rx: You are right, fixed

tests/test_watkins.py
```diff
@@ -0,0 +1,236 @@
...
+    sample = ds[0]
+    assert sample["audio"].dtype == np.float32
+    assert len(sample["audio"]) > 0
+    assert isinstance(sample["sample_rate"], int)
+
+
+def test_data_root_handling(dataset: Dataset) -> None:
+    """Test if data_root parameter works correctly."""
+    assert dataset.data_root is not None
+
```
GaganNarula, line 220: Lets add annotation hashes like in benj's datasets like arctic_bird_sounds

tests/test_watkins.py
```diff
@@ -0,0 +1,236 @@
...
+@pytest.fixture
+def dataset() -> Dataset:
+    """Fixture providing a Watkins dataset instance."""
+    return Watkins(split="train")
+
+
+@pytest.fixture
+def dataset_with_transforms_from_config() -> tuple[Dataset, dict]:
+    """Fixture providing a Watkins dataset with transformations applied from config."""
+    dataset_config = DatasetConfig(
```
GaganNarula: A few things to change here: when possible, using streaming=True and backend=polars. If a non-streaming dataset is needed, use a small split. Also, repeat a dataset fixture as much possible. We really need to cut down on the test run time

GaganNarula (changes requested):
a few changes needed

tests/test_watkins.py
david-rx: Good to know

GaganNarula (approved):
lgtm!

PR #240, opened 2026-02-19
add wabad subdatasets

benjaminsshoffman:
This adds sub-dataset splits for WABAD. The sub-datasets each have their own ontology and were annotated by different groups, resulting in different annotation conventions. Accessing each individually gives us the ability to control for these effects.


esp_data/datasets/wabad.py
```diff
@@ -110,7 +184,7 @@ def __init__(
         self.sample_rate = sample_rate
         self.data_root = anypath(data_root) if data_root is not None else None
 
-        self.available_labels = pd.read_csv(SPECIES_INFO_PATH)["Species"].to_list()
+        self.full_dataset_available_labels = pd.read_csv(SPECIES_INFO_PATH)["Species"].to_list()
```
GaganNarula: If the full dataset labels are only used when the split is "all" we can call read_csv when "get_available_labels" is called (read file and set this variable)

benjaminsshoffman: Done

GaganNarula (approved):
looks good!

PR #243, opened 2026-02-22
Refresh birdset

david-rx:
Original BirdSet implementation was incomplete and had minor issues:
- It was missing a few core test datasets (UHH, SNE, SSW)
- The audios were resampled to 48khz with an unknown resampler, while the originals were at 32khz
- It used old taxonomy, possibly other minor differences from the original because it was processed for use with NatureLM.

This refreshes to the complete BirdSet.

### DIFF
--- a/esp_data/datasets/birdset.py
+++ b/esp_data/datasets/birdset.py
@@ -11,17 +11,6 @@ from esp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
 
 _GCS_ROOT = "gs://esp-ml-datasets/birdset/v0.2.0/raw"
 
-_SUBSETS = ("HSN", "NBP", "NES", "PER", "POW", "SSW", "SNE", "UHH")
-
-
-def _build_split_paths() -> dict[str, str]:
-    paths: dict[str, str] = {}
-    for subset in _SUBSETS:
-        for split in ("test", "test_5s"):
-            paths[f"{subset}-{split}"] = f"{_GCS_ROOT}/{subset}_{split}.csv"
-    paths["all"] = f"{_GCS_ROOT}/birdset_all.csv"
-    return paths
-
 
 @register_dataset
 class BirdSet(Dataset):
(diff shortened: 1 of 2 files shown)


esp_data/datasets/birdset.py
```diff
@@ -1,86 +1,134 @@
...
+    but is a subset of the Xeno-canto dataset.
+
+    Available Metadata Fields
+    -------------------------
+    **Taxonomic Information:**
+        - ``species``: Scientific species name (resolved from eBird code)
+        - ``species_common``: Common English name
+        - ``ebird_code``: eBird species code (primary label)
+        - ``ebird_code_multilabel``: JSON list of all species codes in recording
+        - ``genus``, ``order``: Taxonomic hierarchy
```
GaganNarula, line 39: are these taxonomic fields generated by the GBIFConverter ?

esp_data/datasets/birdset.py
```diff
@@ -1,86 +1,134 @@
...
+        - ``start_time``, ``end_time``: Temporal boundaries (seconds)
+        - ``low_freq``, ``high_freq``: Frequency range (Hz); present for
+          ``test`` soundscape splits, empty for ``test_5s`` clips
+
+    Available Splits
+    ----------------
+    Each of the eight evaluation subsets has two splits:
+
+    - ``{SUBSET}-test``: Full-length soundscape recordings (variable duration)
+    - ``{SUBSET}-test_5s``: 5-second clips extracted from test recordings
```
GaganNarula, line 64: how were these 5 second clips extracted ?

esp_data/datasets/birdset.py
```diff
@@ -1,86 +1,134 @@
...
-            "UHH-validation": "gs://foundation-model-data/data/birdset-train/UHH/UHH_taxonomic.jsonl",
-            "UHH-test": "gs://foundation-model-data/data/birdset-test/UHH/UHH_taxonomic.jsonl",
-            "SSW-train": "gs://foundation-model-data/data/birdset-train/SSW/SSW_taxonomic.jsonl",
-            "SSW-validation": "gs://foundation-model-data/data/birdset-train/SSW/SSW_taxonomic.jsonl",
-            "SSW-test": "gs://foundation-model-data/data/birdset-test/SSW/SSW_taxonomic.jsonl",
-            "SNE-train": "gs://foundation-model-data/data/birdset-train/SNE/SNE_taxonomic.jsonl",
-            "SNE-validation": "gs://foundation-model-data/data/birdset-train/SNE/SNE_taxonomic.jsonl",
-            "SNE-test": "gs://foundation-model-data/data/birdset-test/SNE/SNE_taxonomic.jsonl",
-            "XCM": "gs://foundation-model-data/data/birdset-train/XCM/XCM_taxonomic.jsonl",
+            "HSN-test": f"{_GCS_ROOT}/HSN_test.csv",
```
GaganNarula, line 97: what happened to the train and validation subsets ?

esp_data/datasets/birdset.py
david-rx: These are directly from the BirdSet source.

esp_data/datasets/birdset.py
david-rx: Yes

esp_data/datasets/birdset.py
david-rx: I am not sure what to do here:
- In -most- projects at ESP, we only use the BirdSet test set. It's true we should include the train set if we brand the release as BirdSet, and sometimes we may want to use it e.g. for probing.
- The train set is huge (Xeno-canto.) We could try pointing to the source xeno-canto, but I doubt mapping will be clean. We could re-process from source: however, I do not have the capacity to take this on currently, so the pr will hang, which will leave the current BirdSet version up which is more incorrect (sampling rate issues, derived labels, and missing 3 test sets.) A final option could be to keep the old train / val sets in, and re-process in a follow-up?
- The validation sets included in the previous release are not "official". If I understand correctly, they were a randomly held-out subset of train.

GaganNarula: if we are sure that no will need the original BirdSet (v0.1.0) in esp-data, then I am fine with replacing it.  I agree we shouldn't have an incorrect version. Maybe lets just delete v0.1.0 from esp-ml-datasets and replace that with this ?

GaganNarula (commented):
one last comment but otherwise ok 👌

esp_data/datasets/birdset.py
david-rx: It would be necessary to exactly recreate the probing experiments from the What Matters paper, but imo this is okay to remove (the paper is accepted, and we can migrate to the clean setup.) So sounds good to me

tests/test_birdset.py
```diff
@@ -1,217 +1,233 @@
...
-    Dataset
-        An instance of the BirdSet dataset with output mapping applied.
-    """
-    dataset_config = DatasetConfig(
-        dataset_name="birdset",
-        output_take_and_give={"label": "Label"},
-    )
-    ds = BirdSet(split="HSN-test", output_take_and_give=dataset_config.output_take_and_give)
-    return ds
+# TODO: fill in once v0.2.0 data lands on GCS
```
GaganNarula: lets just add the hashes now .. there are no other PRs that we are blocking as of now.

david-rx: oops yes

GaganNarula (approved):
lgtm!

PR #245, opened 2026-02-24
More subsegmentation

benjaminsshoffman:
This adds a new version of the subsegmentation dataset where individual data items are single songs, rather than full files. The underlying data splits remain the same. This is for a universal subsegmentation model training/eval


GaganNarula (approved):
lgtm

PR #251, opened 2026-03-04
Test self-hosted k8 runner

GaganNarula:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,66 @@
...
+        name: Run pytests tests
+        runs-on: self-hosted
+
+        steps:
+            - uses: actions/checkout@v4
+
+            - name: Install uv
+              uses: astral-sh/setup-uv@v4
+              with:
+                  enable-cache: true
```
mil-ad: this should be `false` or `"auto"`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,66 @@
+name: Tests CI
+
+on:    # yamllint disable-line rule:truthy
+    workflow_dispatch:
+    push:
+        branches: ["main"]
+    pull_request:
+
+jobs:
+    test-self-hosted:
```
mil-ad: I wouldn't put `-self-hosted` in the job's name

mil-ad (changes requested):

.github/workflows/ci.yml
GaganNarula: why do we want it to be "auto" ?

mil-ad (approved):

PR #255, opened 2026-03-17
Add spanish carrion crows dataset

mcusi:
I'm adding the unsynchronized, audio-only version of the carrion crow biologger dataset. It is a detection dataset (Voxaboxen bounding boxes). I followed WABAD/ArcticBirdSounds as templates.

### DIFF
--- a/esp_data/datasets/spanish_carrion_crows.py
+++ b/esp_data/datasets/spanish_carrion_crows.py
@@ -6,6 +6,7 @@ from io import StringIO
 from typing import Any, Iterator
 
 import librosa
+import numpy as np
 import pandas as pd
 
 from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
@@ -38,7 +39,7 @@ class SpanishCarrionCrows(Dataset):
         - Synchronized UTC timestamps across biologgers.
         - Annotations of file quality
           e.g., some biologgers prematurely detached from the bird and recorded
-          background sounds only
+          background sounds only. Therefore some selection tables are empty.
         - Call types
     If any additional data are needed, please contact Maddie.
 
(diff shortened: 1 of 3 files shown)


esp_data/datasets/spanish_carrion_crows.py
```diff
@@ -0,0 +1,343 @@
...
+            "2021_LL_Verde": "gs://esp-ml-datasets/spanish-carrion-crows/v0.1.0/detections/2021_LL_Verde__spanish-carrion-crows_info.csv",
+            "2021_N4_Naranja": "gs://esp-ml-datasets/spanish-carrion-crows/v0.1.0/detections/2021_N4_Naranja__spanish-carrion-crows_info.csv",
+            "2021_N4_Verde": "gs://esp-ml-datasets/spanish-carrion-crows/v0.1.0/detections/2021_N4_Verde__spanish-carrion-crows_info.csv",
+        },
+        version="0.1.0",
+        description="Crow biologger audio with Voxaboxen detections",
+        sources="University of Leon",
+        license=(
+            "For ESP internal, non-commerical use only. "
+            "Should talk to collaborators about usage in work to be published."
```
GaganNarula: can we call it "private" ? i'm trying to standardize these. You can leave a comment about asking colleagues about `license`

esp_data/datasets/spanish_carrion_crows.py
```diff
@@ -0,0 +1,343 @@
...
+            if path_column in row and row[path_column] is not None and row[path_column] != "":
+                audio_path = anypath(self.data_root) / row[path_column]
+                use_presampled = True
+
+        if not use_presampled:
+            audio_path = anypath(self.data_root) / row[self._originals_path_column]
+
+        audio, sr = read_audio(audio_path)
+        # Should all be mono
+        # audio = audio_stereo_to_mono(audio, mono_method="average").astype(np.float32)
```
GaganNarula: you can just remove the line if its already mono (maybe test with an assert in your tests)

esp_data/datasets/spanish_carrion_crows.py
```diff
@@ -0,0 +1,343 @@
...
+            audio = librosa.resample(
+                y=audio,
+                orig_sr=sr,
+                target_sr=self.sample_rate,
+                scale=True,
+                res_type="kaiser_best",
+            )
+            sr = self.sample_rate
+
+        st = pd.read_csv(StringIO(row["selection_table"]), sep="\t")
```
GaganNarula, line 199: are there any selection tables that might be empty and is it an issue to return empty selection_table items? its happened before in other datasets, just checking

esp_data/datasets/spanish_carrion_crows.py
```diff
@@ -0,0 +1,343 @@
...
+    _sample_rate_paths: dict[int, str] = {}
+    _originals_path_column = "audio_fp"
+
+    def __init__(
+        self,
+        split: str = "all",
+        output_take_and_give: dict[str, str] | None = None,
+        sample_rate: int | None = 16000,
+        data_root: str | AnyPathT | None = None,
+        backend: BackendType = "pandas",
```
GaganNarula: polars is faster, consider replacing

esp_data/datasets/spanish_carrion_crows.py
```diff
@@ -0,0 +1,343 @@
...
+    TODO: Add preprint link.
+
+    """
+
+    info = DatasetInfo(
+        name="spanish-carrion-crows",
+        owner="maddie",
+        split_paths={
+            "all": "gs://esp-ml-datasets/spanish-carrion-crows/v0.1.0/detections/all__spanish-carrion-crows_info.csv",
+            "2018_AW_Roja": "gs://esp-ml-datasets/spanish-carrion-crows/v0.1.0/detections/2018_AW_Roja__spanish-carrion-crows_info.csv",
```
GaganNarula: Do we need so many individual splits ? maybe users can just split "all" using a transform ? more splits = more things to manage .. also there's

tests/test_spanish_carrion_crows.py
```diff
@@ -0,0 +1,197 @@
...
+    "1313ead69001c6f7bcd52672e90d51f3f138c73180afca19a7013d562b1877f9"
+)
+ANNOTATIONS_SHA256 = "eb8f7f74193a0eedc611c2ae188583a8ab107d57afebc77db0697597ac8ae81c"
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def ds() -> SpanishCarrionCrows:
+    """Load SpanishCarrionCrows dataset for testing."""
+    return SpanishCarrionCrows(split="all", sample_rate=16000, backend="pandas")
```
GaganNarula: always run tests with `streaming=True` and `backend=polars` ... helps reduce time taken to run tests

GaganNarula (commented):
some comments to discuss but in principle it looks good!

esp_data/datasets/spanish_carrion_crows.py
mcusi: Yes, changed to "private"

esp_data/datasets/spanish_carrion_crows.py
mcusi: No, we don't. I think users can easily split all using a transform (filter to desired individual in column focal_individual). I'll remove these

esp_data/datasets/spanish_carrion_crows.py
mcusi: Ok removed and added a test. I realized that I took out the .astype(np.float32) as well when I did this. Is it important to keep float32 for any reason? (The tests check for float64 so it's consistent with itself right now)

esp_data/datasets/spanish_carrion_crows.py
mcusi: Yes, there are empty selection tables (but no empty strings in the `selection_table` column). 

I did make a note of this in the dataset description:
>           e.g., some biologgers prematurely detached from the bird and recorded
>           background sounds only

@jcauzi @david-rx would this be a problem for you? I can filter those out beforehand if you want

esp_data/datasets/spanish_carrion_crows.py
mcusi: Changed

tests/test_spanish_carrion_crows.py
mcusi: I can switch the backend. But a lot of the tests require len or indexing, so I think I cannot do that with `streaming=True`, or is that mistaken? (I can at least do it where that's not required)

esp_data/datasets/spanish_carrion_crows.py
mcusi: For now adding a more specific note to the description.

mcusi: @david-rx said it's okay to have empty selection tables

esp_data/datasets/spanish_carrion_crows.py
GaganNarula: well float32 saves memory (especially GPU).. and float64 doesn't really offer much more _useful_ precision for audio ?

the other thing is that every dataset (I believe) in esp_data is enforcing float32 so having float64 in just this one will break experiments where this dataset is concatenated / chained with others in a training loop ... wdyt ?

tests/test_spanish_carrion_crows.py
GaganNarula: ok you're right, also i think the csv for "all" is probably not that big anyway. 👍

esp_data/datasets/spanish_carrion_crows.py
mcusi: Okay, I'll leave in the conversion to float32

mcusi:
I think I addressed all the comments so far now. Jules said it would be useful to add call types, I will do that in a second version

PR #256, opened 2026-03-17
Expand taxonomy

benjaminsshoffman:
This PR expands taxonomy support tools:

- Added some manual corrections to resolve places where GBIFConverter doesn't work out-of-the-box. These corrections were required for the SED project and took a lot of labor to put together. The conversion lives as a python dict since it's not super big (relatively speaking), and so that adding lines is easy in the future.
- Added conversions and transforms for going from ebird codes -> Clements taxonomy scientific names and vice versa. This can be done for different years of the ebird/clements taxonomy (default is 2021 since that's what a lot of things eg Perch use).
- Added a conversion for GBIF->Clements (clements -> GBIF is already implemented via GBIFConverter)
- Putting these together, it's now possible to go ebird -> GBIF and GBIF -> ebird through calling converters or through dataset transforms.


esp_data/discover/taxonomy_manual_corrections.py
```diff
@@ -0,0 +1,288 @@
+"""
+This file implements manual taxonomy corrections that are not accessible
+using the logic in GBIFConverter. These corrections are mostly due to
+spelling differences not documented by GBIF, and also to taxonomy changes
+that happened after GBIF's most recent taxonomy release.
+"""
+
```
GaganNarula: I think we dont need a separate python file for this, you can just add this dict to the gbif_taxonomy.py and make changes to the When a query is in this dict, return the value immediately GBIFConverter __call__ step

GaganNarula (changes requested):
So i think I'm going have to ask that we don't add ebird and clements taxonomies / converters to esp-data just yet. Lets create an issue and move the code to a branch if needed. The thing is that we need to keep scope a bit tighter moving forward and currently, it looks like only the SED project needs these mappings. When it becomes a more pertinent issue and demanded from several projects we can reopen a PR

esp_data/discover/taxonomy_manual_corrections.py
benjaminsshoffman: I will move dict into a single file.

The change that you suggested isn't good, for three reasons:

1. The output format of this dict is inconsistent with GBIFConverter; GBIFConverter returns a dict with fields for canonical (binomial) name, as well as higher orders (family, order, etc)
2. Putting the manual correction through the GBIFConverter's main lookup **validates** the output, in that we are guaranteed that it is an accepted GBIF name. If we shortcut this step, the outputs are not validated
3. In light of that, I did not take the time to ensure that each correction was the accepted name in GBIF, ie some of the values are synonyms. This is fine because the GBIFConverter resolves the synonyms to the accepted name with its main lookup. In other words, issue 2 isn't merely theoretical

benjaminsshoffman:
> So i think I'm going have to ask that we don't add ebird and clements taxonomies / converters to esp-data just yet. Lets create an issue and move the code to a branch if needed. The thing is that we need to keep scope a bit tighter moving forward and currently, it looks like only the SED project needs these mappings. When it becomes a more pertinent issue and demanded from several projects we can reopen a PR

I'll ask about it on slack. My guess is that others will want to use it soon. These conversions are required as we run baselines (eg perch and audioprotopnet), which are not constrained to the SED project.

esp_data/discover/taxonomy_manual_corrections.py
GaganNarula: ah sorry my mistake, what you've implemented looks correct

GaganNarula:
> > So i think I'm going have to ask that we don't add ebird and clements taxonomies / converters to esp-data just yet. Lets create an issue and move the code to a branch if needed. The thing is that we need to keep scope a bit tighter moving forward and currently, it looks like only the SED project needs these mappings. When it becomes a more pertinent issue and demanded from several projects we can reopen a PR
> 
> I'll ask about it on slack. My guess is that others will want to use it soon. These conversions are required as we run baselines (eg perch and audioprotopnet), which are not constrained to the SED project.



> > So i think I'm going have to ask that we don't add ebird and clements taxonomies / converters to esp-data just yet. Lets create an issue and move the code to a branch if needed. The thing is that we need to keep scope a bit tighter moving forward and currently, it looks like only the SED project needs these mappings. When it becomes a more pertinent issue and demanded from several projects we can reopen a PR
> 
> I'll ask about it on slack. My guess is that others will want to use it soon. These conversions are required as we run baselines (eg perch and audioprotopnet), which are not constrained to the SED project.

Hey I think lets keep these ebird <-> Clements <-> GBIF conversions for a later release of esp-data. i'm trying to keep things a bit stable for an open source release. Hope you understand! For now I think people can copy over code from the SED repo. The correction to the GBIFConverter is good to go i think!

benjaminsshoffman:
> Hey I think lets keep these ebird <-> Clements <-> GBIF conversions for a later release of esp-data. i'm trying to keep things a bit stable for an open source release. Hope you understand! For now I think people can copy over code from the SED repo. The correction to the GBIFConverter is good to go i think!

Alright, I've reverted those changes and made a separate branch (`ebird_taxonomy` which does have the changes, so we don't have to redo them later).

I've moved the manual correction dict to the main file, so this should be ready to merge upon your approval

GaganNarula (approved):
lgtm!

PR #257, opened 2026-03-27
Read remote audio segments with ffmpeg

GaganNarula:
We currently allow reading audio by time with `start_time` and `end_time` in `read_audio`. However, this results in actually downloading the entire file first. This utility uses ffmpeg's http range request to only read segments from a large file on a google cloud bucket. Currently only supports gcs, but in principle the approach will also work with s3 buckets.

Why do we need it ? This is relation to PR #255 . The problem we're facing is that some datasets like the spanish-carrion-crows dataset contain very large audio files (> 10 hrs). We actually don't need the entire file just segments from it defined by a selection table. 

In general, we need to be able to read very long audio from buckets quickly without memory overhead and long download times. It'll be useful for PAM data as an example.. There is a way to do this using filesystem byte range reads without using ffmpeg but that approach works only for uncompressed audio like WAV format.


Paulchen-git (approved):
Lgtm

GaganNarula:
Superseded by #310, which combines this with #293 into a unified read_audio.

PR #258, opened 2026-04-01
Add downsample and upsample transforms

david-rx:
Downsample and upsample randomly to a given ratio. Could be either here or NatureLM-specific

## Summary

- **Downsample**: New transform that randomly keeps a fraction of rows. Configured via `fraction` (0, 1] and `seed`. Useful for quick dev iterations or ablation experiments.
- **Upsample**: New transform that repeats every row N times via concatenation. Configured via `factor` (>= 2). Useful for epoch-matching small datasets with larger ones during training.

Both transforms follow the existing esp-data conventions (Pydantic config with `Literal` type discriminator, `from_config` classmethod, `DataBackend` protocol, `register_transform` registration).

## Test plan

- [x] `test_downsample.py`: fraction accuracy, full fraction (1.0), small fraction floor (>= 1 row), manual vs config parity, config validation — all parametrized over pandas and polars backends
- [x] `test_upsample.py`: length correctness, row preservation, manual vs config parity, config validation — all parametrized over pandas and polars backends
- [x] All 16 tests pass locally
- [x] Pre-commit hooks pass (ruff, codespell, trailing whitespace, etc.)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/esp_data/transforms/downsample.py
+++ b/esp_data/transforms/downsample.py
@@ -1,0 +1,71 @@
+"""Transform to randomly keep a fraction of rows."""
+
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+from esp_data.backends.protocol import DataBackend
+
+from . import register_transform
+
+
+class DownsampleConfig(BaseModel):
+    type: Literal["downsample"]
+    fraction: float = Field(gt=0.0, le=1.0)
+    seed: int = 42
+
+
+class Downsample:
+    """Randomly sample a fraction of rows from the dataset.
+
+    This transform keeps a random subset of rows, specified as a fraction of
+    the total.  Useful for quickly reducing dataset size for development or
+    ablation experiments.
+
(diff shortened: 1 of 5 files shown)


esp_data/transforms/downsample.py
```diff
@@ -0,0 +1,71 @@
...
+        backend : DataBackend
+            The backend wrapping the data to downsample.
+
+        Returns
+        -------
+        tuple[DataBackend, dict]
+            The downsampled backend and empty metadata dict.
+        """
+        n = max(1, round(len(backend) * self.fraction))
+        return backend.sample_rows(n=n, seed=self.seed), {}
```
GaganNarula, line 68: if fraction == 1.0, then maybe we should just return the backend ? Not sure if the backends are shuffling even if fraction==1.0 (polars sets shuffle = False by default)

GaganNarula (changes requested):
Can we add these transform docs to docs/transforms.md ?

PR #260, opened 2026-04-01
Add prek, add 'exclude-newer' to safeguard

GaganNarula:


mil-ad (approved):
Looks good to go. We reference installing pre-commit in README.md. 

Maybe let's change in this PR? Approving anyway

PR #261, opened 2026-04-06
update nbm to more permissive license

benjaminsshoffman:
NBM dataset updated their license from ND to NC-SA 🎉 I updated the dataset info.


GaganNarula (approved):
lgtm

PR #267, opened 2026-04-15
Add https endpoint support

GaganNarula:
Public buckets on Cloudflare have `https://...` prefixes. `fsspec` can handle them easily. We can then read from Cloudflare R2 public buckets without needing the R2 secrets which are currently on GCP and OSS users won't have access to that.

## What's here

- `PureHTTPSPath` and `PureHTTPPath` path classes
- `filesystem()` accepts `"http"` and `"https"`, both returning an `HTTPFileSystem`.
- `filesystem_from_path()` routes both new path types to `HTTPFileSystem`
- Both are exported from `alp_data.io` and included in `AnyPathT`.

## Why `http` and `https` are separate path classes

`fsspec` registers a single `HTTPFileSystem` for both schemes — `HTTPFileSystem.protocol` is `("http", "https")`, and `fsspec.filesystem("http") is fsspec.filesystem("https")`. So nothing is duplicated at the filesystem layer.

The *path* classes are a different story, and we keep them separate for two reasons:

**`cloud_prefix` holds a single scheme.** It's validated strictly at construction (`paths.py:113`) and used to rebuild the path string in ~15 places across `PureCloudPath`. One class covering both schemes would mean turning that attribute into a tuple *and* tracking the actual scheme per instance — a real change to the base class for the sake of two schemes.

This also mirrors the existing `PureS3Path` / `PureR2Path` split — two classes sharing behaviour, disambiguated in `anypath()`.

Adding `PureHTTPPath` fixes a real issue: before this, `anypath("http://...")` fell through to `Path("http://...")`, which produced a `LocalFileSystem` and a confusing downstream failure rather than an HTTP read.

## Also in this PR

- Merged `main`, which included the `esp_data` → `alp_data` package rename. The two conflicts were both the same collision — the rename landed on the exact import lines that gained `PureHTTPSPath`.
- `filesystem()`'s return annotation now includes `HTTPFileSystem` (it could already return one).
- The unknown-protocol `ValueError` now lists all supported backends; it still named only `gcs` and `r2`.

## Testing

`tests/io/` is 102 passed (up from 86). New coverage: a `TestPureHTTPPath` class mirroring the HTTPS one, `anypath` dispatch for both schemes, assertions that neither scheme is conflated with or joinable into the other, both protocol strings through `filesystem()`, and the unknown-protocol error.


mil-ad (commented):
These look even less like a file system. At least buckets have the file system hierarchy, even though it's fake, but not sure about HTTP URLs.

Isn't it enough to instead change `filesystem_from_path()` to have some sort of HTTP filesystem?

mil-ad:
but this is fine if you imagine a lot of URL path manipulation

GaganNarula:
> These look even less like a file system. At least buckets have the file system hierarchy, even though it's fake, but not sure about HTTP URLs.
> 
> Isn't it enough to instead change `filesystem_from_path()` to have some sort of HTTP filesystem?

its just that it would make life easier for two reasons:
1. We use the filesystem_from_path and anypath at lot in esp-data, especially to join paths in the datasets when loading audio, so we need to be able to do that with "https://" prefixes or anypath() cant create the path object. 

2. Having the public r2.dev bucket lets experiment for now without relying on the secrets and the S3api and there are a few open datasets that Moritz identified which also have https urls 

``filesystem_from_path()` to have some sort of HTTP filesystem?` i think the code is achieving that in the diff ?

ReadyPlayerEmma:
I thought of another potential approach and wrote a working initial example in PR #275 . I am sure it's very possible I am missing important context, but I figured it was worth exploring to get to know the code base a bit more, and I'd learn from the negative case where I find out I am missing something too.

GaganNarula:
> I thought of another potential approach and wrote a working initial example in PR #275 . I am sure it's very possible I am missing important context, but I figured it was worth exploring to get to know the code base a bit more, and I'd learn from the negative case where I find out I am missing something too.

@ReadyPlayerEmma is there a different solution for this https url issue that you are working on ? Should I close this PR ?

ReadyPlayerEmma:
We should update `docs/io.md` to reflect the changes here. HTTP needs a short note: it is read-only, listing and globbing only work when the server returns an HTML index, so on an R2 public bucket they return an empty list or raise FileNotFoundError, and existence checks issue a GET rather than a HEAD.

We should also document, ideally in the class docstrings so it shows up in the API reference too, that .suffix includes any query string or fragment. Several datasets already branch on the suffix of the split location to choose JSONL versus CSV, so a presigned URL would silently take the CSV path. Documenting it is enough for now since the public mirror never produces query strings.

Otherwise I think this looks good for now.

tests/io/test_http_live.py
```diff
@@ -0,0 +1,267 @@
...
+
+import numpy as np
+import pytest
+from fsspec.implementations.http import HTTPFileSystem
+
+import alp_data.io.read_utils as read_utils
+from alp_data.io import anypath, exists, filesystem_from_path, get_audio_info, read_audio, read_text
+from alp_data.io.paths import PureHTTPPath, PureHTTPSPath
+
+R2_PUBLIC_DIR = "https://pub-ad974658c4664976b92e83cd9e627e83.r2.dev/test"
```
GaganNarula, line 30: added this test bucket for a live read of two (already public) files. ffmpeg range read also included. Can drop if this seems excessive

alp_data/io/file_utils.py
```diff
@@ -16,8 +18,51 @@ def exists(path: str | AnyPathT) -> bool:
...
+    fs = filesystem_from_path(path)
+    path_str = str(path)
+
+    if not isinstance(anypath(path), (PureHTTPSPath, PureHTTPPath)):
+        return fs.exists(path_str)
+
+    try:
+        fs.info(path_str)
+    except FileNotFoundError as e:
+        # `HTTPFileSystem.info` reports every failure as FileNotFoundError, a
```
GaganNarula, line 50: Checking for existence with HTTP(s) is tricky, however, since we support it for the buckets we should *try* to best effort support it for these checkpoints ? my main reason is that a number of tests (and sometimes, workflows using alp-data in training/eval) use the filesystem methods and existence checks, like checking that all available splits of a registered dataset exist. These should also pass on a public r2 mirror imo

alp_data/io/file_utils.py
```diff
@@ -38,5 +83,23 @@ def rm(
...
+
+    Notes
+    -----
+    HTTP(S) paths are rejected before any request is made. `fsspec`'s
+    `HTTPFileSystem` does not implement file removal, so it would raise a bare
+    `NotImplementedError` with no message anyway — and with `recursive=True` it
+    would first issue several HTTP requests while expanding the path.
     """
+    if isinstance(anypath(path), (PureHTTPSPath, PureHTTPPath)):
+        raise NotImplementedError(
```
GaganNarula, line 100: throwing this error early (fsspec will also do it but with a blank error message)

alp_data/io/paths.py
```diff
@@ -751,9 +751,53 @@ class PureR2Path(PureCloudPath):
...
 
+class PureHTTPSPath(PureCloudPath):
+    """HTTPS path, e.g. an object in a Cloudflare R2 public bucket.
+
+    Path manipulation only; use `alp_data.io.filesystem_from_path` to read the
+    object. HTTP(S) endpoints are read-only.
+
+    Notes
+    -----
+    A URL is treated as a plain path, so a `?query` or `#fragment` counts as part
```
GaganNarula, line 762: docstring note about the query strings for signed URLs

ReadyPlayerEmma:
@GaganNarula  One last thing that Claude found, otherwise I think we're good.

The exists() fallback depends on an fsspec feature newer than the declared floor. In file_utils.py the last branch calls fs.exists(path_str, strict=True). I traced strict upstream: it landed in fsspec 2025.12.0. The project floors are gcsfs>=2024.9.0.post1 and s3fs>=2024.9.0, and both pin fsspec to their own version. So a user resolving to an older gcsfs gets an HTTPFileSystem.exists without strict, and the kwarg falls through to aiohttp's session.get, which raises a TypeError. The branch is rare in practice, since it only fires when the chained cause isn't an aiohttp error, such as a bare asyncio timeout. But it's a latent crash on a supported install.

Suggested fix that also tidies a second thing: add fsspec[http]>=2025.12.0 to dependencies. That sets the floor and makes the new direct import aiohttp a declared dependency rather than a transitive one borrowed from gcsfs. Deptry won't complain either way because the repo ignores DEP003, so this is hygiene, not a CI blocker. The lockfile is already at 2026.2.0, so nothing changes for current installs.

GaganNarula:
> @GaganNarula One last thing that Claude found, otherwise I think we're good.
> 
> The exists() fallback depends on an fsspec feature newer than the declared floor. In file_utils.py the last branch calls fs.exists(path_str, strict=True). I traced strict upstream: it landed in fsspec 2025.12.0. The project floors are gcsfs>=2024.9.0.post1 and s3fs>=2024.9.0, and both pin fsspec to their own version. So a user resolving to an older gcsfs gets an HTTPFileSystem.exists without strict, and the kwarg falls through to aiohttp's session.get, which raises a TypeError. The branch is rare in practice, since it only fires when the chained cause isn't an aiohttp error, such as a bare asyncio timeout. But it's a latent crash on a supported install.
> 
> Suggested fix that also tidies a second thing: add fsspec[http]>=2025.12.0 to dependencies. That sets the floor and makes the new direct import aiohttp a declared dependency rather than a transitive one borrowed from gcsfs. Deptry won't complain either way because the repo ignores DEP003, so this is hygiene, not a CI blocker. The lockfile is already at 2026.2.0, so nothing changes for current installs.

Resolved: added fsspec[http]>=2025.12.0, also put the right floor for gcsfs and s3fs (>=2025.12.0) 
@ReadyPlayerEmma any other feedback ?

ReadyPlayerEmma (approved):

PR #268, opened 2026-04-17
Start adding an agent file to esp-data

mil-ad:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,14 @@
+# Code style
+- Use NumPy style for docstrings. If any exceptions are raised make sure that they're documented in the "Raises" section.
+- When you make changes to the logic of a function or class make sure its docstring is still valid. If required, update the docstrong to match the changes.
+- Run `ruff check --fix` to make sure your code is formatter correctly.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,14 @@
...
+- When you make changes to the logic of a function or class make sure its docstring is still valid. If required, update the docstrong to match the changes.
+- Run `ruff check --fix` to make sure your code is formatter correctly.
+- Use single backticks for inline code references in docstrings
+- Don't use conditional `if TYPE_CHECKING`
+- Always add a `__all__` in `__init__.py` with strings of units that are exported
+- Make sure that only objects that are needed outside a module are exposed in the `__init__.py`. Tests don't count as external usage; they should import directly from the submodule if something isn't in `__init__.py`.
+
+# Key Directories
+- `esp_data` - main python library
+
```
GaganNarula, line 12: can we give it a directory map so it knows where datasets, docs, transforms, io, utils etc are located and what those modules do ? or in a second PR ?

mil-ad: let's do in a separate PR

GaganNarula (approved):

PR #272, opened 2026-04-23
python 3.11 + polars[all] breaks on osx arm64_x86 

GaganNarula:
because of `gevent` library which doesnt have a universal.
Replace `polars[all]` with `polars` because we actually dont use everything like delta_lake, connectorx. (for now).


pyproject.toml
```diff
@@ -11,7 +11,8 @@ dependencies = [
     "intervaltree>=3.1.0",
     "librosa>=0.11.0",
     "pandas",
-    "polars[all]>=1.33.1",
+    "polars>=1.33.1",
+    "pyarrow",
```
mil-ad: do we need pyarrow explicitly?

GaganNarula: we could also do `polars[pyarrow]` , that would get rid of the explicit dependency. actually there's a bunch of interoperability packages like:

```
pandas | Convert data to and from pandas dataframes/series.
numpy | Convert data to and from NumPy arrays.
pyarrow | Convert data to and from PyArrow tables/arrays.
pydantic | Convert data from Pydantic models to Polars.
```

mil-ad: yeah `polars[pyarrow]` is clear

mil-ad (approved):

PR #273, opened 2026-04-23
AnimalSoundArchive cleanup 

GaganNarula:
Remove absolute `gcs_path` column from all splits

### DIFF
--- a/esp_data/datasets/animal_sound_archive.py
+++ b/esp_data/datasets/animal_sound_archive.py
@@ -34,7 +34,6 @@ class AnimalSoundArchive(Dataset):
 
     **Audio File Paths:**
         - ``originals_path``: Path to original MP3 audio relative to data root
-        - ``gcs_path``: Full GCS path to original audio
         - ``32khz_path``: Path to pre-resampled 32kHz WAV audio relative to data root
         - ``16khz_path``: Path to pre-resampled 16kHz WAV audio relative to data root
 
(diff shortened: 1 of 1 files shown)


ReadyPlayerEmma (approved):
Verified the v1->v2 data diff empirically on the val split: 2,993 rows in both, 88 cols -> 87 cols, and the only column difference is gcs_path (present in v1, absent in v2). Every other column is row-identical after sorting by tsa_id.

The _v2 files were uploaded alongside the originals rather than overwriting them, so there is backward compatibility.

One question out of curiosity: the docstring example now uses streaming=True. Is that just showing a valid usage, or is it signaling streaming as the new recommended default for this dataset?

GaganNarula:
> Verified the v1->v2 data diff empirically on the val split: 2,993 rows in both, 88 cols -> 87 cols, and the only column difference is gcs_path (present in v1, absent in v2). Every other column is row-identical after sorting by tsa_id.
> 
> The _v2 files were uploaded alongside the originals rather than overwriting them, so there is backward compatibility.
> 

Thanks for doing a manual validation of the split! 

> One question out of curiosity: the docstring example now uses streaming=True. Is that just showing a valid usage, or is it signaling streaming as the new recommended default for this dataset?

That's a valid usage *and* its signaling that streaming=True should be a default in docstrings because it reduces I/O for testing. We don't explicitly recommend this in our docs etc, so thats something to remember todo.

PR #279, opened 2026-05-02
Update README & add license

ReadyPlayerEmma:
Initial draft proposal for the OSS README re-write. For review by the team prior to release.

We'll also have to double check everything works correctly after the R2 bucket is available, since that cannot currently be tested.


README.md
```diff
@@ -1,37 +1,90 @@
-# esp-data
-This repository contains data tools for managing data operations in ESP's AI / science projects.
+# EthoHub
```
GaganNarula: Are we changing the name starting with this README ? I am fine with it @mil-ad and @ReadyPlayerEmma

README.md
```diff
@@ -1,37 +1,90 @@
...
 
-## Help
+EthoHub gives you unified Python access to dozens of bioacoustic datasets, including recordings from birds, marine mammals, primates, insects, anurans, and multi-taxon benchmarks. Every built-in dataset shares a common `Dataset` interface, with streaming, on-the-fly transforms, and consistent loading regardless of source format.
```
GaganNarula: we dont have to mention `Python` i think. We should also mention that it enables large scale ML model training on cross-taxonomic bioacoustic datasets.

README.md
```diff
@@ -1,37 +1,90 @@
...
 
-See [documentation](http://esp/docs/esp-data).
+## Why EthoHub
 
-> [!NOTE]
-> You need to be connected to Tailscale for the link to work.
+Bioacoustic recordings live in many places: Zenodo, OSF, GBIF, institutional repositories. Each dataset arrives with its own format, manifest schema, audio organization, sampling rate, and licensing posture. Researchers wanting to listen across datasets first have to write a custom loader per dataset, then a custom mixer for combining them.
 
-## Development Setup
+EthoHub removes that scaffolding. Every built-in dataset surfaces the same Python interface (`for sample in ds`, `ds[i]`, `len(ds)`), with consistent keys for audio arrays, sample rates, labels, and metadata. Sample-rate harmonization, label derivation, and split-aware loading are wired in. Adding a new dataset means writing one class, not rebuilding plumbing.
```
GaganNarula: i wouldnt add `labels and metadata` in this list because we dont have a consistent schema across datasets. The only thing that's common is "sample_rate" and "audio" keys are always returned

README.md
```diff
@@ -1,37 +1,90 @@
...
+## Highlights
+
+- **Source-flexible loading** — built-in datasets stream from a public Cloudflare R2 bucket; the I/O module can also read from local paths, GCS, S3, and other R2 buckets, with CSV / JSON-Lines / Parquet manifests supported at the backend layer.
+- **Iterate or random-access indexing** — every Dataset supports `for sample in ds` and `ds[i]`.
+- **Streaming mode** — process datasets larger than memory with `streaming=True`.
+- **On-the-fly transformations** — filter rows, select columns, derive labels, deduplicate, balance, upsample long tails, subsample by ratio.
+- **Combine datasets** — `ConcatenatedDataset` merges multiple datasets with configurable column-merge strategies; `ChainedDataset` iterates over them sequentially with streaming support.
+- **Pluggable backends** — pandas or polars under the hood, selectable per-Dataset.
+
+## Quickstart
```
GaganNarula, line 27: i'd suggest moving Quickstart before What's Included

README.md
```diff
@@ -1,37 +1,90 @@
...
+# Indexed access
+sample = beans[0]
+print(sample["audio"].shape)
+
+# Streaming mode (lower memory; len() not available)
+beans_streaming = Beans(split="train", streaming=True)
+for sample in beans_streaming:
+    print(sample["audio"].shape)
+    break
 ```
```
GaganNarula, line 52: We should also add a code snippet for instantiating datasets + a simple transform like `Filter` vi a yaml config. I think we want to push the config-driven aspect of it

README.md
```diff
@@ -1,37 +1,90 @@
...
+- **Multi-taxon benchmarks and aggregators** — BEANS and BeansZero (the canonical bioacoustic and zero-shot benchmarks), AnimalSpeak (1M+ audio–caption pairs), AnimalSoundArchive, iNaturalist audio, AudioSet / AudioSetStrong, and Voxaboxen (overlapping vocalization detection).
+
+Most datasets are openly licensed (CC-BY, CC-BY-NC, CC0, public domain). License and source metadata are available per-dataset via `Dataset.info` — provenance matters.
+
+## Highlights
+
+- **Source-flexible loading** — built-in datasets stream from a public Cloudflare R2 bucket; the I/O module can also read from local paths, GCS, S3, and other R2 buckets, with CSV / JSON-Lines / Parquet manifests supported at the backend layer.
+- **Iterate or random-access indexing** — every Dataset supports `for sample in ds` and `ds[i]`.
+- **Streaming mode** — process datasets larger than memory with `streaming=True`.
+- **On-the-fly transformations** — filter rows, select columns, derive labels, deduplicate, balance, upsample long tails, subsample by ratio.
```
GaganNarula: transforms are actually unfortunately not lazy right now.. so the "on-the-fly" terminology can be misleading. Transforms are run after loading metadata into memory (hence our current issue with running transforms over a very large set of datasets chained together)

GaganNarula (commented):
looks like a good start already! :) left some comments

README.md
GaganNarula: looks like EthoHub might not be the new name

mil-ad (approved):

PR #280, opened 2026-05-03
INaturalist and Beans fixes

GaganNarula:
- Convert inaturalist m4a files to wav (our `read_audio` method cant handle m4a right now, but we might want to support it going forward because there were more than 80,000 m4a files in inaturalist alone, or we convert during data ingestion).
-  Also needed to convert broken inaturalist mp3 files -> wav
- fix split audio paths in inat
- fix esc50 paths in train/val/test splits of beans (this should be the last change)
- Some scripts used for clean up and audio conversion


mil-ad (approved):

ReadyPlayerEmma:
**Tracking lossy-source provenance for downstream consumers**

Worth thinking about how we surface the fact that these WAVs originated from a lossy-compressed source, since that information matters for anyone training on this data and isn't recoverable once the `.m4a` extension is replaced with `.wav`.

Two concerns specifically:

1. **Perceptual codec artifacts.** AAC (like MP3) is calibrated for human psychoacoustics — it lowpasses high frequencies, allocates bits based on what humans can hear, and discards "inaudible" information. For animal vocalization models this is a potential confound, since the model may care about exactly the signal content the codec considered throwaway. A WAV converted from M4A is not equivalent to a WAV recorded directly, even though they have the same extension and sample format.

2. **Encoder delay / priming samples.** AAC encoders prepend priming samples (2112 for Apple's encoder, 1024 for ffmpeg's native, different again for HE-AAC), and decoders vary in whether they strip these on decode. This means the WAV produced here may not be sample-aligned with a WAV produced by a different decoder from the same M4A. If anything downstream references timestamps from the original M4A (e.g., iNaturalist within-file annotations, if those exist), this matters; if labels are file-level only, it doesn't.

Some options worth considering, in increasing order of effort:

- Add `source_format` (and ideally `decoder` + version) columns to the split CSVs alongside `originals_path`, so consumers can filter or weight by source format and reproduce the conversion if needed.
- Keep the original `originals_path` pointing at the `.m4a` and add a separate `wav_path` column, rather than rewriting in place. Preserves the pointer to source data and makes the transform visible.
- Log the librosa/ffmpeg versions in the conversion report (already partially there) and pin them in the script so re-runs are deterministic.
- Document at the dataset level (README or dataset card) that any audio with a lossy source has been through a perceptual codec, so this travels with the data.

Probably don't need all of these, but the in-place rewrite of `originals_path` from `.m4a` to `.wav` does lose information that I think downstream consumers will want. At minimum, preserving a pointer to the original source seems worth doing.

GaganNarula:
> **Tracking lossy-source provenance for downstream consumers**
> 
> Worth thinking about how we surface the fact that these WAVs originated from a lossy-compressed source, since that information matters for anyone training on this data and isn't recoverable once the `.m4a` extension is replaced with `.wav`.
> 
> Two concerns specifically:
> 
> 1. **Perceptual codec artifacts.** AAC (like MP3) is calibrated for human psychoacoustics — it lowpasses high frequencies, allocates bits based on what humans can hear, and discards "inaudible" information. For animal vocalization models this is a potential confound, since the model may care about exactly the signal content the codec considered throwaway. A WAV converted from M4A is not equivalent to a WAV recorded directly, even though they have the same extension and sample format.

I agree with this in principle, but currently we're trying to just avoid any breaking iteration over the dataset. I think David ignored this when he was resampling to 16 and 32Khz. Removing the files altogether because they are fit for human psychoacoustics would be a drastic (and who knows, if the recording contains birdsong / insects etc then these are very much in the human spectrum). 

> 2. **Encoder delay / priming samples.** AAC encoders prepend priming samples (2112 for Apple's encoder, 1024 for ffmpeg's native, different again for HE-AAC), and decoders vary in whether they strip these on decode. This means the WAV produced here may not be sample-aligned with a WAV produced by a different decoder from the same M4A. If anything downstream references timestamps from the original M4A (e.g., iNaturalist within-file annotations, if those exist), this matters; if labels are file-level only, it doesn't.
> 
Agree that this is a potential issue, but I dont think we will be decoding the m4a with a different decoder in this library. Also, there are no downstream references to timestamps in these audio files. 

> Some options worth considering, in increasing order of effort:
> 
> * Add `source_format` (and ideally `decoder` + version) columns to the split CSVs alongside `originals_path`, so consumers can filter or weight by source format and reproduce the conversion if needed.

Would be a nice to have but right now dont have time we can make an issue for this (going from _v2 -> _v3)

> * Keep the original `originals_path` pointing at the `.m4a` and add a separate `wav_path` column, rather than rewriting in place. Preserves the pointer to source data and makes the transform visible.

But then we have to fiddle with path resolution code

> * Log the librosa/ffmpeg versions in the conversion report (already partially there) and pin them in the script so re-runs are deterministic.
Good idea, will do

> * Document at the dataset level (README or dataset card) that any audio with a lossy source has been through a perceptual codec, so this travels with the data.
> 
Will add to doc string

> Probably don't need all of these, but the in-place rewrite of `originals_path` from `.m4a` to `.wav` does lose information that I think downstream consumers will want. At minimum, preserving a pointer to the original source seems worth doing.

PR #284, opened 2026-05-05
Replace bucket prefix with env var

GaganNarula:
introduces `ESP_DATA_HOME` as an environment variable that should be changed when the location `cloud-prefix://bucket-name` is changed, and `DATA_HOME` as a constant in `esp_data.io.paths` with a default value (Should we raise an error instead ?). Env var is accessed on import. 

+ Removed `data_root` param from doc string tests

Another question: should we deprecate `data_root` ?


esp_data/datasets/inaturalist.py
mil-ad: doesn't this need updating too?

mil-ad (approved):

PR #285, opened 2026-05-05
Add a default chunk size for iteration, same as pandas

GaganNarula:
the default is somehow not specified
https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.collect_batches.html 
but in testing it seems to reading way too many rows in one go. We have to keep an eye on this api because polars marked it unstable.

### DIFF
--- a/esp_data/backends/polars_backend.py
+++ b/esp_data/backends/polars_backend.py
@@ -30,7 +30,10 @@ class PolarsBackend(DataBackend):
         Whether the backend is in streaming mode (LazyFrame)
     streaming_chunk_size: int
         Number of rows per batch when iterating in streaming mode (default: 1000)
-
+        1000 is a good number because its high enough to reduce I/O and any higher
+        doesnt help because the main latency source in Dataset __getitem__ calls are in 
+        loading audio anyway.
+        
     Examples
     --------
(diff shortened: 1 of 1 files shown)


esp_data/backends/polars_backend.py
```diff
@@ -28,6 +28,8 @@ class PolarsBackend(DataBackend):
         The polars DataFrame or LazyFrame to wrap
     streaming : bool
         Whether the backend is in streaming mode (LazyFrame)
+    streaming_chunk_size: int
+        Number of rows per batch when iterating in streaming mode (default: 1000)
```
mil-ad, line 32: maybe add a comment on why 1000 was chosen

mil-ad (changes requested):

esp_data/backends/polars_backend.py
GaganNarula: done

mil-ad: pre-commit is failing

mil-ad (changes requested):
typo found by pre-commit

mil-ad (approved):

PR #290, opened 2026-05-14
Add spanish carrion crows vox

mcusi:
A different version of the Spanish Carrion Crow datasets that only includes focal vocalizations (i.e., vocalizations made by an adult bird wearing a biologger), in time periods where we have more than one crow tagged. 

We have pre-computed audio clips for each focal vocalization, making it simpler than dataset #255 

Note that the full dataset has more focal vocalizations than this one, but we are subsetting to sections where there is synchronization between >1 bird in the same year/territory.

### DIFF
--- a/esp_data/datasets/spanish_carrion_crows.py
+++ b/esp_data/datasets/spanish_carrion_crows.py
@@ -6,6 +6,7 @@ from io import StringIO
 from typing import Any, Iterator
 
 import librosa
+import numpy as np
 import pandas as pd
 
 from esp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
@@ -38,7 +39,7 @@ class SpanishCarrionCrows(Dataset):
         - Synchronized UTC timestamps across biologgers.
         - Annotations of file quality
           e.g., some biologgers prematurely detached from the bird and recorded
-          background sounds only
+          background sounds only. Therefore some selection tables are empty.
         - Call types
     If any additional data are needed, please contact Maddie.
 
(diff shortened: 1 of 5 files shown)


esp_data/datasets/spanish_carrion_crows_vox.py
```diff
@@ -0,0 +1,329 @@
...
+
+    @property
+    def columns(self) -> list[str]:
+        return list(self._data.columns)
+
+    @property
+    def available_splits(self) -> list[str]:
+        return [
+            "all",
+            1,
```
GaganNarula, line 237: These splits (1,2, .. ) dont have any paths defined ?

esp_data/datasets/spanish_carrion_crows_vox.py
```diff
@@ -0,0 +1,329 @@
...
+        self.fallback_to_noisy = fallback_to_noisy
+        self.padding_sec = padding_sec
+
+        self.data_root = anypath(data_root) if data_root is not None else anypath("gs://")
+
+        self._load()
+
+    @property
+    def columns(self) -> list[str]:
+        return list(self._data.columns)
```
GaganNarula: this can just be `return self._data.columns`, already correct type

esp_data/datasets/spanish_carrion_crows_vox.py
```diff
@@ -0,0 +1,329 @@
...
+        sample_rate : int | None
+            If set, audio is resampled to this rate. Defaults to 16000.
+        data_root : str | AnyPathT | None
+            Root directory prepended to each audio path in the CSV. Defaults to
+            "gs://" (audio paths in the CSV start with the bucket name).
+        backend : BackendType
+            Backend to use ("pandas" or "polars"). Defaults to "polars".
+        streaming : bool
+            Whether to use streaming mode. Defaults to False.
+        denoised : bool
```
GaganNarula, line 204: for datasets with extra parameters like "denoised" , "fallback_to_noisy" you need a separate config, see audioset.py for `AudiosetConfig` for instance

esp_data/datasets/spanish_carrion_crows_vox.py
```diff
@@ -0,0 +1,329 @@
...
+    about usage in work to be published.
+
+    """
+
+    info = DatasetInfo(
+        name="spanish_carrion_crows_vox",
+        owner="maddie",
+        split_paths={
+            # TODO: update path once finalized in GCS
+            "all": "gs://esp-ml-datasets/spanish-carrion-crows-vox/conversational_preprocessed.csv",
```
GaganNarula: is this file already in `esp-ml-datasets` ? if not it should be in esp-data-ingestion otherwise the tests wont run ?

esp_data/datasets/spanish_carrion_crows_vox.py
```diff
@@ -0,0 +1,329 @@
...
+    def _load(self) -> None:
+        location = self.info.split_paths["all"]
+        self._data = self._backend_class.from_csv(
+            location,
+            streaming=self._streaming,
+            keep_default_na=False,
+            na_values=[""],
+        )
+        if self.split != "all":
+            self._data = self._data.filter_isin("overlap_window_id", [float(self.split)])
```
GaganNarula, line 338: oh i see whats happening.. but this is a transform of a split, so i think its better to let the user do this in their code rather than add as a fake "split" which gets generated on _load 

Also keeps are api a bit more stable

esp_data/datasets/spanish_carrion_crows_vox.py
```diff
@@ -0,0 +1,329 @@
...
+        if self._data is None:
+            raise RuntimeError("No split has been loaded yet. Call _load() first.")
+        if self._streaming:
+            raise NotImplementedError(
+                "Length is not available in streaming mode. Iterate over the dataset instead."
+            )
+        return len(self._data)
+
+    def _process(self, row: dict[str, Any]) -> dict[str, Any]:
+        start_sec = max(0.0, float(row["derived.centered_focal.long.start_sec"]) - self.padding_sec)
```
GaganNarula, line 355: are these files also very long (e.g. more than 30 min) .. our current `read_audio` actually doesn't do a kind of "streaming cut" of a long audio in a bucket, which means we'll hit OOM issues if the files are long (related to PR #257)

esp_data/datasets/spanish_carrion_crows_vox.py
```diff
@@ -0,0 +1,329 @@
...
+        end_sec = float(row["derived.centered_focal.long.end_sec"]) + self.padding_sec
+
+        if self.denoised:
+            denoised_success = row["derived.denoised_focal.success"]
+            if isinstance(denoised_success, str):
+                denoised_success = denoised_success == "True"
```
GaganNarula: seems like a check for a dtype that should really be a `bool` in the underlying data (all rows should be bool not a mixed type)

GaganNarula (changes requested):
changes needed

esp_data/datasets/spanish_carrion_crows_vox.py
mcusi: They refer to an index within the CSV file. Is that okay?

esp_data/datasets/spanish_carrion_crows_vox.py
mcusi: yes it is

esp_data/datasets/spanish_carrion_crows_vox.py
mcusi: No, they are like five seconds

esp_data/datasets/spanish_carrion_crows_vox.py
mcusi: I think maybe I don't understand the transforms well enough. Could you point me to an example of how the desired usage should be if people want the vox where overlap_window_id = split_idx?

GaganNarula: well basically a user would just have to use the FilterTransform (already a part of alp_data.transforms) . its quite easy to use, the idea would be that there'd be only one split ('all') and by declaring a `filter_cfg = FilterConfig(mode="include", property="overlap_window_id", values=[1,2, .. <any other needed splits>])` and then using `new_dataset, metadata = my_spanish_carrion_crows_ds.apply_transformations([filter_cfg])` the user receives the required splits ... In the current implementation actually only one split can be loaded. Everything would work from yaml as well, defining the same transformation in the yaml. The docs have all the information necessary, i think Claude would manage to separate that out easily.

The *other* option is to have all 288 splits as separate files, but i think thats a bit ugly..

esp_data/datasets/spanish_carrion_crows_vox.py
GaganNarula: i think this would work now

esp_data/datasets/spanish_carrion_crows.py
```diff
@@ -0,0 +1,299 @@
...
+            "focal",
+            "not focal",
+            "crowchicks",
+            "cuckoo",
+        ]  # for Annotation column
+
+        self._load()
+
+        if data_root is None:
+            self.data_root = anypath("gs://")
```
GaganNarula, line 112: the `data_root` parameter is used to change the location of the dataset.. from the looks of this default ("gs://") it seems the bucket name is embedded into the `audio_fp` column ?  thats ok for now, but if we make the dataset public we would need to revisit it

GaganNarula (changes requested):
basically one change needed

GaganNarula:
oh and it looks like you have to resolve conflicts first (and then probably merge main into this branch)

Issue #287, opened 2026-05-11
Remove unsupported audio format check


GaganNarula:
In `esp_data.io.read_utils` we have `_AUDIO_FORMATS = (".wav", ".flac", ".ogg", ".mp3")` . During data migration to public bucket I noticed that some audios in xeno-canto dont even have a file extension. These files do have valid content (manual check). I propose removing ValueError thrown by `read_audio` that checks against are allowed format list

PR #293, opened 2026-05-21
Fix for issue #287 and adds librosa fallback

GaganNarula:
- Insert `librosa.load` between `sf.read` and tmpfile decode in `_read_audio_from_file` and `read_audio_by_time`. Reason is that some files
- Remove checking for extensions. Some audio files in our datasets are actually fine, they contain reasonable audio content but have no extension 
- Drop noisy fallback warnings in the audio read path.


GaganNarula:
Superseded by #310, which combines this with #257 into a unified read_audio.

PR #295, opened 2026-05-22
Tar exporter rebased

Paulchen-git:
Clean #234.
Rebased on on #212 with a clean commit stack


Paulchen-git:
Doesn't work with datasets that have selection table because dataframe are not json serializable for now (like selection tables).

PR #297, opened 2026-05-29
feat: add StreamingBackend protocol for streaming-only backends. (Webdataset)

Paulchen-git:
Defines StreamingBackend protocol for backends that support sequential iteration only (e.g., tar-based formats), alongside the existing DataBackend. Updates backend registry type hints to accept both protocol types.

### DIFF
--- a/alp_data/backends/protocol.py
+++ b/alp_data/backends/protocol.py
@@ -3,11 +3,196 @@
 This module defines the interface that all backend implementations must follow.
 It enables support for multiple data handling libraries (pandas, polars, etc.) through
 a common abstraction layer.
+
+Two protocols are defined:
+- StreamingBackend: For streaming-only data formats (e.g., WebDataset/tar files)
+  that only support iteration, not random access.
+- DataBackend: For in-memory data formats (e.g., pandas, polars DataFrames)
+  that support both random access and iteration.
 """
 
 from __future__ import annotations
 
 from typing import Any, Callable, Iterator, Literal, Protocol, overload
+
+
+class StreamingBackend(Protocol):
+    """Protocol defining the interface for streaming-only data backends.
+
+    This protocol is designed for data formats that only support sequential
+    iteration, such as WebDataset (tar files). These backends cannot support:
+    - Random access (__getitem__)
+    - Known length (__len__)
+    - Operations requiring full data scan (drop_duplicates, get_unique, histogram)
+    - Sampling operations (subsample_by_column, upsample_by_column, sample_rows)
+
+    Implementations should apply filtering and transformations lazily during iteration.
+    """
+
+    @property
+    def is_streaming(self) -> bool:
+        """Check if backend is in streaming mode.
+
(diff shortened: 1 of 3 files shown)


alp_data/backends/protocol.py
```diff
@@ -95,6 +280,32 @@ def from_parquet(cls, path: str, *, streaming: bool = False, **kwargs: Any) -> "
         """
         ...
 
+    @classmethod
+    def from_path(cls, path: str, *, streaming: bool = False, **kwargs: Any) -> "DataBackend":
```
GaganNarula, line 284: is `from_path` needed in DataBackend? what would it return ?

Paulchen-git: The from_path method of a protocol is redirecting to from_json, from_csv, from_parquet for polars and pandas backend and it is just loading the tar files for webdataset

esp_data/backends/__init__.py
```diff
@@ -2,17 +2,27 @@
...
+    # Protocols
     "DataBackend",
+    "StreamingBackend",
+    # DataBackend implementations
     "PandasBackend",
     "PolarsBackend",
+    # Registry
     "BackendType",
     "get_backend",
 ]
```
mil-ad, line 25: don't think we need comments here. doesn't add much

esp_data/backends/__init__.py
```diff
@@ -2,17 +2,27 @@
...
+  that only support iteration, not random access.
+- DataBackend: For in-memory data formats (e.g., pandas, polars DataFrames)
+  that support both random access and iteration.
 """
 
 from .backends import BackendType, get_backend
 from .pandas_backend import PandasBackend
 from .polars_backend import PolarsBackend
-from .protocol import DataBackend
+from .protocol import DataBackend, StreamingBackend
```
mil-ad: maybe `StreamingDataBackend` is a better name? Wdyt?

mil-ad (commented):
Left small comments

esp_data/backends/__init__.py
Paulchen-git: Oh yeah, sounds good, I think it was already something suggest in the previous PR. I'll rename it

mil-ad (approved):
LGTM to me if @GaganNarula is happy too

alp_data/backends/protocol.py
```diff
@@ -3,13 +3,198 @@
...
+            Value to use for unmapped keys, by default None
+
+        Returns
+        -------
+        StreamingDataBackend
+            Self with mapping configured (applied during iteration)
+        """
+        ...
+
+    def apply_fn(
```
GaganNarula, line 178: we dont actually do anything with `apply_fn` as of now (also exists in DataBackend) , but thats something we can take advantage of with a `LazySampleTransform` ... just wanted to make a mental note

GaganNarula (approved):
yes looks good!

PR #302, opened 2026-06-02
Publish esp-data on public PyPi

Paulchen-git:


.github/workflows/publish_public.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Package to PyPI
+
+on:    # yamllint disable-line rule:truthy
+  workflow_dispatch:
+  workflow_call:
```
mil-ad: what's `workflow_call`?

mil-ad (commented):
one tiny comment

.github/workflows/publish_public.yml
Paulchen-git: Makes workflow reusable — other workflows can invoke it as a job step via uses: ./.github/workflows/publish_public.yml. 

Probably not needed here, I'll remove it, I've just copy-pasted it from avex workflow

.github/workflows/publish_public.yml
```diff
@@ -0,0 +1,36 @@
...
+          python-version-file: ".python-version"
+
+      - name: Install the project
+        run: uv sync --all-extras --dev
+
+      - name: Build the package
+        run: uv build
+
+      - name: Publish to PyPI
+        run: uv publish --publish-url https://upload.pypi.org/legacy/ --trusted-publishing always --keyring-provider disabled
```
GaganNarula, line 35: dont we need an account and a key to do this ? i didnt set this up for avex though so i'm probably wrong

Paulchen-git: with `--trusted-publishing` and `permissions: id-token: write`, it looks like we don't need any key but you would have to set this repo and this workflow as  Trusted Publishers in the PyPi settings of the account that owns the package

mil-ad: yeah we do it the other way around, i.e. on pypi add this repo as the publisher

GaganNarula (approved):
lgtm!

PR #305, opened 2026-06-10
Public release v1.10.0

GaganNarula:
also removes the private publish workflow


Paulchen-git (approved):
After merging this PR, does this mean that to use alp-data we can just do `pip install alp-data`??

mil-ad (approved):

PR #308, opened 2026-06-16
Add VFPA SRKW standard split to DCLDE2026

nkundiushuti:
## Summary
  - Adds `vfpa_srkw_standard` split to `DCLDE2026`, used in "what matters" 
  - Points the split to `gs://foundation-model-data/data/dclde2026/splits/vfpa_srkw_standard.csv` but it can be copied in the standard data path
  - Uses `_VFPA_CSV_ROOT` to keep the DCLDE dataset code changes minimal
  - Keeps DCLDE audio resolution rooted at the existing DCLDE data location

  ## Validation
  - Recreated the split from the old `representation-learning` VFPA custom dataset semantics:
    - `JASCO_VFPA` + `JASCO_VFPA_ONC`
    - `SRKW`
    - 27 standard S-call labels
    - 1,336 event windows
  - Compared against the old `VFPAKillerWhales` loader:
    - all 1,336 `filename/start/end/call_type` signatures match
    - sampled audio outputs have matching lengths and near-identical waveforms
  - Ran:
    - `uv run ruff check --fix alp_data/datasets/dclde2026.py tests/unittests/datasets_tests/test_dclde2026.py`
    - `uv run pytest -q tests/unittests/datasets_tests/test_dclde2026.py -q`

### DIFF
--- a/alp_data/datasets/dclde2026.py
+++ b/alp_data/datasets/dclde2026.py
@@ -62,7 +62,6 @@ PROVENANCE_COLUMNS = [
 ]
 
 _DCLDE_DATA_ROOT = f"{DATA_HOME}/dclde2026/v0.1.0/raw/2026/dclde_2026_killer_whales"
-_VFPA_CSV_ROOT = "gs://foundation-model-data/data/dclde2026/splits"
 
 
 @register_dataset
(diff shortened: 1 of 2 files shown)


GaganNarula (changes requested):
birdset commits are here as well ?

nkundiushuti:
> birdset commits are here as well ?

yep, sorry, it was a mistake. I excluded those

GaganNarula:
ok i've copied the split csv to both target buckets. Can you remove `_VFPA_CSV_ROOT` and try again ?

nkundiushuti:
should we merge ?

GaganNarula (approved):
lgtm!

PR #310, opened 2026-06-28
read_audio improvements with ffmpeg GCS segment streaming

### LINKED ISSUE #287: Remove unsupported audio format check

GaganNarula:
Combine #293 + #257 into a unified `read_audio`

`read_audio` is now our only public audio reader which dispatches data over private helpers:

- **Full-file reads** use the `soundfile tries first → librosa for m4a files → tmpfile approach for mp3` decode chain, with extension gating removed (= we try reading any audio format)(#293).
- **Time-range reads on `gs://`** stream just the segment via ffmpeg HTTP range requests (no full download), falling back to a download-based read — warning once per distinct cause — when ffmpeg/credentials are unavailable (#257).

And bug fixes:
- `read_audio_by_time` turned into a private helper (was not in `__all__`).
- ffmpeg output normalized to the soundfile shape convention (`(frames,)` / `(frames, channels)`).
- GCS token (needed for time-range ffmpeg reads) cached at module level (refreshed only when invalid).
- `anonymous` flag added for public objects. Basically, `anonymous=False` means try and fetch a valid gcs token for that session (only authenticated users). However, we keep `anonymous=None` by default which allows a non-authenticated (public) user to read from our public datasets using the ffmpeg time-range reads.
- Clean up: unused `frames`/`start` params dropped from `read_audio`.

CURRENTLY ONLY SUPPORT GCS bucket reads with ffmpeg! TODO: s3/r2 support

Closes #287.

This PR supersedes and replaces #293 and #257, which are being closed.


alp_data/io/read_utils.py
ReadyPlayerEmma, line 1: An inverted time range silently returns the wrong audio instead of raising.

read_audio(path, start_time=5.0, end_time=1.0) yields frames_to_read = -64000; the if frames_to_read > 0 clamp skips negatives, and sf.read(frames=<negative>) treats it as "read to end" — I confirmed this against libsndfile locally.

So the caller gets audio from 5.0s→EOF with no error. On the ffmpeg path the same input builds -t -4.0, which makes ffmpeg fail and fall back to this same wrong result (plus a misleading "ffmpeg decode failed" warning). A one-line if end_time is not None and end_time <= start_time: raise ValueError in read_audio fixes both paths. Silent wrong data in a research pipeline is the worst failure mode, so this is my top ask.

alp_data/io/read_utils.py
```diff
@@ -2,20 +2,271 @@
...
+        Short, stable description of why the ffmpeg path was unavailable.
+    """
+    logger.warning(
+        "ffmpeg segment read unavailable (%s); falling back to full download. "
+        "This is slower for large remote files.",
+        cause,
+    )
+
+
+def _gcs_path_to_url(file_path: str | AnyPathT) -> str:
```
ReadyPlayerEmma, line 67: _gcs_path_to_url doesn't percent-encode the object name. Bioacoustics object names with spaces, #, %, or ? produce a malformed URL. # and ? are the worst cases — the path gets truncated into a request for a different object → 404 → silent fallback to full download, with a confusing "ffprobe failed" warning. Fix is urllib.parse.quote(path_str, safe="/") before building the URL.

alp_data/io/read_utils.py
```diff
@@ -176,52 +431,152 @@ def _read_audio_from_file(
...
+            # Validate input sample rate if provided
+            if input_sr is not None and input_sr != file_sr:
+                logger.warning(
+                    f"Input sample rate {input_sr} doesn't match file sample rate {file_sr}"
+                )
+
+            # Convert time to frame indices
+            start_frame = int(start_time * file_sr)
+
+            if end_time is not None:
```
ReadyPlayerEmma, line 537: Maybe add `if end_time <= start_time: raise ValueError` in here?

alp_data/io/read_utils.py
```diff
@@ -2,20 +2,271 @@
...
+        "-select_streams",
+        "a:0",
+        "-show_entries",
+        "stream=sample_rate,channels",
+        "-of",
+        "csv=p=0",
+        gcs_url,
+    ]
+    try:
+        probe_result = subprocess.run(probe_cmd, check=True, capture_output=True, text=True)
```
ReadyPlayerEmma, line 199: ffmpeg applies no network read timeout by default (tcp timeout defaults to -1, reconnect* all default off — verified on ffmpeg 8.1.2), so a connection that stalls silently mid-read (dropped NAT/firewall state, node network blip) blocks the read forever — reproduced locally: default ffprobe hangs indefinitely against a stalled server, while -rw_timeout 5000000 exits cleanly in 5s. Rare per-read, but across millions of dataloader reads on a long training run it's a matter of time, and one wedged worker stalls the whole job. Suggest adding -rw_timeout (e.g. 30s = 30000000) to both the ffprobe and ffmpeg commands — a stall detector rather than a wall-clock cap, so slow-but-progressing reads are unaffected, and failures flow into the existing FFmpegSegmentError → warn-once → download-fallback path with no new code.

I have personally seen this issue with ffmpeg for a security feed use-case I built a few years back, so it can practically bite.

alp_data/io/read_utils.py
```diff
@@ -2,20 +2,271 @@
...
+        logger.warning(f"Input sample rate {input_sr} doesn't match file sample rate {sample_rate}")
+
+    command = ["ffmpeg", *headers_args, "-ss", str(start_time), "-i", gcs_url]
+    if end_time is not None:
+        command += ["-t", str(end_time - start_time)]
+    # Output raw PCM float32 to stdout, preserving native channel layout.
+    command += ["-f", "f32le", "-acodec", "pcm_f32le", "pipe:1"]
+
+    try:
+        result = subprocess.run(command, check=True, capture_output=True)
```
ReadyPlayerEmma, line 240: Same timeout issue here potentially.

ReadyPlayerEmma (changes requested):
Pinged you on Slack about one other issue as well.

alp_data/io/read_utils.py
GaganNarula: great catch !

alp_data/io/read_utils.py
GaganNarula: makes sense!

ReadyPlayerEmma (approved):

Issue #289, opened 2026-05-13
`dataset_from_config` doesn't use correct config class when chaining


GaganNarula:
We have a couple of datasets (audioset, voxaboxen) which have their own config, namely `AudioSetConfig` and `VoxaboxenConfig` inheriting from `DatasetConfig`. We have these configs because these datasets need more params than the default defined in `DatasetConfig`, for e.g. "version" param in AudioSet. However, when we call `dataset_from_config` with a `chain` or `concat` key in yaml, these particular configs are not used. The build process doesnt fail because `DatasetConfig` has `extra="allow"` in its config dict.

We should ideally: 

1. Set `extra="forbird"`
2. Make sure these specific configs are used when chaining or concatenating

PR #312, opened 2026-07-02
Add v0.2.0 (new dumps) to xeno-canto and iNaturalist via a version parameter

david-rx:
## What

Adds an audioset-style `version` parameter to the Xeno-canto and iNaturalist datasets so the new dumps are available **without dropping access to the existing data**.

- **`0.1.0`** (default) — the existing splits (`*_20260203_v2` / `*_20260201_v3`). Behaviour is unchanged for existing callers.
- **`0.2.0`** — the new Jun 2026 dumps (Xeno-canto `20260622`, iNaturalist `20260616`), filtered and normalized (see below).

Implemented the same way as `AudioSet`: a `VERSIONS` registry, a registered `XenoCantoConfig` / `INaturalistConfig` carrying `version`, and version selection in `__init__` / `from_config`. Both versions keep their metadata under the `v0.1.0/raw/` path, so `data_root` is shared and **no `_process` changes are needed**.

```python
XenoCanto(split="all")                     # -> 0.1.0 (existing data)
XenoCanto(split="all", version="0.2.0")    # -> new dump
```

Docs also gain two Xeno-canto metadata fields present in the new dump: `quality` (A–E / "no score") and `playback_used` (yes/no/unknown).

## How 0.2.0 was prepared (offline, data-only)

The `0.2.0` split CSVs are filtered/normalized copies of the raw dumps (`*_v1.csv`):
- rows missing a 32 kHz path are dropped
- rows that did not link to the GBIF backbone (`gbif_link_ok == False`) are dropped — for Xeno-canto this removes ~26.5k "Mystery"/soundscape/synonym recordings that carry no taxonomic class; iNaturalist has none
- iNaturalist's absolute, mixed-bucket (`esp-data-ingestion` / `esp-ml-datasets`) 32 kHz/16 kHz paths are normalized to paths relative to the raw root, so the loader resolves them against a single `data_root` with no read-time path handling
- Xeno-canto's `playback_used` column (added upstream after this snapshot was cut) is joined in by `xc_id` from the undated `all.csv` — 100% coverage of the snapshot's rows

They are staged as `*_v1.csv` in `esp-data-ingestion` and need to be copied into `DATA_HOME`.

## Copy job (only needed for v0.2.0)

`0.1.0` already resolves against data present in `DATA_HOME`, so the default path works today. `0.2.0` needs the following into `DATA_HOME` (`esp-data-274503`):

**CSVs:**
```
gsutil -m cp "gs://esp-data-ingestion/xeno-canto/v0.1.0/raw/*_20260622_v1.csv"  gs://esp-data-274503/xeno-canto/v0.1.0/raw/
gsutil -m cp "gs://esp-data-ingestion/inaturalist/v0.1.0/raw/*_20260616_v1.csv" gs://esp-data-274503/inaturalist/v0.1.0/raw/
```

**Audio (raw / 16k / 32k):**
```
# xeno-canto
gsutil -m rsync -r gs://esp-data-ingestion/xeno-canto/v0.1.0/raw/audio/     gs://esp-data-274503/xeno-canto/v0.1.0/raw/audio/
gsutil -m rsync -r gs://esp-data-ingestion/xeno-canto/v0.1.0/raw/audio_16k/ gs://esp-data-274503/xeno-canto/v0.1.0/raw/audio_16k/
gsutil -m rsync -r gs://esp-data-ingestion/xeno-canto/v0.1.0/raw/audio_32k/ gs://esp-data-274503/xeno-canto/v0.1.0/raw/audio_32k/

# iNaturalist (note: older 32k audio lives under audio_32khz/, not audio_32k/)
gsutil -m rsync -r gs://esp-data-ingestion/inaturalist/v0.1.0/raw/audio/     gs://esp-data-274503/inaturalist/v0.1.0/raw/audio/
gsutil -m rsync -r gs://esp-data-ingestion/inaturalist/v0.1.0/raw/audio_16k/ gs://esp-data-274503/inaturalist/v0.1.0/raw/audio_16k/
gsutil -m rsync -r gs://esp-data-ingestion/inaturalist/v0.1.0/raw/audio_32k/ gs://esp-data-274503/inaturalist/v0.1.0/raw/audio_32k/
gsutil -m rsync -r gs://esp-ml-datasets/inaturalist/v0.1.0/raw/audio_32khz/  gs://esp-data-274503/inaturalist/v0.1.0/raw/audio_32khz/
```

## Delta 0.1.0 -> 0.2.0 (`all` split, after filters)

**Xeno-canto:** 596,526 -> 744,242 rows (+24.8%); 10,799 -> 11,748 species. New `playback_used` breakdown on the 0.2.0 `all` split: no 628,936 / unknown 76,906 / yes 38,360.

| class | rows (old->new) | species (old->new) |
|---|---|---|
| Aves (birds) | 588,141 -> 700,638 | 9,733 -> 9,891 (+158) |
| Insecta | 5,035 -> 33,792 | 423 -> 1,008 (+585) |
| Amphibia | 1,836 -> 3,388 | 508 -> 679 (+171) |
| Mammalia | 1,514 -> 6,424 | 135 -> 170 (+35) |

**iNaturalist:** 709,217 -> 785,629 rows (+10.8%); 9,168 -> 9,465 species (birds +161, amphibians +37, insects +50, mammals +30).

### DIFF
--- a/alp_data/datasets/inaturalist.py
+++ b/alp_data/datasets/inaturalist.py
@@ -7,7 +7,49 @@ import numpy as np
 
 from alp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
 from alp_data.backends import BackendType
+from alp_data.dataset import register_config
 from alp_data.io import DATA_HOME, AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+# Both versions keep their metadata under the v0.1.0 path, so data_root is shared
+# across versions (only the split CSVs differ).
+_RAW_ROOT = f"{DATA_HOME}/inaturalist/v0.1.0/raw"
+
+
+@register_config
+class INaturalistConfig(DatasetConfig):
+    """Configuration for the iNaturalist dataset.
+
+    Parameters
+    ----------
+    dataset_name : str
+        The name of the dataset. Must be "inaturalist".
+    split : str
+        The split to load. One of INaturalist.info.split_paths keys.
+    version : str | None
+        The version of the dataset to use. If None, uses DEFAULT_VERSION.
+        Available versions: "0.1.0", "0.2.0".
+    output_take_and_give : dict[str, str] | None
+        A dictionary mapping the original column names to the new column names.
+        It acts as a filter as well.
+    sample_rate : int | None
+        The sample rate to which audio files should be resampled.
+    data_root : str | AnyPathT | None
+        The root directory for the dataset. If None, uses the default for the
+        selected version.
+    backend : BackendType
+        The backend to use ("pandas" or "polars"), by default "polars".
+    streaming : bool
+        Whether to use streaming mode, by default False.
+    """
+
+    dataset_name: str = "inaturalist"
+    split: str = "train"
+    version: str | None = None
+    output_take_and_give: dict[str, str] | None = None
+    sample_rate: int | None = None
+    data_root: str | AnyPathT | None = None
+    backend: BackendType = "polars"
+    streaming: bool = False
 
 
 @register_dataset
@@ -23,7 +65,7 @@ class INaturalist(Dataset):
     following ESP's taxonomy app (GBIF backbone),
     including species scientific and common names, family, genus, order.
     There is additional metadata including location, date, and recordist information.
-    The current version 0.1.0 includes iNaturalist data up to Jun 2026 (20260616 dump).
+    Version 0.1.0 uses the 20260201 split; version 0.2.0 adds the Jun 2026 (20260616) dump.
 
     Available Metadata Fields
     -------------------------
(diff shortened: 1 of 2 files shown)


GaganNarula:
@david-rx  question: can i run the rsync with `-no-clobber` I'm guessing most of the audio is already there ?

alp_data/datasets/inaturalist.py
```diff
@@ -7,8 +7,50 @@
...
+        selected version.
+    backend : BackendType
+        The backend to use ("pandas" or "polars"), by default "polars".
+    streaming : bool
+        Whether to use streaming mode, by default False.
+    """
+
+    dataset_name: str = "inaturalist"
+    split: str = "train"
+    version: str | None = None
```
GaganNarula, line 47: Just one thing, we have this one bug #289 . The bug doesn't allow reading the right config like INaturalistConfig when the dataset yaml defines the dataset within a `chain` or a `concat` list. The right version is still loaded because `DatasetConfig` allows extra parameters. Will fix in a future PR.

alp_data/datasets/inaturalist.py
```diff
@@ -23,7 +65,7 @@ class INaturalist(Dataset):
     following ESP's taxonomy app (GBIF backbone),
     including species scientific and common names, family, genus, order.
     There is additional metadata including location, date, and recordist information.
-    The current version 0.1.0 includes iNaturalist data up to July 2025.
+    Version 0.1.0 uses the 20260201 split; version 0.2.0 adds the Jun 2026 (20260616) dump.
```
GaganNarula, line 69: Maybe a link to the source of the data would be useful for tracking ? like the gbif darwin core dump file or wherever we got the metadata

alp_data/datasets/xeno_canto.py
```diff
@@ -50,6 +92,8 @@ class XenoCanto(Dataset):
         - ``behavior``: Behavior being recorded (e.g., "calling song")
         - ``sex``: Sex of the recorded animal(s)
         - ``lifeStage``: Life stage (e.g., "adult")
+        - ``quality``: Xeno-canto recording quality rating from ``A`` (highest) to
```
GaganNarula, line 95: maybe comment that v0.2.0 adds `quality` column

GaganNarula (commented):
a few comments but looks good. I have a question regarding the copy job, so if you let me know i'll start the transfer over the weekend

david-rx:
> @david-rx question: can i run the rsync with `-no-clobber` I'm guessing most of the audio is already there ?
Yes, there shouldn't be a need to overwrite and no-clobber sounds right. Most audio is already in the destination (and much of it only at the destination.)

alp_data/datasets/inaturalist.py
david-rx: Yes that is interesting for tracking, I could upload the dumps inside data-ingestion somewhere here

GaganNarula (approved):
transfer is finished, might be worth it to run a full iteration check over the two datasets ("all" split) , code looks good to go

PR #313, opened 2026-07-10
Fix -rw_timeout placement; downscope GCS tokens to per-bucket read-only

ReadyPlayerEmma:
Targets #310's branch (follow-up to the review discussion there). Three commits:

## 1. Fix `-rw_timeout` placement on the ffmpeg decode command

ffmpeg's CLI parsing is positional: options placed after the last output (`pipe:1`) are trailing options and are silently ignored (ffmpeg logs "Trailing option(s) found in the command" and continues). So the 30s `-rw_timeout` added on the decode command had no effect — a GCS connection that stalls silently mid-transfer would still hang the read indefinitely. The ffprobe placement was already correct.

This moves the pair into the input options (before `-i`), with a comment explaining the positional constraint.

**Verification** (local, ffmpeg 8.1.2):
- Against a deliberately stalling HTTP server (accepts the connection, sends headers + a few bytes, then goes silent), routed through the real `_read_audio_ffmpeg` path: previously hung until killed; with this fix it raises `FFmpegSegmentError("ffmpeg decode failed")` at 30.2s, reaching the existing warn-once → download-fallback path.

## 2. Downscope GCS tokens to per-bucket read-only

The raw ADC token carries the caller's full identity. Before handing a token to ffmpeg, `auth.py` now exchanges it through a [GCP Credential Access Boundary](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials) so the token only grants `roles/storage.objectViewer` on the single bucket being read, for at most the source token's lifetime.

- The bucket is derived from the `gs://` path — no configuration, no bucket allowlist to maintain.
- Downscoped credentials are cached per bucket; the STS exchange happens once per (bucket, token lifetime). Source credentials are cached as before and never leave the process.
- The gcsfs download/fallback path is unchanged (in-process credentials, full identity as before).
- Exchange failures surface as `GCSAuthError` and flow into the existing fallback machinery.

New tests: a live assertion that the token can read its own bucket (HTTP 206 ranged GET) but is rejected by non-storage services (401/403), and a unit test of the per-bucket credential cache.

**Verification**: `ruff check` / `ruff format --check` clean; `pytest tests/test_auth.py tests/test_read_audio.py`: 36 passed (the read_audio suite exercises the downscoped tokens end-to-end against GCS via the ffmpeg path).

## 3. Back off on auth failures instead of caching a credential-less verdict

The sticky `_gcs_credentials_unavailable` flag previously flipped on *any* `GCSAuthError` in the auto path, so a transient STS/network failure mid-session would silently downgrade all subsequent reads to anonymous (losing the ffmpeg fast path for private buckets for the rest of the process).

Now the credential-less verdict only sticks when ADC discovery itself fails (no ambient credentials at all). Failures with otherwise-present credentials — a network blip, a blocked STS endpoint, a revoked credential — set a 60s backoff instead: transient errors recover within a minute, while persistent ones cost one failed network call per backoff window rather than one per read in hot dataloader loops. Tests pin the sticky verdict, the backoff, and the post-backoff re-attempt.

### DIFF
--- a/alp_data/io/auth.py
+++ b/alp_data/io/auth.py
@@ -3,19 +3,19 @@
 Provides access-token retrieval for authenticating GCS REST/HTTP requests
 (e.g. the ffmpeg range-read path in `alp_data.io.read_utils`).
 
-Tokens handed out by this module are *downscoped* via a GCP Credential Access
-Boundary: each token is only valid for read-only access
-(`roles/storage.objectViewer`) to the single bucket it was requested for, and
-is rejected by every other GCP service and bucket. This limits the blast
-radius when a token leaves the process (e.g. on the ffmpeg command line,
-where it is visible to co-tenants via `ps` on shared hosts): a leaked token
-grants at most short-lived read access to the one bucket that was already
-being read.
+Tokens are *downscoped* through a GCP Credential Access Boundary: each token
+only grants read-only access (`roles/storage.objectViewer`) to the single
+bucket it was requested for. This limits the blast radius when a token leaves
+the process (e.g. on the ffmpeg command line, visible via `ps` on shared
+hosts): a leaked token gives at most short-lived read access to the one
+bucket that was already being read. The full-identity source credentials
+never leave the process.
 
-Source credentials are cached at module level so the Application Default
-Credentials lookup happens once per session; downscoped credentials are
-cached per bucket and refreshed (a token exchange with the GCP STS endpoint)
-only when they expire.
+Caching: the source credentials are resolved once per session; downscoped
+credentials are cached per bucket and re-exchanged (one STS call) only on
+expiry. A session with no ambient credentials at all is remembered as
+credential-less; other auth failures back off for
+`_AUTH_RETRY_BACKOFF_SECONDS` before re-attempting.
 """
 
 import logging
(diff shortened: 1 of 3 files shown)


alp_data/io/read_utils.py
```diff
@@ -105,6 +105,28 @@ def _gcs_path_to_url(file_path: str | AnyPathT) -> str:
     return f"https://storage.googleapis.com/{path_str}"
 
 
+def _gcs_bucket_name(file_path: str | AnyPathT) -> str:
+    """Extract the bucket name from a GCS path.
```
GaganNarula: i think we don't need this function: see `alp_data.io.paths.PureCloudPath` property `bucket` ... just `anypath(file_path).bucket` is enough

alp_data/io/auth.py
```diff
@@ -1,57 +1,121 @@
...
+    )
+
+
+def get_gcs_token(bucket: str) -> str:
+    """Fetch a valid access token downscoped to read-only access on `bucket`.
 
     Relies on the caller having authenticated with GCP, e.g. via
-    `gcloud auth application-default login` or a service account. The
-    credentials are cached in _gcs_credentials.
+    `gcloud auth application-default login` or a service account. The ambient
```
GaganNarula: i think the comments are quite verbose and there's a bit of repetition between the module doc string and function doc strings.. lets make it a bit concise

alp_data/io/auth.py
```diff
@@ -60,23 +124,43 @@ def get_gcs_token() -> str:
...
+        Name of the GCS bucket the token must grant read access to.
+
+    The credential-less verdict is only cached (sticky for the session) when
+    Application Default Credentials discovery itself fails — i.e. the
+    environment has no ambient credentials. A failure while refreshing or
+    downscoping an otherwise-working credential (e.g. a network blip or a
+    blocked STS endpoint) returns None and pauses re-attempts for
+    `_AUTH_RETRY_BACKOFF_SECONDS`, so transient errors recover quickly while
+    persistent ones cost one failed network call per backoff window instead
+    of one per read.
```
GaganNarula: if auth *is* required for a bucket (a private one), the code will throw an exception anyway right ? so this retry backoff window is for the case when we're fetching the credential even though we dont need to (public bucket ?)

GaganNarula (changes requested):
a couple of small changes and a comment, looks good!

alp_data/io/auth.py
ReadyPlayerEmma: Almost. In the auto path (anonymous=None) nothing actually throws for private buckets either. The chain is: token fetch fails → ffmpeg tries anonymously → 403 → FFmpegSegmentError → warn-once → falls back to the gcsfs download path, which authenticates with its own in-process credentials and succeeds. So a private-bucket read degrades to the slow path rather than erroring, and the backoff saves the failed STS call (~300ms measured) that every read — public or private — would otherwise pay per read.

"Our auth fails but gcsfs works" is not just a hypothetical: gcsfs only needs ADC + the storage API, while the token path additionally needs sts.googleapis.com (the downscoping exchange) and a cloud-platform-scoped source token. Concretely:

- An egress allowlist that includes storage.googleapis.com + oauth2.googleapis.com (all gcsfs needs) but not sts.googleapis.com — persistent and per-environment.
- GCE VMs with legacy default access scopes: their tokens carry devstorage.read_only but not cloud-platform, which gcsfs happily uses but the STS exchange requires ([docs](https://cloud.google.com/iam/docs/create-downscoped-short-lived-credentials): "You must provide the cloud-platform scope"). Worth a one-off check from an actual compute node — just call get_gcs_token("esp-ci-cd-tests") there — before we count on the authenticated fast path on the cluster.
- Very large array jobs cold-starting in the same minute could brush the STS quota (6,000 exchanges/project/minute) — unlikely at our scale, but if it happens the 60s backoff absorbs it and reads proceed via the fallback.

You're right though that for public buckets the token is pure overhead. Skipping auth entirely for known-public buckets (or having the dataset layer pass anonymous=True) would avoid the whole machinery for most of our reads.

GaganNarula: alright makes sense, main problem with being able to pass `anonymous=True` is that then we have to keep a registry of public and non-public buckets which needs maintenance

PR #316, opened 2026-07-21
Add additional CI gate

ReadyPlayerEmma:
Adds an additional CI gate in case of manual approval which provides another layer of protection for the runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,15 @@ on:    # yamllint disable-line rule:truthy
 jobs:
     tests:
         name: Run pytests tests
+        # SECURITY: secondary gate — keep fork-PR code off self-hosted infra.
+        # The primary gate is the repo's fork-PR approval policy ("require
+        # approval for all external contributors"): every fork-PR run needs a
+        # maintainer click. This condition backstops that human step: an
+        # approved fork PR with an unmodified workflow still skips this job.
+        # It cannot be the only gate — fork PRs execute their own copy of this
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,6 +9,15 @@ on:    # yamllint disable-line rule:truthy
...
         name: Run pytests tests
+        # SECURITY: secondary gate — keep fork-PR code off self-hosted infra.
+        # The primary gate is the repo's fork-PR approval policy ("require
+        # approval for all external contributors"): every fork-PR run needs a
+        # maintainer click. This condition backstops that human step: an
+        # approved fork PR with an unmodified workflow still skips this job.
+        # It cannot be the only gate — fork PRs execute their own copy of this
+        # file, so a PR that edits this workflow can strip this line. Before
+        # approving a fork-PR run, always check the diff for workflow changes.
+        if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```
GaganNarula, line 20: lets remember to remove / replace this huge comment when we fix the primary gate

GaganNarula (approved):
lgtm!

.github/workflows/ci.yml
ReadyPlayerEmma: This is an additional layer of protection that remains useful even with the primary gate (which I also enabled). And the comment itself helps reduce the risk of human factors in case a PR shows up that removes these protections. It helps remove ambiguity about why the line is there for any future reviewers who might otherwise approve the CI run.

PR #322, opened 2026-08-16
Add FASD13 few-shot detection benchmark

david-rx:
## What

Adds `FASD13`, the evaluation benchmark from created by benj in the DRASDIC paper ([Zenodo 15843741](https://zenodo.org/records/15843741), [arXiv:2503.00296](https://arxiv.org/abs/2503.00296)): 13 bioacoustics sub-datasets, 109 recordings, ~143 h, 11,315 positive events, each recording annotated for a single predetermined target category. Anurans, birds, primates, whales, insects, spiders and gunshots; terrestrial and underwater PAM, on-body recorders, substrate-borne recording and the lab.

It is unrelated in content to the synthetic *training* corpus released by the same paper — FASD13 shares no audio with it and is deliberately evaluation-only.

The dataset was built and validated in the `esp-data` fork; this PR ports it into `alp-data` and trims it to the conventions here.

```python
from alp_data.datasets import FASD13

ds = FASD13(split="AS", sample_rate=32000)   # "all" or one of 13 sub-dataset codes
item = ds[0]                                 # audio, sample_rate, selection_table, metadata
```

## Design

**WABAD-shaped, no custom surface.** One row per recording plus a Raven-style selection table. The fork's multi-audio `audios` mode is deliberately *not* ported — it exists to serve one downstream model's prompt format and has no consumer here.

**Windowed reads, so the benchmark is actually usable.** A row carrying `window_start_sec` / `window_end_sec` reads only that segment, streamed from GCS via ffmpeg range requests. This is the existing alp-data convention, not a FASD13 invention: `dclde2026.py:315` has the identical branch (with those columns shipping in its manifest), and four datasets here already do segment reads.

It matters more for FASD13 than for most: several sub-datasets ship 8-hour files, so one `HG` item is **3.69 GB of float32** at 32 kHz. Cutting a 10 s support clip or chunking a query region has to happen at read time or not at all.

**Selection tables are re-based onto the window.** FASD13's tables are recording-absolute, so returning them unchanged next to windowed audio would hand every caller an internally inconsistent row. Non-overlapping events are dropped, the rest shifted and clipped to the returned audio; `event_index` is preserved so a caller can still tell which of the recording's events a row refers to.

**The N-shot protocol is metadata, not loader behaviour.** Each row carries `n_shots_available` and `shot_end_times` (end times of the first five non-`UNK` events). With windowed reads, that is everything an episode builder needs, so episode construction, prompt rendering and support-clip selection all stay downstream.

**Labels.** The positive class is nameless — in few-shot detection the target is defined by the support examples rather than by a name — so the selection table's `Label` is the constant `"target"`, and the row-level `detection_target` records the kind of category (Species, Call Type, Sound Type, Production Mechanism, Species+Life Stage). Per the official protocol `UNK` events are neither positives nor negatives and must be masked out of scoring.

**Licensing is per sub-dataset**, not uniform: CC-BY-1.0, CC-BY-4.0, CC-BY-SA-4.0, CC-BY-NC-4.0, CC-BY-NC-SA-4.0, public domain, and one citation-required entry (`MS`). It is carried in per-row `license` / `citation` columns, with `info.license` pointing at them.

**Audio.** 16 kHz and 32 kHz mono mirrors; 32 kHz is the default. Native-rate audio was never mirrored — `audio_fp` points at the 32 kHz file — so any other target rate resamples *from 32 kHz* and `native_sample_rate` is informational only. The docstring states this rather than implying an originals pointer exists.

## Downstream few-shot construction works on public API alone

Verified end-to-end, not assumed. A user-registered transform (the mechanism from `docs/tutorials/07_user_datasets_and_transforms.ipynb`) expands one 8-hour HG recording into 5 support windows + 3 query windows; iterating the dataset then yields:

```
support_1   window  5606.00- 5616.00s -> 10.00s audio, events(relative)=[(1.0, 2.0)]
support_5   window  5667.00- 5677.00s -> 10.00s audio, events(relative)=[(5.0, 10.0)]
query_0     window  5677.00- 5687.00s -> 10.00s audio, events(relative)=[]
query_1     window  5687.00- 5697.00s -> 10.00s audio, events(relative)=[(4.0, 8.0)]

loaded 80.0s of audio total (10.2 MB) in 6.8s
peak python heap during reads: 11.6 MB
a full read would have been 3.69 GB per item, 8 times over
```

Support clips stay bounded by the Nth event and the first query window is clean, so there is no leakage across the support/query boundary.

## Files

- `alp_data/datasets/fasd13.py` — the dataset, modelled on `wabad.py`.
- `alp_data/datasets/__init__.py` — export.
- `tests/unittests/datasets_tests/test_fasd13.py` — 21 tests.
- `scripts/data_preprocessing_scripts/fasd13.py` + three job scripts — the four-stage build (download → resample → manifests → support clips), for provenance and rebuilds.
- `.dict-allowed.txt` — `Hart`, a surname in the citation table.

No version bump — that's a separate release PR.

## Test design

Every audio read is pinned to `AS`, the smallest sub-dataset (12 files, 0.20 h), at **fixed** indices — with one deliberate exception. Random-index sampling à la `test_wabad.py` would be a CI hazard here given the 8-hour recordings.

The exception is `test_windowed_read_on_long_recording`, which reads a 10 s window from an HG file on purpose: if the windowed path ever regresses to a full-file read, it surfaces as a multi-GB test rather than a silent slowdown.

Coverage: published sizes (109 recordings / 12 in AS), split names, invalid-split `LookupError`, audio integrity and dtype, 16 kHz mirror selection (exactly half the samples of the 32 kHz read), pandas/polars parity, streaming, selection-table schema and in-bounds event times, windowed reads (bit-identical to slicing the full decode) and window-relative tables, absolute tables when unwindowed, N-shot columns, per-row licensing, `from_config`, `output_take_and_give`, and audio + manifest SHA256 stability.

## Validation

```
uv run ruff check . / ruff format --check   pass
uv run pytest tests/unittests/datasets_tests/test_fasd13.py    21 passed
uv run pytest tests/test_dataset.py tests/test_chained.py tests/test_concat.py    33 passed, 2 skipped
uv run deptry .                             pass
uv run mkdocs build --strict                pass (renders on Available Datasets)
uv run prek run --files <changed>           pass
```

The tests were run green against the staged data. Reference hashes were computed from those same objects, so they carry over unchanged after the copy below.

## ⚠️ Before merging: data copy

The loader points at the canonical `f"{DATA_HOME}/fasd13/v0.1.0/raw"` (as 33 of ~36 datasets do), but the data is still in the ingestion bucket, **so CI will be red until the copy lands.**

Manifests carry audio pointers relative to their own directory, so the tree must land layout-preserving directly under `raw/`:

```
gsutil -m rsync -r -x '^raw/' gs://esp-data-ingestion/fasd13/v0.1.0/ \
                              gs://esp-data-274503/fasd13/v0.1.0/raw/
gsutil -m rsync -r gs://esp-data-ingestion/fasd13/v0.1.0/raw/ \
                   gs://esp-data-274503/fasd13/v0.1.0/raw/source/
```

The second line moves the Zenodo provenance annotations from `raw/` to `source/`, avoiding a `raw/raw/`; the job scripts already write to the new layout. This also copies `fasd13_support.csv` and the 545 pre-cut support clips — negligible in size, unused by this class, and kept so the existing downstream few-shot path keeps working. Say the word if you'd rather leave them out.


scripts/data_preprocessing_scripts/fasd13.py
```diff
@@ -0,0 +1,996 @@
...
+    -------
+    str
+    """
+    out = pd.DataFrame(
+        {
+            "Selection": range(1, len(df) + 1),
+            "Begin Time (s)": [round(float(v), 4) for v in df["Starttime"]],
+            "End Time (s)": [round(float(v), 4) for v in df["Endtime"]],
+            "Q": list(df["Q"]),
+            "Label": [TARGET_LABEL] * len(df),
```
benjaminsshoffman: what happens to UNK labels here? It seems like they get re-labeled as positive in the "Label" column?

scripts/data_preprocessing_scripts/fasd13.py
```diff
@@ -0,0 +1,996 @@
+"""Build the FASD13 few-shot detection benchmark (Zenodo 15843741).
+
+Hoffman et al. 2025, "Synthetic data enables context-aware bioacoustic sound
+event detection" (arXiv:2503.00296). FASD13 is the *evaluation* benchmark from
+the DRASDIC paper: 13 bioacoustics sub-datasets, 109 recordings, ~143 h, each
+recording paired with a DCASE-style CSV marking onsets/offsets of a single
+predetermined target category.
+
+Not to be confused with DRASDIC, the *synthetic training* corpus released by the
```
benjaminsshoffman: The synthetic training corpus wasn't named DRASDIC, I guess it was the model that was named DRASDIC

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,461 @@
...
+    ------
+    Every recording is annotated for a single, *nameless* target category. In
+    few-shot detection the target is defined by the support examples rather
+    than by a name, so the selection table's `Label` column is the constant
+    `"target"`, and the row-level `detection_target` column records what kind
+    of category the sub-dataset annotates (Species, Call Type, Sound Type,
+    Production Mechanism, or Species+Life Stage).
+
+    Selection tables carry `Selection`, `Begin Time (s)`, `End Time (s)`, `Q`,
+    `Label` and `event_index`. `Q` is one of `POS`, `UNK` or `NEG`. Per the
```
benjaminsshoffman: what's the difference between "Label" and "Q"? Probably only one should be included.
Also, "NEG" is weird since everything not POS or UNK is implicitly NEG. So these events can just be removed (if they exist to begin with)

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,461 @@
...
+    "RS",
+    "RW",
+)
+
+
+@register_dataset
+class FASD13(Dataset):
+    """FASD13 Dataset
+
+    Description
```
benjaminsshoffman, line 43: Readme should provide a per-dataset name that is not an abbreviation, and also a citation

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,461 @@
...
+    not in the expectation of extra bandwidth.
+
+    Every `HA` recording is a FLAC bitstream carrying a `.wav` extension. Two
+    of them (`Hawaii_UHH_494_S04_20190418_203000.wav` and
+    `Hawaii_UHH_627_S02_20220323_100400.wav`) hold frames libsndfile cannot
+    decode, so their originals are published transcoded to plain PCM by ffmpeg,
+    which reads them in full. `HA` is 32 kHz mono at source, so this is a
+    container change only: both transcodes are bit-identical to the published
+    32 kHz mirror. The files exactly as Zenodo ships them are kept under
+    `raw/HA/` for provenance.
```
benjaminsshoffman: not sure if this claudespeak should be in readme

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,461 @@
...
+    dropped and the rest are shifted and clipped so their times line up with
+    the returned audio. `event_index` is preserved, so a caller can still tell
+    which of the recording's events a row refers to.
+
+    This is what makes the benchmark usable without loading whole recordings.
+    Several sub-datasets ship 8-hour files -- one `HG` item is ~3.7 GB of
+    float32 at 32 kHz -- so cutting an N-shot support clip or chunking a query
+    region has to happen at read time.
+
+    Pre-resampled Audio
```
benjaminsshoffman, line 140: might be worth leaving a comment that KD often contains event audio above the nyquist of 32khz

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,461 @@
...
+    segment, streamed from cloud storage rather than downloaded whole, and its
+    selection table is re-based onto the window: non-overlapping events are
+    dropped and the rest are shifted and clipped so their times line up with
+    the returned audio. `event_index` is preserved, so a caller can still tell
+    which of the recording's events a row refers to.
+
+    This is what makes the benchmark usable without loading whole recordings.
+    Several sub-datasets ship 8-hour files -- one `HG` item is ~3.7 GB of
+    float32 at 32 kHz -- so cutting an N-shot support clip or chunking a query
+    region has to happen at read time.
```
benjaminsshoffman: this is fine but you need to be careful that across windowed reads, you're using the same support set from the beginning of that file.

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,461 @@
...
+    must be masked out of scoring. There are no frequency bounds -- FASD13 is
+    time-only.
+
+    N-shot protocol
+    ---------------
+    Following Nolasco et al. 2023, events are ordered by end time and `UNK`
+    events are excluded from shot selection. An N-shot system is given the
+    audio up through the end of the Nth event and must detect events in the
+    remainder, which is the only region scored. Each row carries
+    `n_shots_available` and `shot_end_times` (the end times of the first five
```
benjaminsshoffman: seems delicate with the windowed reads -- end times might refer to another row

benjaminsshoffman: Also, am I understanding right that these are pre-computed and stored in the dataset csv? Seems inflexible, and recomputing these is cheap so can be accomplished within the dataset.

benjaminsshoffman (changes requested):

scripts/data_preprocessing_scripts/fasd13_rebuild_support_job.sh
```diff
@@ -0,0 +1,40 @@
...
+#SBATCH --output="/home/%u/logs/%A_%x.log"
+#SBATCH --qos=naturelm
+
+# Rebuild the FASD13 few-shot support clips with the support region correctly
+# bounded by the Nth event, and re-upload.
+#
+# Reads slices straight from the GCS mirrors, so the ~50 GB of audio does not
+# need re-staging. Only the annotation CSVs (small) are pulled down.
+set -euo pipefail
+cd "${HOME}/alp-data"
```
GaganNarula: do we need to commit this job ?

scripts/data_preprocessing_scripts/fasd13.py
david-rx: Great catch thanks

alp_data/datasets/fasd13.py
david-rx: Hopefully fixed haha

alp_data/datasets/fasd13.py
david-rx: Added helpers instead and removed from csv

alp_data/datasets/fasd13.py
david-rx: Added explanation

scripts/data_preprocessing_scripts/fasd13_rebuild_support_job.sh
david-rx: removed

alp_data/datasets/fasd13.py
david-rx: Removed to be only Q and removed the NEG thing (it is not there)

scripts/data_preprocessing_scripts/fasd13_build_job.sh
```diff
@@ -0,0 +1,81 @@
...
+SCRIPT="scripts/data_preprocessing_scripts/fasd13.py"
+WORKERS="${SLURM_CPUS_PER_TASK:-16}"
+
+# Ignore stale user ADC; use the node's attached SA (gsutil + gcsfs).
+export CLOUDSDK_CONFIG="$(mktemp -d)"
+export GCE_METADATA_MTLS_MODE=none
+echo "Node: $(hostname)  stage=${STAGE}  workers=${WORKERS}"
+echo "ok" | gsutil cp - "${GCS}/.auth_probe" && gsutil rm "${GCS}/.auth_probe" && echo "auth probe OK"
+
+# 1. Download + verify + extract (resumable; cached files are checksum-checked).
```
GaganNarula: why do we need _build_job and _validate_job ? if recording the workflow is important let is just be an addition to the preprocessing script doc string detailing steps take. Also, why is the GCS variable pointing to the target bucket already ? i'm guessing that copy job never ran ?

GaganNarula: it seems its already in the doc string (at least the build job). Is the validate job supposed to run on the target bucket ?

scripts/data_preprocessing_scripts/fasd13_validate_job.sh
```diff
@@ -0,0 +1,102 @@
...
+short=[(r["subdataset"],r["sound_name"],len(ds.shot_end_times(i)))
+       for i,r in enumerate(rows) if len(ds.shot_end_times(i))<5]
+print("recordings with <5 shots:",len(short))
+assert not short, short
+
+# UNK must never be usable as a shot
+import pandas as _pd
+from io import StringIO as _S
+_st=_pd.read_csv(_S(rows[0]["selection_table"]),sep="\t")
+assert "Label" not in _st.columns, "Label column marks UNK events as positive"
```
GaganNarula: these seem like unit tests..

scripts/data_preprocessing_scripts/fasd13_validate_job.sh
```diff
@@ -0,0 +1,102 @@
...
+#SBATCH --cpus-per-task=4
+#SBATCH --mem=16G
+#SBATCH --time=01:00:00
+#SBATCH --output="/home/%u/logs/%A_%x.log"
+#SBATCH --qos=naturelm
+set -euo pipefail
+cd "${HOME}/alp-data"
+export CLOUDSDK_CONFIG="$(mktemp -d)"   # attached SA for gsutil + gcsfs
+export GCE_METADATA_MTLS_MODE=none
+GCS="gs://esp-data-274503/fasd13/v0.1.0/raw"
```
GaganNarula: i think i'd rather not include this job, can be kept on your nfs or elsewhere.. pretty easy to regenerate if needed

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,520 @@
...
+            Duration of the audio actually returned, in seconds. Used instead
+            of the requested window end so that a window running past the end
+            of the file is handled correctly.
+
+        Returns
+        -------
+        pd.DataFrame
+            The windowed selection table, with times relative to the window.
+        """
+        window_end = window_start + duration
```
GaganNarula, line 321: i thought this was a helper in the other PR ? should that go in first ?

alp_data/datasets/fasd13.py
```diff
@@ -0,0 +1,520 @@
...
+            streaming=cfg["streaming"],
+        )
+
+        if dataset_config.transformations:
+            meta = ds.apply_transformations(dataset_config.transformations)
+            return ds, meta
+
+        return ds, {}
+
+    def shot_end_times(self, idx: int, n_shots: int = MAX_SHOTS) -> list[float]:
```
GaganNarula: is this public method necessary ? it adds another branch out of the dataset api.. `get_available_labels` was already concession (many datasets dont have that one). i'd rather that the client using this dataset has its own helpers

GaganNarula (changes requested):
some more comments to work through. I can transfer the data already, let me know

alp_data/datasets/fasd13.py
david-rx: Just wanted it clear how to do it in a consistent way, but fine for it to be done by the client

alp_data/datasets/fasd13.py
david-rx: That's true, would rather get this one in first if possible, that PR is not 100% baked and slightly lower priority for me. I've left the windowing as-is here for now but open to what you think (e.g. I can update it in that PR when a shared util is added?)

scripts/data_preprocessing_scripts/fasd13_validate_job.sh
david-rx: deleted both jobs

david-rx:
Addressed with one remaining question, happy for the data to be copied whenever

scripts/data_preprocessing_scripts/fasd13.py
benjaminsshoffman: This still doesn't look corrected

GaganNarula (approved):
Data has been copied, so you can first change the _RAW_ROOT to contain DATA_HOME. For that question about shot times, i think the current comment is good enough. An agent using this dataset should (hopefully) catch it.

I'm guessing we will refactor the private window selection table method in the next PR ?

benjaminsshoffman (approved):

PR #323, opened 2026-08-19
Add directory map to AGENTS.md

### LINKED ISSUE #271: Add directory map to agents.md
can we give it a directory map so it knows where datasets, docs, transforms, io, utils etc are located and what those modules do ? or in a second PR ?

_Originally posted by @GaganNarula in https://github.com/earthspecies/esp-data/pull/268#discussion_r3112394744_

GaganNarula:
Closes #271.

`AGENTS.md` had a `Key Directories` section listing only `alp_data` with no breakdown, so agents had no way to know where datasets, transforms, io or docs live without exploring the tree first. This expands it into a full directory map.

What's added

- **`alp_data` top-level modules** — `dataset.py` (the `Dataset` ABC, the pydantic config models, and the dataset registry, flagged as the place to start for the dataset contract), `concat.py`, `chain.py`, `utils.py`.
- **Subpackages** — `io` broken down file-by-file (`paths`, `filesystem`, `file_utils`, `read_utils`, `auth`), plus `datasets`, `transforms`, `backends` and `discover`. Each names its key entry points and the conventions that aren't obvious from the file listing, e.g. `transforms/registry.py` must be imported before the transform modules, and adding a dataset means adding a module *and* exporting it from `datasets/__init__.py`.
- **Everything else** — `tests` (the split between core tests, `tests/io` and `tests/unittests/datasets_tests`), `docs` (noting new pages must be added to `nav` in `mkdocs.yml`), `_hooks`, `scripts` and its subdirectories, `jobs`, `notebooks`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,7 +8,74 @@
 - Make sure that only objects that are needed outside a module are exposed in the `__init__.py`. Tests don't count as external usage; they should import directly from the submodule if something isn't in `__init__.py`.
 
 # Key Directories
-- `alp_data` - main python library
+
+## `alp_data` - main python library
+
+Top-level modules:
+
+- `dataset.py` - the `Dataset` abstract base class plus the pydantic config models
+  (`DatasetConfig`, `ConcatConfig`, `ChainedDatasetConfig`), `DatasetInfo`, and the dataset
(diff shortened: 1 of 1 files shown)


Paulchen-git (approved):
👍

PR #324, opened 2026-08-27
Add four strongly-labeled soundscape datasets (Dartmouth, PteroSet, Indian Fauna Type-A, Xeno-canto Strong)

david-rx:
## What

Ports four strong-detection soundscape datasets from `david-nlm`:

| dataset | source | recordings | events | species |
|---|---|---|---:|---:|
| `DartmouthAvianSoundscapes` | [Zenodo 20038954](https://zenodo.org/records/20038954) | 1,302 × 10 min | 184,207 | 99 |
| `PteroSet` | [Zenodo 19137071](https://zenodo.org/records/19137071) | 563 | 15,372 | 159 |
| `IndianFaunaStrong` | [Zenodo 18743214](https://zenodo.org/records/18743214) | 1,702 | 36,639 | 397 |
| `XenoCantoStrong` | xeno-canto annotated subset | 21,192 | 68,413 | 808 |

```python
from alp_data.datasets import DartmouthAvianSoundscapes, PteroSet, XenoCantoStrong

ds = DartmouthAvianSoundscapes(split="acad", sample_rate=32000)
item = ds[0]        # audio, sample_rate, selection_table, metadata
```

`david-nlm` predates the `esp_data` → `alp_data` rename, so this is a transplant rather than a cherry-pick.

## They really are one family

Verified against the shipped data, not assumed. All four are WABAD-shaped: one row per recording, an inline Raven-style `selection_table`, `annotation_columns == ["Species"]`, 16 k/32 k mirrors, `available_sample_rates == [16000, 32000]`, and windowed reads. Every selection table carries `Begin Time (s)`, `End Time (s)`, `Low Freq (Hz)`, `High Freq (Hz)`, `Species`, plus per-dataset extras (dartmouth `Species Code`/`Common Name`/`Background`; pteroset `Species Code`/`Identification`/`Type`; indian `Selection`; xc `sound_type`/`sex`/`life_stage`/`annotator`).

## Fixes a bug all four shared

Each honoured `window_start_sec`/`window_end_sec`, then clipped its selection table with `Begin Time (s) < audio_dur` against the **window** length while leaving the times recording-absolute. A window at [1000, 1010] therefore dropped every event it contained.

Five datasets now need identical logic (these four plus FASD13 in #322), so the fix lands as a shared helper — `alp_data/datasets/_windowing.py` — with the semantics settled in #322: drop non-overlapping events, shift onto the returned audio, clip to it, leave identity columns untouched. Each dataset has a regression test.

Once #322 merges, a one-line follow-up switches FASD13 from its private copy to this helper.

## GBIF linkage

Three of four resolve cleanly:

- **Dartmouth** (AOU 4-letter codes) and **PteroSet** (PteroSet codes) were coerced to GBIF canonical names through *this repo's own* `GBIFConverter`, and ship a `species_labels.csv` in exactly the shape of WABAD's `gbif_labels.csv`.
- **xc_strong** carries file-level GBIF columns from the XC pipeline; event labels are annotator-supplied binomials.

Exact-string overlap with WABAD's 1,188-name vocabulary is 68 % / 47 % / 20 % / 40 %. The gaps are biogeographic — xc is global and overlaps all three, the regional ones barely overlap each other — and the names that *do* overlap match exactly, which is the evidence they share one canonical-name convention.

One caveat for reviewers: the fork's `GBIFConverter` has since diverged from `main`'s by ~495 lines (`main` reads a precomputed cache and applies `SCI_NAME_CORRECTION_MANUAL`). A Slurm job is confirming the older converter produces no disagreement; nothing here depends on the outcome.

## ⚠️ `indian_fauna_strong` ships with known gaps

Documented in its docstring rather than hidden. Neither affects the 16 k/32 k read paths, and the other three are unaffected:

- **Labels are not normalised through this repo's converter** — they come from the upstream release's `taxa_info`, so `Species` mixes ranks: genera (`Corvus`), families (`Gryllidae`), classes (`Aves`, `Insecta`), subspecies trinomials, one common name (**`Dugong`, 320 events**) and one non-taxon value (`Camera`). Filtering to binomials silently drops the Dugong events.
- **No native-rate audio** — `audio_fp` points at the 32 kHz mirror because originals were never staged. Native rates run 8–384 kHz and ~15 % of Mammalia boxes are fully ultrasonic, so that content is currently unreachable. The other three point at true originals (`recordings/`, `recordings/`, `audio/`).

Both are being cleaned up now and will land as a follow-up.

## Data location

Everything stays in `gs://esp-data-ingestion/` and moves under `DATA_HOME` at merge. Each loader reads a single `_RAW_ROOT` constant, so that's a one-line change per dataset.

Worth recording: `gs://esp-data-274503` is **write-protected** — an attempted copy failed with `storage.objects.create denied` for `david@earthspecies.org`, so the move needs someone with write access. Nothing was written.

## Validation

```
uv run ruff check . / ruff format --check         pass
uv run pytest <the four test files>               72 passed
uv run pytest tests/test_dataset.py tests/test_chained.py tests/test_concat.py   33 passed, 2 skipped
uv run deptry .                                   pass
uv run mkdocs build --strict                      pass (all four render on Available Datasets)
uv run prek run --files <changed>                 pass
```

Tests pin audio reads to the smallest split at fixed indices — dartmouth ships 1,302 × 10-minute recordings and indian runs to 60 minutes, so random sampling would be a CI hazard. 18 tests per dataset from one shared template, so the four stay consistent as they evolve.

### DIFF
--- a/alp_data/datasets/_windowing.py
+++ b/alp_data/datasets/_windowing.py
@@ -1,0 +1,74 @@
+"""Shared helpers for datasets that support windowed reads.
+
+Several strong-detection datasets accept `window_start_sec` / `window_end_sec`
+on a row and read only that segment (see `alp_data.datasets.dclde2026`, whose
+manifest ships those columns). Their selection tables hold recording-absolute
+times, so a windowed read has to re-base the table onto the returned audio or
+the two disagree.
+"""
+
+from __future__ import annotations
+
+import pandas as pd
+
+BEGIN = "Begin Time (s)"
+END = "End Time (s)"
+
+
+def window_selection_table(st: pd.DataFrame, window_start: float, duration: float) -> pd.DataFrame:
+    """Re-base a recording-absolute selection table onto a windowed read.
+
+    Events that do not overlap the window are dropped; the rest are shifted so
+    their times are relative to the start of the returned audio, and clipped to
+    it. All other columns are left untouched, so identity columns such as
+    `Selection` or `event_index` still say which of the recording's events a
+    row refers to.
+
+    Parameters
+    ----------
+    st : pd.DataFrame
+        Selection table with recording-absolute `Begin Time (s)` and
+        `End Time (s)`. A table missing those columns is returned unchanged.
+    window_start : float
+        Start of the window, in seconds from the start of the recording.
+    duration : float
+        Duration of the audio actually returned, in seconds. Used in place of
+        the requested window end so a window running past the end of the file
+        is handled correctly.
+
+    Returns
+    -------
+    pd.DataFrame
+        The windowed selection table, with times relative to the window.
+    """
+    if BEGIN not in st.columns or END not in st.columns:
+        return st
+    window_end = window_start + duration
+    st = st[(st[END] > window_start) & (st[BEGIN] < window_end)].copy()
+    st[BEGIN] = (st[BEGIN] - window_start).clip(lower=0.0)
+    st[END] = (st[END] - window_start).clip(upper=duration)
+    return st
+
+
+def clip_selection_table(st: pd.DataFrame, duration: float) -> pd.DataFrame:
+    """Drop events that begin past the end of the decoded audio.
+
+    The guard WABAD and DCLDE2026 apply to unwindowed reads, kept here so every
+    dataset in this family spells it the same way.
+
+    Parameters
+    ----------
+    st : pd.DataFrame
+        Selection table with recording-absolute times. A table missing
+        `Begin Time (s)` is returned unchanged.
+    duration : float
+        Duration of the decoded audio, in seconds.
+
+    Returns
+    -------
+    pd.DataFrame
+        The clipped selection table.
+    """
+    if BEGIN not in st.columns:
+        return st
+    return st[st[BEGIN] < duration].copy()
--- a/alp_data/datasets/dartmouth_avian_soundscapes.py
+++ b/alp_data/datasets/dartmouth_avian_soundscapes.py
@@ -1,0 +1,350 @@
+"""Dartmouth Avian Soundscapes dataset."""
+
+from __future__ import annotations
+
+from io import StringIO
+from typing import Any, Iterator
+
+import librosa
+import numpy as np
+import pandas as pd
+
+from alp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from alp_data.backends import BackendType
+from alp_data.datasets._windowing import clip_selection_table, window_selection_table
+from alp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+# Staged in the ingestion bucket. Moves under DATA_HOME (as WABAD and
+# most datasets resolve) when this lands; the loader picks the new
+# location up from this one constant.
+_RAW_ROOT = "gs://esp-data-ingestion/dartmouth-avian-soundscapes/v0.1.0"
+SPECIES_INFO_PATH = f"{_RAW_ROOT}/species_labels.csv"
+
+
+@register_dataset
+class DartmouthAvianSoundscapes(Dataset):
+    """Dartmouth Avian Soundscapes Dataset.
+
+    Description
+    -----------
+    A large-scale, stratified, fully annotated acoustic forest soundscape
+    dataset of avian vocalisations from eastern North America. Each entry is a
+    10-minute soundscape recording plus a selection table; each row of the
+    selection table is a time/frequency-bounded acoustic event annotated by
+    expert ornithologists. Species labels are coerced to GBIF canonical
+    scientific names (in the ``Species`` column), while the original 4-letter
+    AOU code is retained in the ``Species Code`` column for traceability.
+
+    The collection contains 1,302 ten-minute recordings (~183,000 annotated
+    vocalisations) of 96 bird species (plus a few mammal and unknown labels)
+    recorded across four protected areas in the northeastern United States
+    (Acadia NP, Hubbard Brook Experimental Forest, Katahdin Woods and Waters
+    NM, and Marsh-Billings-Rockefeller NHP) during the 2022 and 2023 breeding
+    seasons. Annotations were produced in Raven Pro 1.6 over three review
+    rounds. Originally included in alp-data for use as a zero-shot bird
+    detection evaluation dataset.
+
+    Splits
+    ------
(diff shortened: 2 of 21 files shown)


scripts/data_preprocessing_scripts/indian_fauna_strong_job.sh
```diff
@@ -0,0 +1,61 @@
...
+#SBATCH --partition=cpu
+#SBATCH --ntasks=1
+#SBATCH --cpus-per-task=16
+#SBATCH --mem=64G
+#SBATCH --time=08:00:00
+#SBATCH --output="/home/%u/logs/%A_%x.log"
+#SBATCH --qos=naturelm
+
+# ───────────────────────────────────────────────────────────────────
+# Build the IEN Type-A STRONG detection dataset (Zenodo 18743214) into
```
GaganNarula, line 12: do we need these jobs ? i dont think we do

alp_data/datasets/xeno_canto_strong.py
```diff
@@ -0,0 +1,395 @@
...
+
+# Staged in the ingestion bucket. Moves under DATA_HOME (as WABAD and
+# most datasets resolve) when this lands; the loader picks the new
+# location up from this one constant.
+_RAW_ROOT = "gs://esp-data-ingestion/xeno-canto/v0.1.0/raw"
+
+
+@register_dataset
+class XenoCantoStrong(Dataset):
+    """Xeno-canto event-level strongly-annotated dataset (WABAD-shaped).
```
GaganNarula, line 41: is there overlap between this one and XenoCanto jeantet ?  if there is significant overlap its worth noting, might even be nice to point to jeantet as being a similar dataset

alp_data/datasets/xeno_canto_strong.py
```diff
@@ -0,0 +1,395 @@
...
+import pandas as pd
+
+from alp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from alp_data.backends import BackendType
+from alp_data.datasets._windowing import clip_selection_table, window_selection_table
+from alp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+# Staged in the ingestion bucket. Moves under DATA_HOME (as WABAD and
+# most datasets resolve) when this lands; the loader picks the new
+# location up from this one constant.
```
GaganNarula, line 35: no need for this comment, after i move the data

alp_data/datasets/pteroset.py
```diff
@@ -0,0 +1,350 @@
...
+import pandas as pd
+
+from alp_data import Dataset, DatasetConfig, DatasetInfo, register_dataset
+from alp_data.backends import BackendType
+from alp_data.datasets._windowing import clip_selection_table, window_selection_table
+from alp_data.io import AnyPathT, anypath, audio_stereo_to_mono, read_audio
+
+# Staged in the ingestion bucket. Moves under DATA_HOME (as WABAD and
+# most datasets resolve) when this lands; the loader picks the new
+# location up from this one constant.
```
GaganNarula, line 19: same comment here

alp_data/datasets/dartmouth_avian_soundscapes.py
```diff
@@ -0,0 +1,350 @@
...
+            elif isinstance(raw_st, pd.DataFrame):
+                st = raw_st
+            else:
+                st = pd.DataFrame()
+
+            audio_dur = len(audio) / float(sr)
+            if window_start is not None and window_end is not None:
+                st = window_selection_table(st, float(window_start), audio_dur)
+            else:
+                st = clip_selection_table(st, audio_dur)
```
GaganNarula, line 250: interesting, why would the selection table's begin times be longer than audio duration in the first place ?

alp_data/datasets/_windowing.py
```diff
@@ -0,0 +1,74 @@
+"""Shared helpers for datasets that support windowed reads.
```
GaganNarula, line 1: can we create a module called `helpers.py` in datasets/ and add these functions there ?

alp_data/datasets/_windowing.py
```diff
@@ -0,0 +1,74 @@
...
+manifest ships those columns). Their selection tables hold recording-absolute
+times, so a windowed read has to re-base the table onto the returned audio or
+the two disagree.
+"""
+
+from __future__ import annotations
+
+import pandas as pd
+
+BEGIN = "Begin Time (s)"
```
GaganNarula, line 14: hmm, these constant column names being defined in a helper module here is a bit of code smell to me... can we instead add them as tail parameters to the functions themselves with default values as the ones listed as constants ?

GaganNarula: a signature like `st: pd.DataFrame, *, window_start: float, duration: float, begin_column: str = "Begin Time (s)", end_column: str = "End Time (s)")` would be nice

alp_data/datasets/xeno_canto_strong.py
GaganNarula: i wonder if this name is too generic... will this set of annotations on XC grow over time ? more of the same expected ? if yes, we can keep this name ..

tests/unittests/datasets_tests/test_dartmouth_avian_soundscapes.py
```diff
@@ -0,0 +1,218 @@
...
+CORE_ST_COLUMNS = ["Begin Time (s)", "End Time (s)", "Low Freq (Hz)", "High Freq (Hz)", "Species"]
+EXTRA_ST_COLUMNS = ["Species Code", "Common Name", "Background"]
+AUDIO_FP_PREFIX = 'recordings/'
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def ds_all() -> DartmouthAvianSoundscapes:
+    """Full manifest (metadata-level assertions; avoid bulk audio here)."""
+    return DartmouthAvianSoundscapes(split="all", backend="pandas")
```
GaganNarula, line 36: generally better to use `streaming=True` unless these splits are relatively small (few 10's of MB)

GaganNarula (commented):
some comments

alp_data/datasets/xeno_canto_strong.py
david-rx: From what I understand, there is no overlap with this and the current Xeno-canto dataset, nor with Jeantet. This will grow over time!

